PR #36, opened 2017-02-14
Fix recipe references in getRecipe.

bcl:
The returned object has the recipe under recipes[0].recipe and the
dependencies under recipes[0].modules


bcl:
Fix in a different way by the metadata fetch refactoring.

PR #42, opened 2017-02-27
including dependencies for supporting web components in all browsers

jgiardino:
Pages that included tabs, which is a patternfly web component, were not working in Firefox. This update includes dependencies required to include support patternfly web components in all browsers.


jgiardino:
that second commit was supposed to say "updating readme"

jgiardino:
Are you referring to this line: ENV NODE_VERSION 6.7.0?

PR #43, opened 2017-02-27
Including ability to create a new recipe

jgiardino:
These updates include the ability to click Create Recipe on recipes page, specify a recipe name and submit the form.

After clicking Save or pressing Enter in the Create Recipe dialog, the Edit Recipe page will display with the new recipe name.


jgiardino:
Right, I'm going to hold off on being able to add a component from the list of filtered components until I work on the filter feature again.

PR #52, opened 2017-03-16
Update README

atodorov:
I have removed documentation which came from the original patternfly starter application and replaced it with relevant information about Welder!

ATM I'm not sure if we need to delete the `How to Deploy` section b/c it talks about deployment on Google Firebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,83 @@
-# Patternfly React Starter
+# Welder Web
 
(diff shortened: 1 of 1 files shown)


atodorov:
[[UI] update README](https://trello.com/c/pylav4V1/236-ui-update-readme)

PR #55, opened 2017-03-17
Pagination

jgiardino:
This update includes a pagination control above the list of available inputs.

The three main elements of this control are:

- previous page button - Disabled if the current page is the first page. Allows the user to navigate to the previous page.
- next page button - Disabled if the current page is the first page. Allows the user to navigate to the previous page.
- page input - Allows the user to specify a page to navigate to. Does not accept a value other than empty or a number within the page range. If the user changes the value then exits the field without submitting, the value resets to match the current visible page.

These controls make updates to how inputs are set in state. The state is now an array of arrays, with each array being the set of items in that page.

These controls also work with filters. The list of all available inputs is maintained, whereas filtered inputs are cleared when the filter is cleared. This might be something we want to review later to see if it affects performance.

### DIFF
--- a/components/Pagination/Pagination.js
+++ b/components/Pagination/Pagination.js
@@ -76,6 +76,50 @@ class Pagination extends React.Component {
         />
       );
     }
+    let previousPage = null;
+    if (currentPage === 0) {
+      previousPage = (
+        <li className="disabled">
+          <a>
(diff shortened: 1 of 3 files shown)


atodorov:
I managed to test a bit late but looks good to me too.

PR #56, opened 2017-03-20
Allow app to be used as a cockpit package

petervo:
This really more of a proposal suggesting we run this web interface as a cockpit package, instead of requiring a stand alone server.

I'm sure this doesn't quite port everything over correctly and there are still issues to be worked out, but it's more of a POC of how this could happen.

Some of the basic principles are:

1) All http requests made from the javascript need to use relative paths.
2) There should be no inline js or css
3) All requests to the api should be made using cockpit.http type requests, and not directly from the javascript.
4) Because the app itself is loaded in a iframe, use hashes for navigation within the SPA. That way the cockpit shell can keep the top level navigation up to date.


stefwalter:
@petervo Most of the code is just plain cleanup. Should we move the code cleanup parts (ie: Using bower, and cleanup sitemap) into their own pull request?

Lets keep in mind that Weldr would want to be able to run both standalone *and* as a Cockpit package. So:

> 1. All http requests made from the javascript need to use relative paths.

This holds.

> 2. There should be no inline js or css

Agree that having inline javascript or CSS is a bad security practice. But in the worst case, Cockpit can relax its strict default security if a component like Weldr is included. See the ```content-security-policy``` setting here: http://cockpit-project.org/guide/latest/packages.html#package-manifest

> All requests to the api should be made using cockpit.http type requests, and not directly from the javascript.

Here's where the POC needs some adjustment. This needs to be done opportunistically, and only if running within Cockpit. I imagine there's a simple way to pull this off.

> Because the app itself is loaded in a iframe, use hashes for navigation within the SPA. That way the cockpit shell can keep the top level navigation up to date.

This is just best practice anyway.

petervo:
> Here's where the POC needs some adjustment. This needs to be done opportunistically, and only if running within Cockpit. I imagine there's a simple way to pull this off.

Didn't realize we wanted to keep old behavior as well. Should be relatively simple. I'll adjust tomorrow.

petervo:
Updated so now it works both with and without cockpit.

core/constants.js
```diff
@@ -4,19 +4,19 @@
...
-  get_compose_types: welder_api_host + "/api/v0/compose/types",
-  post_recipes_new: welder_api_host + "/api/v0/recipes/new",
+  get_recipes_list: "/api/v0/recipes/list",
+  get_recipes_info: "/api/v0/recipes/info/",
+  get_recipes_deps: "/api/v0/recipes/depsolve/",
+  get_modules_list: "/api/v0/modules/list",
+  get_module_info: "/api/v0/projects/info/",
+  get_dependencies_list: "/api/v0/modules/info/",
+  get_compose_types: "/api/v0/compose/types",
+  post_recipes_new: "/api/v0/recipes/new",
```
bcl: This will need to rebase on the new master -- API constants should match the route path to keep things clear.

bcl:
This looks pretty good, thanks! How would I go about running this inside cockpit, from a f25 install of it?

petervo:
Simplest way build the project

```
node run.js build
```

Link the built files to cockpit's package dir.

```
mkdir -p ~/.local/share/cockpit
ln -s /path/to/welder-web/public ~/.local/share/cockpit/welder-web
```

bcl:
Thanks, so I've been able to test non-cockpit running from inside docker. I can't seem to get it to run from cockpit, it expects things like patternfly.min.css, etc. to be in .../welder-web/dist/public/

petervo:
Did you run bower install?

petervo:
If that is the problem I'll add that step to the build task in run.js

bcl:
> Did you run bower install?

No, I didn't see it mentioned anywhere. Adding that to run.js would be a good idea since I'm not sure how to install it (tried npm install bower and it did some things, but isn't available anywhere that I can find on the PATH).

jgiardino:
I'm not an expert on this, but wanted to pass this feedback along from Patrick Riley regarding the use of bower vs npm:

> So the only issue I see with that PR is the adoption of Bower. Almost all front end packages now are moving to NPM. Bower is kind of dying at this point (even though it's still technically viable). 

If there's an issue with long-term support of bower, should we be using npm instead?

stefwalter:
> If there's an issue with long-term support of bower, should we be using npm instead?

Yup. Especially for a newer project like yours. Makes sense to think ahead.

petervo:
Ok I rebased and removed all the bower bits. Though I still think loading from external CDNs is not a good idea. I'll leave it alone for now.

jgiardino:
I agree with you @petervo. It would be good to remove the CDN links and use npm for patternfly css instead.

bcl:
Ok, I think this is working. It was a bit hard to figure out, so here's what ended up working:

 * `git clean -xdf && npm install && node run build`
 * `ln -s /path/to/public /usr/share/cockpit/welder-web/`

Restart cockpit *after* making changes and running node run build, otherwise the file hashes won't match what it expects and main.js won't load. That tripped me up for a bit.

@jgiardino  I also agree that CDN sources shouldn't be used. We can change that in a separate PR though.

petervo:
@bcl you should link it to ~/.local/share/cockpit/welder-web instead. That way you only need to log out of cockpit when you first make the link. After that the hashes don't matter and a simple refresh will pick up changes after each rebuild. Of course you need to make sure you log into cockpit as the same user who's home dir you used.

stefwalter:
> ln -s /path/to/public /usr/share/cockpit/welder-web/

During development do the following, and then hashes will be disabled, and you'll just be able to refresh Cockpit without logging out and logging back in:

 * ln -s /path/to/public ~/.local/share/cockpit/welder-web

Now log in with your local user. The one that owns ```~```

bcl:
Thanks, I had tried that initially and it didn't work, but that was before removing bower so that's probably what the real issue was.

I think this is almost ready to merge. The commit message needs to be cleaned up (No WIP: commits allowed) and maybe add any guidelines we need to follow so that future development doesn't break cockpit compatibility. "Always use relative URLs" seems like the big one :)

petervo:
> I think this is almost ready to merge. The commit message needs to be
> cleaned up (No WIP: commits allowed) and maybe add any guidelines we
> need to follow so that future development doesn't break cockpit
> compatibility. "Always use relative URLs" seems like the big one :)

I can do this, where should I put those instructions, commit message? 
README?

bcl:
> I can do this, where should I put those instructions, commit message?
> README?

A README would be best, so it doesn't get lost in the mists of time.

petervo:
Update readme and commit message

Issue #48, opened 2017-03-09
Save recipe needs to provide visual indication on the screen


jgiardino:
### Description of problem
On the Edit Recipe page, when you click Save, visual indication should display communicating that save is in process and save is successful.

### Component (web, API, etc)
Web

### Steps to Reproduce

1. Navigate to the Edit Recipe page
2. Make changes (add or remove components)
3. Click Save

### Actual results
No visual change on the screen

### Expected results
Visual indication in the screen that Save was clicked and save is happenening and has happened.

PR #63, opened 2017-04-07
Save updates - include version; include notification

jgiardino:
Includes updates for:
* https://trello.com/c/jpScBhhz/228-5-ui-update-save-to-use-updated-api
* https://trello.com/c/gZVZ2XEN/212-3-rfe-add-notifications-when-recipe-is-saved

Save was updated to include version number so that
version is automatically incremented, then recipe details
are reloaded to display new version number. Version number
is now visible in the Edit Recipe header.

Notifications were updated to include messages for save;
messages were also refactored; close function was updated
to accept either a message index or message label; notifications
component was also updated to support all message types (e.g.
warning, error, info).

### DIFF
--- a/components/Notifications/Notification.js
+++ b/components/Notifications/Notification.js
@@ -63,6 +63,15 @@ class Notification extends React.PureComponent {
     } else if (notification.type === 'process') {
       icon = <span className="pficon"><div className="spinner spinner-inverse"></div></span>;
       modifier = 'alert-info';
+    } else if (notification.type === 'info') {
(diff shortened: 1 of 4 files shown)


henrywang:
- Revision field is able to be increased and refreshed when save button clicked, which is mentioned on https://github.com/weldr/bdcs-api-rs/pull/38.
- Comment is not implemented on UI.
- Notification works well when save button clicked.

jgiardino:
This fixes issue #48

PR #65, opened 2017-04-11
UI cleanup

jgiardino:
Includes updates for these cards:

- https://trello.com/c/YUynepPP/254-2-summit-ui-cleanup
- https://trello.com/c/qTssJc8J/255-1-summit-update-header

Updates are intended to remove contents that are not related to the app, and to remove any instances of a product name from the UI. These updates are for the purposes of demo-ing at Summit.


henrywang:
Hi Jenn, 
One more thing I found:

1. The Users page has title "Patternfly React Boiler | Users". It should be "Users".

2. The Overview page tile is "Home". It should be "Overview" if follow the same name standard.

Thanks,
Xiaofeng

jgiardino:
Thanks for catching that, @bcl! I made updates so that the tooltip is hidden on click, and fixed the text that displays so that it matches.

@henrywang, thanks for reviewing! I updated the title for the Users page. That's a good point about the Overview page title. I left it as is for now, and we can finalize the title when we start finalizing the contents on the page.

henrywang:
Test passed. Thanks Jenn!

PR #67, opened 2017-04-18
UI-API updates

jgiardino:
Addresses general UI cleanup related to latest updates to API.

https://trello.com/c/NvRWvlWp/257-5-summit-ui-api-refactoring

For any areas where data was not available but we would like feedback during user testing, I included "---".

There were also some areas where data was available but needed to be updated to display correctly.

Some of updates included are:

- updating list of inputs to not show Dependency count in the popover, and to include version and release data
- showing the full set of inputs (previously only the inputs for which metadata was availalble were getting listed)
- getting a list of dependencies for a list item when expanded. 
- displaying icons for dependency list items
- not displaying Edit action when viewing details of a dependency

Some updates not included are:

- saving release along with version
- running depsolve against recipe on Add
- removing duplicate options in the Version-Release menu during component edit/selection


henrywang:
Found an issue which is not related with this request. The URL on the components/ListView/ListItemComponents.js is empty, but not on components/ListView/ComponentDetailsView.js.

components/ListView/ListItemComponents.js
```diff
@@ -200,15 +218,15 @@ class ListItemExpand extends React.Component {
                       </div>
                       <div className="list-view-pf-body">
                         <div className="list-view-pf-description">
-                          <a href="#" data-item="name">{listItem.requiredBy}</a>
+                          <a href="#" data-item="name">{listItem.requiredBy}---</a>
```
henrywang: Should be '---' only inside \<a\> or 
```javascript
{`${(listItem.requiredBy.length > 0) ? listItem.requredBy : '---'}`}
```

henrywang:
This PR passed end-to-end test (19 cases passed).

jgiardino:
Thanks for the review! I made updates based on @henrywang's comments.

PR #69, opened 2017-04-20
fixing bug with removing components

jgiardino:
This fixes a bug related to removing components. 

Steps to reproduce:

1. Find a recipe where recipe.modules length > 0
2. Remove one of the components that is listed in recipe.modules
3. Save the recipe then refresh the page
4. Result => the component is still listed in the recipe

On save, the components that are saved with the recipe are the ones in recipe.modules and recipe.packages. If the component that's removed is originally from recipe.modules, but isn't removed from recipe.modules, then the result of remove and save will not actually remove the component.

To fix this, the components listed in recipe.packages are identified as "RPM" and components listed in recipe.modules are identified as "Module". This way, recipe.packages and recipe.modules can be correctly updated when components are removed from the recipe.

### DIFF
--- a/data/RecipeApi.js
+++ b/data/RecipeApi.js
@@ -23,16 +23,17 @@ class RecipeApi {
         let p = new Promise((resolve, reject) => {
             utils.apiFetch(constants.get_recipes_deps + recipeName)
             .then(data => {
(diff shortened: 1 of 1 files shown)


henrywang:
Verified. Passed 19 cases of regression tests.

jgiardino:
Thanks for testing, @henrywang!

PR #70, opened 2017-04-20
UI refinements related to creating/viewing recipes

jgiardino:
- Recipes page was updated so that expand/collapse ability is removed (previously a hard-coded composition was listed in the expanded section of a recipe list item). For now, the code to expand/collapse is just commented out, since this will probably be added back in.
- Create Recipe dialog has updates related to not showing error messages on Cancel, and removing the background color on Save
- Ability to edit the description was added to the Recipe page. On the recipe page, Details tab, clicking the description or Edit icon next to the description displays an input field. When the description is modified, the change can be saved or canceled. If saved, the recipe is immediately updated in the database.

### DIFF
--- a/components/ListView/RecipeListView.js
+++ b/components/ListView/RecipeListView.js
@@ -5,48 +5,48 @@ import CreateComposition from '../../components/Modal/CreateComposition';
 class RecipeListView extends React.Component {
   state = { recipe: '' };
 
-  componentDidMount() {
-    this.bindExpand();
-  }
+  // componentDidMount() {
+  //   this.bindExpand();
+  // }
+  //
+  // componentDidUpdate() {
+  //   this.unbind();
+  //   this.bindExpand();
+  // }
+  //
+  // componentWillUnmount() {
+  //   this.unbind();
+  // }
 
-  componentDidUpdate() {
-    this.unbind();
-    this.bindExpand();
-  }
-
-  componentWillUnmount() {
-    this.unbind();
-  }
-
-  bindExpand() {
-    // click the list-view heading then expand a row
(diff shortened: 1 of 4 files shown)


henrywang:
1. cleaning up UI in list of recipes: Passed.

2. includes ability to edit recipe description: It's able to edit recipe description. But there's a problem about display new changed recipe on Details page.
    how to reproduce:
    1. go to recipe Details page, edit description and save.
    2. back to recipes list page, the description of that recipe changed.
    3. enter recipe Details page again, the recipe description shows before changed content.
    4. to show changed content, I have to refresh the page.

3. updates for Create Recipe modal dialog:
    Issue: the Create Recipe window dismissed and page switched to Edit Recipe page after click Save button. But the whole page is covered by gray mask, and all elements in page are not click-able or edit-able. Refreshing page is the only way to make it normal.

jgiardino:
Thanks for reviewing, @henrywang! I was testing in debug mode, which somehow prevented me from seeing the issue you noted about the Create Recipe dialog. I included a commit that fixes that issue.

I'm not able to reproduce the issue you note for editing the description, though.

henrywang:
Thanks for fix Create Recipe issue. Your fix is verified.
The editing description issue was happened on some of recipes, not all of recipes. I'm trying to find some common patterns of recipes with this issue. But I encountered another bdcs-api issue when I change recipe description.

jgiardino:
Thanks for reviewing again. I will go ahead and merge these updates. If you happen to notice which recipes are having issues with Edit Description, please let me know. This is just a preliminary implementation for that feature. It still needs a few updates, but we also want to get user feedback on this feature as well, so it could change.

PR #71, opened 2017-04-20
Patternfly css

jgiardino:
This PR replaces the cdn links to patternfly css and instead uses:

- npm install to add patternfly as a dependency
- copy-webpack-plugin to copy the patternfly files to the /public/css, /public/fonts, and /public/img folder

This PR also includes a blank placeholder svg in the header for the purposes of usability testing.


henrywang:
Test passed and regression test passed.

PR #74, opened 2017-04-21
implementing Edit action in component li kebab

jgiardino:
This enables the Edit action that displays in the kebab menu of a component listed under Selected Components in the Edit Recipes page.

Before this PR, you could click the component name to navigate to the details view, then click the Edit button to display the form for selecting a different version-release. This is still available.

But now, clicking Edit in the kebab on a component list item will navigate to the details view with the form automatically visible.


henrywang:
Test passed and regression test passed.

jgiardino:
Thanks for reviewing!

PR #75, opened 2017-04-21
removing href from nonfunctional links

jgiardino:
These are updates related to general UI clean up for user testing:

There are non-functional links that are included in the UI for the purposes of user review. But when the user clicks on one of these links, the user is redirected to the home page (i.e. "Overview") due to the `href="#"`. For any link that is not implemented yet, these href attributes have been removed, to avoid any unnecessary confusion during user testing.

Values for "Revision" are replaced with a hard-coded value of "3" so that it matches the mock data. This is to avoid any confusion during user testing.

On the Recipe page, Revisions tab, classes for the expand/collapse arrow are updated to fix a sizing issue. When the down arrow became visible after expanding a list item, the arrow was not the correct size.

### DIFF
--- a/components/Layout/Header.js
+++ b/components/Layout/Header.js
@@ -36,8 +36,8 @@ function Header() {
               <span className="caret"></span>
             </a>
             <ul className="dropdown-menu" aria-labelledby="dropdownMenu1">
-              <li><a href="#">Help</a></li>
-              <li><a href="#">About</a></li>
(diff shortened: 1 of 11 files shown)


jgiardino:
Thanks for reviewing, @henrywang! I updated the Recipes page and also fixed merge conflicts.

jgiardino:

PR #78, opened 2017-04-21
Includes scrollbars for nested lists

jgiardino:
In a list view there list items can be expanded, if the content are
includes a nested list, this sets a max height so that the parent list
item doesn't get too tall.

An example of this is any component list item (Selected Component or Dependency) on either the Recipe page or Edit Recipe page. The nested list is the list of Dependencies that displays in the expanded content area of a component list item.


jgiardino:

PR #79, opened 2017-04-22
Fixing issue with 'Saving' notification

jgiardino:
When clicking Save on Edit Recipe page, a saving
notification should display immediately. This update
triggers the Layout component to reload and display
the saving notification. When the recipe is saved,
the 'saving' notification is then replaced with a 'saved'
notification.


jgiardino:

PR #81, opened 2017-04-24
fixes issue with updating cached description

jgiardino:
@henrywang, this fixes the issue you had noted before about Edit Description and navigating to the Edit Recipe page.

Steps to reproduce the issue being fixed:
1) edit the description on the Recipe page, Details tab
2) navigate to the Edit Recipe page
3) navigate back to the Recipe page

Results:  the previous description before the edit
is visible. But clicking reload on the page will show the
description that was saved.

This issue occurred because the recipe data gets cached,
but the cached description wasn't getting updated upon
editing the description.

### DIFF
--- a/data/RecipeApi.js
+++ b/data/RecipeApi.js
@@ -173,6 +173,9 @@ class RecipeApi {
   }
 
   handleEditDescription(description) {
+    // update cached recipe data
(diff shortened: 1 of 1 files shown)


jgiardino:
Thanks for reviewing, @bcl!

PR #84, opened 2017-04-27
Add recipe save notification case.

henrywang:
Acceptance criteria: When I click the Save button after some
changes on a recipe I should be able to see a notification
which tells me that the recipe was saved.


henrywang:
Thanks @bcl!

PR #85, opened 2017-04-27
Add wait selector step when switching pages to make case more stable.

henrywang:
That will not increase case running by adding wait selector step.
From my debug result, all of wait selector step just cost 0ms.

`nightmare queueing action "wait" +0ms`


henrywang:
Thanks @bcl!

PR #86, opened 2017-04-28
Remove kickstart composition type and add more types due to UI change.

henrywang:
Add two wait steps in Save Recipe Test to improve case running stability.

### DIFF
--- a/test/end-to-end/config.json
+++ b/test/end-to-end/config.json
@@ -14,8 +14,25 @@
             "packages": []
         }
     },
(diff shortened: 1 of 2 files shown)


henrywang:
Due to UI change, there's no kickstart composition type any more. UI added more composition type, so we need to set them up in config.json.

atodorov:
[(2) Trim down recipes examples](https://trello.com/c/zYNcegoP/293-2-trim-down-recipes-examples)

PR #88, opened 2017-05-02
Add test case ID in each case and test type in each group cases.

henrywang:
Update README.md to support case ID and test type.

Add a new npm running script in package.json to support run Acceptance
test only.

Update "Save Recipe Test" to fix failed issue in Travis only.

### DIFF
--- a/test/end-to-end/README.md
+++ b/test/end-to-end/README.md
@@ -29,6 +29,13 @@ $ npm run test                                  # Run end-to-end test
 $ npm run xunittest                             # Run end-to-end test with xunit xml result output
 ```
 
+```shell
+$ mocha --grep=#acceptance                                 # Run acceptance case only
+$ mocha --grep=@create-recipe-page                         # Run create recipe page related case only
+$ mocha --grep=#acceptance.+@create-recipe-page            # Run acceptance case in create recipe page cases
+$ mocha --grep="@create-recipe-page|@edit-recipe-page"     # Run create recipe and edit recipe pages cases
+```
+
 ### Running end-to-end test in Docker
 
 The end-to-end test docker image is an executable image, which starts container, runs test, and exits.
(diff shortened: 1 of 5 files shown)


test/end-to-end/README.md
```diff
@@ -49,6 +49,10 @@ Run end-to-end test like this:
 
 XUnit XML format result will be generated at */var/lib/docker/volumes/test-result-volume/_data/result.xml*
 
+## Test Suites and Cases Management
+
+The test suites and cases are managed by Google Sheets - [Welder-Web End-to-End Quality Testsuites](https://docs.google.com/a/redhat.com/spreadsheets/d/1ZYr8xLG-BNT6qMWVxlNSZEVELwo0rSEseA40oaIMNyw/edit?usp=sharing).
+
```
bcl: Could you explain how this relates to the changes? And is it publicly visible? If not, it shouldn't be in the README

henrywang: Thanks @bcl! Removed this section from README.md.

test/end-to-end/README.md
```diff
@@ -105,18 +105,18 @@ describe('View Recipe Page', function () {                // Which page does thi
     after((done) => {                                     // Scenario de-provision, like apply a snapshot
     });
 
-    context('Menu Nav Bar Check', () => {                 // Element or elements to be test
-      it('should show a recipe name', (done) => {         // Expected result description
+    context('Menu Nav Bar Check #Acceptance', () => {     // Test content and test type
+      it('should show a recipe name @Welder-Web-40003', (done) => {    // Case description and case ID
```
atodorov: What are these numbers? I don't like having this defined by hand b/c it will get out of sync pretty quickly. What is the purpose of having the numbers ?

atodorov: Test type is fine, but IMO we don't need the case ID because the titles are descriptive enough. I don't see how ID add any new information to the mix.

henrywang: Sorry for confused you. Let me explain the use case of case ID.
The case ID number is based on page:
* 1xxxx - Edit Recipe Page
* 2xxxx - Create Recipe Page
* 3xxxx - Recipes Page
* 4xxxx - View Recipe Page
If only Recipes Page cases need to be run, "mocha --grep @Welder-Web-3*" command will be executed.

atodorov: Ahh, In this case I prefer doing something like 
`mocha --grep @welder-web-recipes-page-*`

this is more human friendly and easier to remember.

henrywang:
Updated case ID naming method based on @atodorov's suggestion.

PR #101, opened 2017-05-26
Fix issues.

henrywang:
1. Fix prop type issue on edit recipe page.
2. Update end-to-end eslint setting to have the same max-len (130).
3. Fix accessing PropTypes via the main React package deprecated issue.

### DIFF
--- a/components/Actions/Actions.js
+++ b/components/Actions/Actions.js
@@ -1,11 +1,9 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import constants from '../../core/constants';
 import utils from '../../core/utils';
 
 class Actions extends React.Component {
-  static propTypes = {
-    className: React.PropTypes.node,
-  }
 
   state = { actions: [] };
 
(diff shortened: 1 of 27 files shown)


henrywang:
[[UI] Resolve React package deprecated issue.](https://trello.com/c/qDIyWDmn/304-ui-resolve-react-package-deprecated-issue)

atodorov:
@jgiardino @bcl can you guys take a closer look at this PR. From what I can tell from the current builds the react/prop-types errors are now gone, which means we probably don't need this PR:
https://travis-ci.org/weldr/welder-web/builds/236222532

Why did the prop-types errors disappear? I don't really know.

@henrywang can you split out the .eslintrc change into a separate commit. That can be merged I think but my React.js is not strong enough to tell about the rest of the changes.

jgiardino:
This PR looks good to me. There were commits I made before summit where I had fixed some of the prop-type errors, which may be why they disappeared. But I didn't realize the method I was using was deprecated, as stated in this card:   [[UI] Resolve React package deprecated issue.](https://trello.com/c/qDIyWDmn/304-ui-resolve-react-package-deprecated-issue) 

I'm good with keeping the updates made in this PR to how prop-types are defined.

PR #102, opened 2017-05-26
Create ./core/build-version.json with the git build version

bcl:
This will contain a JSON object with "build-version" set to the git
tag and hash, or to UNKNOWN if git fails.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ public/css
 public/fonts
 public/img
 db.json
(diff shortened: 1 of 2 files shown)


bcl:
This is just the data, I'm not sure where to add it to the UI. To a footer maybe? Or in an About dialog? The API's build version will also be available once https://github.com/weldr/bdcs-api-rs/pull/69 is pushed.

PR #103, opened 2017-06-09
Add recipe export test case.

henrywang:
1. add case to test recipe export on recipes, view recipe, and edit recipe page.

### DIFF
--- a/test/end-to-end/config.json
+++ b/test/end-to-end/config.json
@@ -4,35 +4,28 @@
         "uri": "http://localhost:4000/",
         "test": "api/v0/test",
         "newRecipe": "api/v0/recipes/new",
-        "deleteRecipe": "api/v0/recipes/delete/"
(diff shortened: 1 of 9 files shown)


henrywang:
[[QE][UI][Automation] Add test case for (3) Manifest file - include link in UI](https://trello.com/c/o8cSAabs/318-qeuiautomation-add-test-case-for-3-manifest-file-include-link-in-ui)

henrywang:
@atodorov, @bcl already approved my PR #104 which fixed docker build error.
For the eslint issue in my code is all about max-len, this issue got fixed in PR #101. I updated .eslintrc in end-to-end folder with ```"max-len": ["error", 130]``` added (100 by default). The 130 max-len comes from package.json setting in root folder. If the new 130 applied, the eslint issue in this PR will not be an issue any more.

PR #106, opened 2017-06-12
Update importSanity test with some changes:

henrywang:
1. Add wait time before getting total count to make this number loaded.
2. Add a before hook to check exist of metadata.db file to avoid useless
   metadata.db generated by new dqlite3.database command.

### DIFF
--- a/test/end-to-end/test/test_importSanity.js
+++ b/test/end-to-end/test/test_importSanity.js
@@ -4,10 +4,20 @@ const EditRecipePage = require('../pages/editRecipe');
 const apiCall = require('../utils/apiCall');
 const pageConfig = require('../config');
 const sqlite3 = require('sqlite3').verbose();
+const fs = require('fs');
 
 describe('Imported Content Sanity Testing', function () {
   this.timeout(15000);
-  const db = new sqlite3.Database(process.env.MDDB || 'metadata.db');
+  let db;
+
+  before((done) => {
+    // Check exist of metadata.db file first
+    fs.access(process.env.MDDB || 'metadata.db', (error) => {
+      if (error) return done(error);
+      db = new sqlite3.Database(process.env.MDDB || 'metadata.db');
+      return done();
+    });
+  });
 
   before((done) => {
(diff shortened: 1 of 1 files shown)


test/end-to-end/test/test_importSanity.js
```diff
@@ -4,10 +4,20 @@ const EditRecipePage = require('../pages/editRecipe');
...
 
 describe('Imported Content Sanity Testing', function () {
   this.timeout(15000);
-  const db = new sqlite3.Database(process.env.MDDB || 'metadata.db');
+  let db;
+
+  before((done) => {
+    // Check exist of metadata.db file first
+    fs.access(process.env.MDDB || 'metadata.db', (error) => {
+      if (error) return done(error);
```
atodorov: Where does the error message come from? Is it a 'File not found' type of error ?

test/end-to-end/test/test_importSanity.js
```diff
@@ -33,6 +43,7 @@ describe('Imported Content Sanity Testing', function () {
       const nightmare = new Nightmare();
       nightmare
         .goto(editRecipePage.url)
+        .wait(1000) // Have to wait for a while to make total count loaded
```
atodorov: Can we add  a more robust wait? For example in other frameworks (selenium or capybara+rspec) you can instruct the browser to wait until a specific condition is met, like an HTML element has been loaded for example. 

My problem with hard-coded timeouts is that they tend to expire and we have to constantly bump them up, which increases overall test execution time.

test/end-to-end/test/test_importSanity.js
henrywang: Yes, for example.
```
1 failing

  1) Imported Content Sanity Testing "before all" hook:
     Error: ENOENT: no such file or directory, access 'metadata.db'
      at Error (native)
```

atodorov:
I am merging this after discussing it with @henrywang. We will open another card to track updates to the wait condition to make the test more robust.

PR #107, opened 2017-06-12
Some imporvements

henrywang:


henrywang:
[More robust JS wait for DB count e2e test case](https://trello.com/c/CYZnk83i/319-more-robust-js-wait-for-db-count-e2e-test-case)

test/end-to-end/test/test_createRecipe.js
```diff
@@ -51,23 +53,23 @@ describe('Create Recipe Page', function () {
 
       it('should switch to Edit Recipe page - recipe creation success @create-recipe-page', (done) => {
         // Highlight the expected result
-        const expected = editRecipePage.url;
+        // const expected = editRecipePage.url;
```
atodorov: Remove this comment

test/end-to-end/test/test_editRecipe.js
```diff
@@ -120,14 +126,19 @@ describe('Edit Recipe Page', function () {
         const nightmare = new Nightmare();
         nightmare
           .goto(editRecipePage.url)
+          .wait(editRecipePage.componentListItemRootElement)
           .wait(editRecipePage.btnCreateCompos)
           .click(editRecipePage.btnCreateCompos)
-          .wait(createComposPage.btnCreate)
+          .wait(page => document.querySelector(page.dialogRootElement).style.display === 'block'
+            , createComposPage)
```
atodorov: Why don't you query for the button instead ? I don't understand this change.

test/end-to-end/test/test_editRecipe.js
```diff
@@ -136,7 +147,10 @@ describe('Edit Recipe Page', function () {
           })
           .then(() => nightmare
             .wait(toastNotifPage.iconComplete)
-            .wait(toastNotifPage.labelStatus)
+            .wait((page) => {
+              const recipeName = document.querySelector(page.labelRecipeName).innerText;
+              return recipeName !== page.varEmptyName && recipeName.includes(page.varEmptyName);
+            }, toastNotifPage)
```
atodorov: Same question as above. Why are we waiting for something different here ?

test/end-to-end/test/test_editRecipe.js
```diff
@@ -151,18 +165,35 @@ describe('Edit Recipe Page', function () {
...
+            expect(element).to.equal(expectedSaving);
+          })
+          .then(() => nightmare
+            .wait(toastNotifPage.iconComplete)
+            .wait((page) => {
+              const recipeName = document.querySelector(page.labelRecipeName).innerText;
+              return recipeName !== page.varEmptyName && recipeName.includes(page.varEmptyName);
+            }, toastNotifPage)
+            .evaluate(page => document.querySelector(page.labelStatus).innerText
+            , toastNotifPage)
```
atodorov: This wait is what I've been looking for in the above comments.

test/end-to-end/test/test_importSanity.js
```diff
@@ -43,8 +43,7 @@ describe('Imported Content Sanity Testing', function () {
       const nightmare = new Nightmare();
       nightmare
         .goto(editRecipePage.url)
-        .wait(1000) // Have to wait for a while to make total count loaded
-        .wait(editRecipePage.totalComponentCount)
+        .wait(editRecipePage.componentListItemRootElement) // list item and total number are rundered at the same time
```
atodorov: rundered -> rendered

test/end-to-end/test/test_editRecipe.js
henrywang: The Create Composition dialog page is already in the Dom even the dialog does not pop up. The way to control this page visible is update the div property style from "display: none;" to "display: block;". So the whether Create button in the Dom tree or not cannot prove the page is ready to use.

test/end-to-end/test/test_editRecipe.js
henrywang: Due to the page render is async, the label is rendered as ": Creating composition." without recipe name. When the api fetch get returned, the page will be rendered again with recipe name added. In this case we need to wait until a full notification like "automation: Creating composition." rendered.

test/end-to-end/test/test_editRecipe.js
henrywang: The same reason as above comment.

PR #108, opened 2017-06-15
Updated Composition tests to run for all types, not just ostree.

jkozol:
I have taken the existing composition test and looped it over all 5 composition types by placing the test within a forEach loop. I also re-factored the config so that "composition" is now an array of objects. Thanks for any feedback you can offer.


jkozol:
The eslint errors should be fixed now.

PR #109, opened 2017-06-21
Add unit test and code coverage.

henrywang:
1. Add unit test for pages/home.
2. Add code coverage support for unit test.
3. Add coveralls.io support.


henrywang:
[Produce coverage from welder-web unit tests](https://trello.com/c/7wndVO3n/327-produce-coverage-from-welder-web-unit-tests)

.travis.yml
```diff
@@ -14,7 +14,7 @@ before_install:
 env:
   - COMMAND=eslint
   - COMMAND=stylelint
-  - COMMAND=test
+  - COMMAND=unit-test
```
atodorov: Where does the coverage report get submitted to Coveralls? That should be a command inside `.travis.yml`. Also take into account that there are several jobs in the text matrix, see:
https://github.com/weldr/bdcs-api-rs/blob/master/.travis.yml#L26

README.md
```diff
@@ -41,7 +42,7 @@ available at [http://localhost:3000/](http://localhost:3000/).
 
 #### Unit Test
 
-The unit tests are powered by [chai](http://chaijs.com/) and [mocha](http://mochajs.org/).
+The unit tests are powered by [Jest](https://facebook.github.io/jest/), [Enzyme](http://airbnb.io/enzyme/) and [Sinon.JS](http://sinonjs.org/)
```
atodorov: I'm not super keen on the fact that we are introducing a new framework for the unit tests and this is different from the framework used for the e2e tests. Can we make use of the same frameworks?   From what I see in the code below there's no need to change the framework.

package.json
```diff
@@ -36,10 +36,12 @@
     "babel-preset-stage-1": "^6.5.0",
     "babel-register": "^6.11.6",
     "babel-runtime": "^6.11.6",
+    "bootstrap": "^3.3.7",
     "browser-sync": "^2.13.0",
     "chai": "^3.5.0",
     "connect-history-api-fallback": "^1.2.0",
     "copy-webpack-plugin": "^3.0.1",
+    "coveralls": "^2.13.1",
```
atodorov: npm install this only inside Travis! See above for example.

package.json
```diff
@@ -36,10 +36,12 @@
     "babel-preset-stage-1": "^6.5.0",
     "babel-register": "^6.11.6",
     "babel-runtime": "^6.11.6",
+    "bootstrap": "^3.3.7",
```
atodorov: Why does this appear?

package.json
```diff
@@ -55,6 +57,7 @@
     "front-matter": "^2.1.0",
     "highlight.js": "^9.5.0",
     "jest": "^16.0.0",
+    "jquery": "^3.2.1",
```
atodorov: What requires jquery?

package.json
```diff
@@ -86,9 +90,17 @@
     "webpack-hot-middleware": "^2.12.2"
   },
   "jest": {
-    "testRegex": "(/test/.*\\.spec.js)$",
+    "testRegex": "(/test/unit/.*\\.spec.js)$",
     "setupFiles": [
-      "test/env.js"
+      "test/unit/env.js"
+    ],
+    "collectCoverageFrom": [
```
atodorov: Is there any way to blackist the tests directory and leave all the others? Whitelisting files will always lead to us forgetting to whitelist some new directory/file when it becomes available. I want to reverse that.

package.json
```diff
@@ -133,6 +146,7 @@
     "start": "node run",
     "test": "jest",
     "test:watch": "npm test -- --watch",
-    "test:cov": "npm test -- —coverage"
+    "test:cov": "npm test -- --coverage",
+    "test:ci": "node --harmony_proxies node_modules/.bin/jest --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"
```
atodorov: IMO the submission to Coveralls should happen in `after_success` inside Travis, not here. I'm pretty sure this will lead to errors if executed locally b/c we won't have the Coveralls token configured.

test/.eslintrc
```diff
@@ -1,10 +0,0 @@
-{
-  "env": {
-    "jest": true
-  },
-  "rules": {
-    "no-console": 0,
-    "no-unused-expressions": 0,
-    "padded-blocks": 0
-  }
-}
```
atodorov: Why do we remove this file?

test/unit/env.js
```diff
@@ -0,0 +1,27 @@
+import jQuery from 'jquery';
+
+global.$ = global.jQuery = jQuery;
+require('bootstrap');
+global.welderApiHost = 'localhost';
+global.welderApiScheme = 'HTTP';
+
+// <mockLocalStorage>
+const localStorageMock = (() => {
```
atodorov: Why do we need this mock?

README.md
henrywang: The description in Readme.md was wrong, the default unit test runner was jest, not mocha. There's one line - "test": "jest" in package.json. So that the unit test did not use mocha.

package.json
henrywang: There're lots of code are based on jquery and bootstrap. The jquery and bootstrap are imported in index.html by following two lines of code.
```html
<script src="//code.jquery.com/jquery-2.1.4.min.js"></script>
<script src="//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"></script>
```
But in the unit test, these two libs has to be imported by node lib.

.travis.yml
henrywang: Good suggestions.

package.json
henrywang: OK. That makes sense to me.

package.json
henrywang: The same as above comment.

package.json
henrywang: The js files which need do unit test are distributed. The same situation as enslint checking.

package.json
henrywang: OK.

test/.eslintrc
henrywang: All of settings are covered by package.json.

test/unit/env.js
henrywang: We need a global window to attach our mocked localStorage API to in the Jest environment.

README.md
atodorov: Ok, question still stands. From what I understand now e2e tests use Mocha but unit tests use Jest. I'd like for us to use the same frameworks in both places.

package.json
atodorov: OK, I understand the need to import the libraries but the above snippet needs to be refactored as well. 

1) We can't rely on code being loaded from a public CDN, that needs to be vendored into the application itself
2) ATM we have version information for jquery and bootstrap in two places which can easily get out of sync. This needs to be only in `package.json`.

package.json
atodorov: Is there any way we can work around this? For example in Python you can instruct the coverage tool to "discover" all existing files in the project and include them in the report, even if they are not covered. This is an easy shortcut to say "print me a coverage report for everything". 

OTOH in Haskell we can't autodiscover files and include them in coverage. I'd like to avoid this for JavaScript if possible. The problem is that once you whitelist some files it is very easy to get a high number of coverage for them and leave it as is. Then when a new directory is added the new files will not show in the report and it will still remain at the previous number.  Which may lead to the false impression that there is sufficient test coverage while there isn't.  Can we avoid that somehow? Maybe try a different coverage collection tool for JS if available.

test/unit/env.js
atodorov: Sorry but I still don't understand why do we need to create the `localStorageMock` object? What will break if we don't create it?

henrywang: the line 478 in PfVerticalNavigation.js.
```javascript
if (window.localStorage.getItem('patternfly-navigation-primary') === 'collapsed') {
        collapseMenu();
 }
```
That will report error.

package.json
henrywang: I agree with you.

README.md
henrywang: Yes. But the jest has lots of advantage in React unit testing. Jest is recommended by Facebook to do unit test for React. And react has code coverage integrated, mocha needs a 3rd package to implement code coverage. The snapshot feature in Jest is very useful for React unit test. So I think we need to keep Jest to run unit test.

atodorov: OK then, but I will urge to tackle the migration of the e2e tests as the next priority item, otherwise it will get tedious to maintain and write tests in 2 different frameworks.

PS: IMO you can create a card for the e2e migration and work on it once this PR is merged.

test/unit/env.js
atodorov: Hmmm, let's leave it for now but I think this file was only used for debugging or something like that. We should investigate if we can remove it and drop the mock.

atodorov:
@jgiardino there are other bits and pieces which need to be done but they will be handled in separate cards. I'm fine to merge this if you don't have any objections.

PR #110, opened 2017-06-22
Modal updates and fixed cockpit integration.

jkozol:
Updated state for modals. Added new modal for pending changes in edit recipe. Fixed api port when run through cockpit.

### DIFF
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -1,0 +1,123 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class PendingChanges extends React.Component {
+
+  // dummy data
+  state = {
+    componentUpdates: [
+      { componentOld: null, componentNew: 'rsync-3.0-17.317' },
+      { componentOld: 'httpd-2.3-45.el7', componentNew: 'httpd-2.4-45.el7' },
+      { componentOld: 'basesystem-10.0-7.el7', componentNew: null },
+    ],
+  };
+
+  componentDidMount() {
+    $(this.modal).modal('show');
+    $(this.modal).on('hidden.bs.modal', this.props.handleHideModal);
+  }
+
+  handleSaveChanges() {
+    $('#cmpsr-modal-pending-changes').modal('hide');
+  }
+
+  render() {
+    return (
+      <div
+        className="modal fade"
+        id="cmpsr-modal-pending-changes"
+        ref={(c) => { this.modal = c; }}
+        tabIndex="-1"
+        role="dialog"
+        aria-labelledby="myModalLabel"
+        aria-hidden="true"
+      >
+        <div className="modal-dialog">
+          <div className="modal-content">
+            <div className="modal-header">
+              <button
+                type="button"
+                className="close"
+                data-dismiss="modal"
+              >
+                <span className="pficon pficon-close"></span>
+              </button>
+              <h4 className="modal-title" id="myModalLabel">Changes Pending Save</h4>
+            </div>
+            <div className="modal-body">
+              <form className="form-horizontal">
+                <div className="form-group">
+                  <label
+                    className="col-sm-3 control-label"
+                  >Recipe</label>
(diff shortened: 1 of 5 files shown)


components/Modal/PendingChanges.js
```diff
@@ -0,0 +1,123 @@
...
+                      rows="1"
+                      value={this.props.recipe.comment}
+                    />
+                  </div>
+                </div>
+                <div className="alert alert-info">
+                  <span className="pficon pficon-info"></span>
+                  Only changes to selected components are shown. <a href="#" className="alert-link">View all changes.</a>
+                </div>
+                <strong>Updates</strong>
```
jgiardino: Please change `<strong>Updates</strong>` to `<strong>Pending Changes</strong><span class="text-muted"> (most recent first)</span>`

This captures the suggestion that Steve made on the design, and also avoids using "update" to mean too many different things.

components/Modal/PendingChanges.js
```diff
@@ -0,0 +1,123 @@
...
+                <ul className="list-group">
+                  {this.state.componentUpdates.map((componentUpdated) => (
+                    <li className="list-group-item">
+                      {componentUpdated.componentNew && componentUpdated.componentOld &&
+                        <div className="row">
+                          <div className="col-sm-3"><strong>Updated</strong></div>
```
jgiardino: Please remove the use of `<strong>` tags for the first column of text. Otherwise, it's a lot of bold text.

pages/recipeEdit/index.js
```diff
@@ -470,6 +481,12 @@ class EditRecipePage extends React.Component {
       >
         <div className="cmpsr-edit-actions pull-right">
           <ul className="list-inline">
+            <li className="text-muted">
+              3 changes
+            </li>
+            <li>
+              <a onClick={(e) => this.handleShowModal(e, 'modalPendingChanges')}>View and Comment</a>
```
jgiardino: The UI needs to be keyboard accessible for general accessibility purposes. For an <a> tag to be keyboard accessible, it needs `href="#"`. Alternativley, `tabindex="0"` will also work, but for <a> I've been using the href attribute.

jgiardino:
Great job on this component! I included just a couple of minor comments for text strings and accessibility compliance. 

I'll create a separate card to track the comment field behavior (i.e. that it grows as the contents grow). I'm thinking that having it automatically get focus when the modal opens, if empty, would be a nice feature too, but we can track that as a separate story.

components/Modal/PendingChanges.js
```diff
@@ -0,0 +1,123 @@
...
+                  <span className="pficon pficon-info"></span>
+                  Only changes to selected components are shown. <a href="#" className="alert-link">View all changes.</a>
+                </div>
+                <strong>Updates</strong>
+                <ul className="list-group">
+                  {this.state.componentUpdates.map((componentUpdated) => (
+                    <li className="list-group-item">
+                      {componentUpdated.componentNew && componentUpdated.componentOld &&
+                        <div className="row">
+                          <div className="col-sm-3">Pending Changes<span className="text-muted"> (most recent first)</span></div>
```
jgiardino: This line of code should go at line 75 (i.e. replacing the title for this list). Here, you would just have "Updated", so that the list indicates what was added, removed or updated in the recipe.

jgiardino:
The updates look good! The only issue is that you replaced the wrong "Update" line of text ;-) I included a comment in the file with more details.

jgiardino:
Almost there. The text string "(most recent first)" should follow the text string "Pending Changes" above the list group, like the example below:
`<strong>Pending Changes</strong><span class="text-muted"> (most recent first)</span>`

The text "Updated" should not include "(most recent first)". In this case, "Updated" is indicating the type of change that was made.

jgiardino:
Looks great! Thanks!!

bitmoji

jkozol:
Great! Have a nice weekend.

PR #113, opened 2017-06-27
Updated CreateRecipe modal to prevent duplicate recipe names with a test.

jkozol:
When creating a new recipe there will now be an alert if one tries to create a recipe with a duplicate name. I also created a test to test this functionality.

### DIFF
--- a/components/Modal/CreateRecipe.js
+++ b/components/Modal/CreateRecipe.js
@@ -50,13 +50,14 @@ class CreateRecipe extends React.Component {
 
   handleEnterKey(event) {
     if (event.which === 13 || event.keyCode === 13) {
+      this.handleErrors(this.state.recipe.name);
       setTimeout(() => {
         if (this.state.showErrorName || this.state.showErrorDuplicate) {
           this.showInlineError();
         } else {
           this.handleCreateRecipe(event, this.state.recipe);
         }
-      }, 10);
+      }, 100);
     }
   }
 
@@ -154,6 +155,7 @@ class CreateRecipe extends React.Component {
                       id="textInput-modal-markup"
                       className="form-control"
                       value={this.state.recipe.name}
+                      onFocus={(e) => { this.dismissErrors(); this.handleErrorDuplicate(e.target.value); }}
                       onChange={(e) => this.handleChange(e, 'name')}
                       onBlur={(e) => this.handleErrors(e.target.value)}
                     />
--- a/components/Modal/CreateRecipe.js
+++ b/components/Modal/CreateRecipe.js
@@ -42,20 +42,21 @@ class CreateRecipe extends React.Component {
     const o = Object.assign({}, this.state.recipe);
     o[prop] = e.target.value;
     this.setState({ recipe: o });
-    this.dismissErrors();
-    this.handleErrors();
+    if (prop === 'name') {
+      this.dismissErrors();
+      this.handleErrors(e.target.value);
+    }
   };
 
   handleEnterKey(event) {
     if (event.which === 13 || event.keyCode === 13) {
-      this.handleErrors();
       setTimeout(() => {
         if (this.state.showErrorName || this.state.showErrorDuplicate) {
           this.showInlineError();
         } else {
           this.handleCreateRecipe(event, this.state.recipe);
         }
-      }, 400);
+      }, 10);
     }
   }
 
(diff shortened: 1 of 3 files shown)


jgiardino:
Great job implementing this! 

Originally I expected the following message to display after exiting the field, but currently it displays as I type. My expectation was that this behavior would be annoying, but now that I see it I think it's fine since it's giving more immediate feedback, which is nice in this case.
>The name "http-server" already exists.

However, I noticed an issue with the following message. With these updates, the following message now displays after entering one character into the name field.
>A recipe name is required.

components/Modal/CreateRecipe.js
```diff
@@ -107,17 +132,27 @@ class CreateRecipe extends React.Component {
...
+                :
+                <div></div>
+              }
+              {(this.state.inlineError && this.state.showErrorDuplicate) ?
+                <div className="alert alert-danger">
+                  <span className="pficon pficon-error-circle-o"></span>
+                  <strong>Specify a new recipe name.</strong>
+                </div>
+                :
+                <div></div>
```
jgiardino: Should we be using the other syntax (i.e. `&&` and `||` instead of `:` and `?`) so that we don't have to include the empty `<div></div>`?

components/Modal/CreateRecipe.js
```diff
@@ -65,6 +71,25 @@ class CreateRecipe extends React.Component {
...
+
+  handleErrorDuplicate() {
+    utils.apiFetch(constants.get_recipes_list)
+      .then(listdata => {
+        for (const listedName of listdata.recipes) {
+          // regex to replace a space with a '-'
+          const ourName = this.state.recipe.name.replace(/\s+/g, '-');
+          if (listedName === ourName) {
+            this.setState({ showErrorDuplicate: true });
+          }
```
henrywang: Add break here, otherwise, all of recipes will be looped.
Can we use Array.prototype.includes() to replace for..of, like this? IMO, this one is more efficient than for...of.
```javascript
listdata.recipes.includes(ourName)
```

test/end-to-end/test/test_createRecipe.js
```diff
@@ -31,6 +31,40 @@ describe('Create Recipe Page', () => {
...
+        const expected = createRecipePage.varAlertDuplicateInfo;
+
+        const nightmare = new Nightmare();
+        nightmare
+          .goto(recipesPage.url)
+          .wait(recipesPage.btnCreateRecipe)
+          .click(recipesPage.btnCreateRecipe)
+          .wait(page => document.querySelector(page.dialogRootElement).style.display === 'block'
+            , createRecipePage)
+          .insert(createRecipePage.inputName, pageConfig.recipe.simple.name)
```
henrywang: Could you please try .type instead of .insert. Thanks!

jgiardino:
Thanks for making updates! There are a couple of issues I see with the latest commit...

**Issue with when the following message displays:**
>A recipe name is required.

**Steps to reproduce:**

1. open the Create Recipe dialog
2. enter a character into the name field then delete it
**actual result:** the message will display, even though the field has focus
**expected result:** the message should not display until the user exits the field or clicks Save

Ideally, we wouldn't display error messages to the user until the error condition is true. In the case where a required field is empty, the right time to tell the user that it's required is when they proceed to the next field in the form or try to submit.

**Issue with when the following message displays:**
>The name "example" already exists.

**Steps to reproduce:**

1. open the Create Recipe dialog
2. enter a name that already exists so that the error message displays
3. close the dialog then open the dialog again
**actual result:** the value is still visible but the message is not
**expected result:** the message should be visible as long as the condition is true.

With the "required" error message, if it displayed and the user exits the modal then opens it again, we don't want the message to display, because we now assume the user is starting over in the form and we shouldn't display any "required" error messages when the user has not yet taken any action in the form.

But for this message, the use case is different. The previous error is based on a user action, whereas this error is based on a form input value. 

So if a value is entered in the recipe name field, and the user closes the dialog, opens it and still sees the value in the field, AND if this name already exists, then that error condition is still true--it doesn't change when the user closes and reopens the dialog.

components/Modal/CreateRecipe.js
```diff
@@ -40,16 +42,22 @@ class CreateRecipe extends React.Component {
     const o = Object.assign({}, this.state.recipe);
     o[prop] = e.target.value;
     this.setState({ recipe: o });
+    if (prop === 'name') {
+      this.dismissErrors();
+      this.handleErrors(e.target.value);
```
jgiardino: The issue related to displaying the required field message on a field that has focus is still present because `this.handleErrors(e.target.value);` is also calling `this.handleErrorName(recipeName);`.  Should this line be replaced with just `this.handleErrorDuplicate(e.target.value);`?

jgiardino:
The second issue is resolved but the first one isn't. See my inline comments for a suggestion.

components/Modal/CreateRecipe.js
jkozol: I have added this recommendation to the most recent commit. It appears to fix the issue.

jgiardino:
I see the fix works as expected. Thanks!

I didn't realize this until just now, but each time a character is entered, an api call for the recipe list is made. Should we just make this api call once? Or should the list just be passed to the modal as a prop? 

I understand there is a case where another user creates a recipe using the same name while the current user has this dialog open and is typing. But the current implementation doesn't fully handle this case (i.e. if user 1 and user 2 type the same name at the same time, and user 2 hits save first, what would happen with the current implementation?)

Getting the recipe list on each key press seems excessive, and I wonder if we should get it once on opening the dialog (or even just passed via props)? But regardless, we should definitely get the recipe list on save to double-check the name hasn't been used in between the time the user exited the name field and clicked Save.

jkozol:
The test for a duplicate recipe name has been dropped temporarily. The changes to the modal will be migrated to a new PR (114) and this PR will be closed.

PR #114, opened 2017-06-30
Added a duplicate recipe name alert to the CreateRecipe modal.

jkozol:
When creating a new recipe there will now be an alert if one tries to create a recipe with a duplicate name. This is a continuation of PR 113.

### DIFF
--- a/test/end-to-end/test/test_createRecipe.js
+++ b/test/end-to-end/test/test_createRecipe.js
@@ -20,22 +20,82 @@ describe('Create Recipe Page', () => {
 
   describe('Input Data Validation Test', () => {
     describe('Required Field Missing #acceptance', () => {
-      test('should show alert message when create recipe without name @create-recipe-page', (done) => {
+      test('should show alert message by clicking Save button when create recipe without name @create-recipe-page', (done) => {
         // Highlight the expected result
-        const expected = createRecipePage.varAlertMissingInfo;
+        const expectedAlertInfo = createRecipePage.varAlertInfo;
+        const expectedHelpBlockMsg = createRecipePage.varHelpBlockMsg;
 
         const nightmare = new Nightmare();
         nightmare
           .goto(recipesPage.url)
           .wait(recipesPage.btnCreateRecipe)
           .click(recipesPage.btnCreateRecipe)
-          .wait(page => document.querySelector(page.dialogRootElement).style.display === 'block'
+          .wait(page => document.activeElement.id === page.inputNameEleId
             , createRecipePage)
           .insert(createRecipePage.inputName, createRecipePage.varEmptyName)
           .wait(createRecipePage.btnSave)
           .click(createRecipePage.btnSave)
           .wait(createRecipePage.labelAlertInfo)
           .evaluate(page => document.querySelector(page.labelAlertInfo).innerText
+            , createRecipePage)
+          .then((element) => {
+            expect(element).toBe(expectedAlertInfo);
+          })
+          .then(() => nightmare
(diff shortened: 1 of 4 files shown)


jgiardino:
This is consistent with the implementation in PR #113. I created issue #115 to capture the comments made in #113. We can merge and close this PR, but will need to revisit that issue in the future.

henrywang:
Here's my analysis and solution.

Analysis:
The Save button receives the mouse down event but since the element(button) moves immediately due to onBlur after then mouse up happens outside and **no** onClick is registered. The onClick does not get triggered.

Solution: (This's only a demo to resolve this issue)

```javascript
  shouldComponentUpdate() {
    if (this.mouseIsDownOnSave) {
      return false;
    }
    return true;
  }

  mouseDownHandler = () => {
    this.mouseIsDownOnSave = true;
  }

  mouseUpHandler = () => {
    this.mouseIsDownOnSave = false;
  }

 <button
    type="button"
    className="btn btn-primary"
    onMouseDown={this.mouseDownHandler}
    onMouseUp={this.mouseUpHandler}
    onClick={(e) => this.showInlineError(e)}
 >Save</button>
```

jkozol:
After looking over earlier commits I believe that this error was already accounted for by having a timeout in the handleErrorName() call. I have added this timeout back in and believe it fixes this issue. If not let me know.

jgiardino:
Good catch, @henrywang!

I just reviewed the latest commit and I still see the issue as noted above.

Currently I only see the message "Required information is missing" display when I:

1. Open the Create Recipe modal, exit the name field, then click Save
2. Open the Create Recipe modal, then hit the Enter key

But if the Name field has focus and I click the Save button, only the field-level message displays.

You can test these interactions in the [demo site](http://devel.weld.lab.eng.bos.redhat.com/#/recipes) for an easy way to compare what you have in your local build to what was there before.

jgiardino:
I was not actually reviewing the latest commit in my previous comment, so the issue noted above is not present.

However, I did notice another issue that occurs when I complete the following steps:

1. Click Create Recipe button
2. Hit Enter key on keyboard

**Current result:** 404 error message displays
**Expected result:** the error messages that display on clicking Save (top of form and field-level) should also display on hitting Enter key when the modal is displayed

henrywang:
@jgiardino You're right. I missed this check point. This issue is able to be reproduced in my local env. I'll add a case to cover this scenario.

jgiardino:
Just reviewed and it works as expected.
@henrywang Can you please review one more time?

Thanks!

jgiardino:
BTW, there is another issue I had noticed related to exiting the dialog. I captured that in #117

test/end-to-end/test/test_createRecipe.js
```diff
@@ -32,6 +32,7 @@ describe('Create Recipe Page', () => {
           .click(recipesPage.btnCreateRecipe)
           .wait(page => document.activeElement.id === page.inputNameEleId
             , createRecipePage)
+          .insert(createRecipePage.inputName, createRecipePage.varEmptyName)
```
henrywang: Hi @jkozol, there're a question about this line of code.
Why do we need to type a empty string in Name input? We don't need to this action for this test case. The reason that I setup this case is to simulate a scenario that a user clicks Save button accidentally without any input.
The change is able to pass this case even though the case does not include this line of code.

henrywang:
Please pay attention to the code conflicts. Thanks.

jkozol:
Hi. You are correct that line was written before some of the recent changes and should no longer be necessary.

PR #116, opened 2017-07-11
Fix issue and add new case

henrywang:
1. Fix eslint eslint-config-airbnb version compatibility issue.
2. Update Create Recipe case to check Help-Block message.
3. Add Create Recipe case to support Enter key instead of Save button.
4. Add Create Recipe case to verify alert message when changing input focus.

### DIFF
--- a/test/end-to-end/package.json
+++ b/test/end-to-end/package.json
@@ -14,10 +14,10 @@
   },
   "devDependencies": {
     "eslint": "^3.19.0",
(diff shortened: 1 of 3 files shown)


henrywang:
[Add end to end test case to verify alert message on different scenarios](https://trello.com/c/Oxi9JsjR/364-add-end-to-end-test-case-to-verify-alert-message-on-different-scenarios)

PR #119, opened 2017-07-17
Add new unit test case and some improvements.

henrywang:
1. Add unit test case for CreateComposition.js
2. Update Jest to 20.x.x to support toHaveLength and more mock features
3. Remove sinon because Jest has full mock support now

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ available at [http://localhost:3000/](http://localhost:3000/).
 
 #### Unit Test
 
-The unit tests are powered by [Jest](https://facebook.github.io/jest/), [Enzyme](http://airbnb.io/enzyme/) and [Sinon.JS](http://sinonjs.org/)
+The unit tests are powered by [Jest](https://facebook.github.io/jest/) and [Enzyme](http://airbnb.io/enzyme/)
 
 ```shell
 $ npm run lint                  # Check JavaScript and CSS code for potential issues
--- a/package.json
+++ b/package.json
@@ -50,11 +50,12 @@
     "eslint-plugin-jest": "^20.0.3",
     "eslint-plugin-jsx-a11y": "^2.2.3",
     "eslint-plugin-react": "^5.2.2",
+    "faker": "^4.1.0",
     "file-loader": "^0.9.0",
     "firebase-tools": "^3.0.4",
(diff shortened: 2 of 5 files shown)


henrywang:
[Add unit test for Create Composition component](https://trello.com/c/tNNhGO36/375-add-unit-test-for-create-composition-component)

henrywang:
[Add unit test for Create Composition component](https://trello.com/c/tNNhGO36/375-add-unit-test-for-create-composition-component)

atodorov:
LGTM, however I'm not certain what benefit do React component unit tests have. From what I can see we're instantiating the component, checking if rendering didn't fail (not much else) and validating that some methods were executed. There's hardly any logic in these components (correct me if I'm wrong) so it looks to me that these unit tests are pretty much duplicating what e2e tests do. 

@jgiardino what is your opinion on this ?

@henrywang what is the current state of the art wrt testing React applications? Can you provide some information as to what other people are doing and in which scenarios?

jgiardino:
@atodorov @henrywang 

I chatted with Patrick Riley on the UXD team about this. The main purpose is to test the logic embedded within the component, but also test that the data provided by the api isn't changing. But for testing the api, are these tests set up so that we would catch changes in the api that would break the components?

Here are additional comments from Patrick on the subject:
> Snapshots are meant to quickly analyze React Component APIs (and would be included in e2e tests)… however they are much easier to maintain and quickly give us feedback when something changes… if you are already doing e2e work, then Snapshots may not be necessary, but I still find them quite useful for more fine grained API testing (you may find that easier but will have to experiment there)

atodorov:
@jgiardino, @henrywang 

I also did some research and I've changed my mind about unit testing React components. Honestly I didn't think we needed it but it's best to have it: 

1) Even if our components don't have any logic now sometime later they will and it is better to have a culture of testing and having everything already established.

2) +1 for what Jenn says, we need tests to make sure the API returns json with all the attributes we expect and break if the response changes. However I think there are easier ways to do this (like record the response history and compare). I will open a card for that.

I think we need to start a "testing guidelines" section in the readme.


With this said I'm going to go ahead and merge this PR so we can have at least some unit tests available. Everything else should be added as a new PR.

PR #121, opened 2017-07-20
Fix welder-web container repo issue.

henrywang:
Check out welder-web code not from master but from local which is checked out by Travis based on PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,9 @@ end-to-end-test:
 
 	sudo docker build -f ./test/end-to-end/Dockerfile -t weld/end-to-end:latest ./test/end-to-end/
 
+	git clone ../welder-web welder-deployment/welder-web
 	make -C welder-deployment/ repos
+	
 	if [ ! -d ./welder-deployment/mddb ]; then \
 		mkdir ./welder-deployment/mddb; \
 	fi; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,6 +14,10 @@ end-to-end-test:
 	sudo docker build -f ./test/end-to-end/Dockerfile -t weld/end-to-end:latest ./test/end-to-end/
 
 	make -C welder-deployment/ repos
+	git clone ../welder-web
```
atodorov:
```
cp -r . /tmp/welder-web
make -C welder-deployment/ repos
cp -r /tmp/welder-web repos/
```

henrywang: welder-web has to be in welder-deployment folder according to the docker-compose.yml setup.
And the there's no repos folder.
the git clone ../welder-web is to clone repos in current folder.

atodorov: why not instead of all these commands `git clone ../welder-web welder-deployment/welder-web` ?

henrywang: make repos will clone welder-web code from master into welder-deployment, that'll replace the repo we just cloned.
so we must make repos first, and delete welder-web repo which is cloned from master.
then, clone welder-web from local folder which contains the PR code.

atodorov: The command sequence you need is:

git clone https://github.com/weldr/welder-deployment - clones the welder-deployment repo under ./welder-deployment
git clone ../welder-web welder-deployment/welder-web - clones (aka copies) the current checkout under welder-deployment/welder-web
make -C welder-deployment/ repos - will git clone bdcs and bdcs-api-rs repositories. Will not clone welder-web because the directory already exists.

PR #123, opened 2017-07-25
html/css updates for independent scrolling

jgiardino:
These updates include the following: 

- ability to separately scroll the the list of inputs and the list of recipe components
  - additionally, this required changing the position of the tooltips that display in the inputs panel, since contents in one panel cannot overlap another panel
- consistent html/css for headers that display in the Recipe page, Recipe Edit page, and the component details
- component details display as a panel with a panel title on both the Recipe page and Recipe Edit page
- updated styles for the form that displays when viewing component details

### DIFF
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -149,33 +149,28 @@ class ComponentDetailsView extends React.Component {
     const { component } = this.props;
 
     return (
-      <div className="cmpsr-compon-details">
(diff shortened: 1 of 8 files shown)


jgiardino:
[(5) [UI dev] Edit Recipe page scrolling](https://trello.com/c/5zlpc4K2/368-5-ui-dev-edit-recipe-page-scrolling)

henrywang:
The badge indicating number of dependencies cannot update the number when component is added and removed.

PR #127, opened 2017-08-01
Recipe page in redux. Actions, reducers, and sagas for editRecipe page.

jkozol:
I have migrated the recipe page to redux. It's state is now in the store and it makes API calls through saga. I have also included the new reducers, actions, and sagas for the editRecipe page in preparation for that PR. Would it be possible for you to take a look at the recipe page and see if anything is broken. Also, the tests are failing and I can't figure out why. I will push a fix as soon as I find one but if you know what I am doing wrong that would really help me.

### DIFF
--- a/core/store.js
+++ b/core/store.js
@@ -23,8 +23,7 @@ const store = createStore(
   ),
 );
 sagaMiddleware.run(rootSaga);
-persistStore(store, { whitelist: ['recipes', 'recipePage'] }, () => {
-  this.setState({ rehydrated: true });
-});
+
+persistStore(store, { whitelist: ['recipes', 'recipePage'] });
 
 export default store;
--- a/core/store.js
+++ b/core/store.js
@@ -1,8 +1,8 @@
 import { createStore, applyMiddleware, compose } from 'redux';
 import createSagaMiddleware from 'redux-saga';
+import { persistStore, autoRehydrate } from 'redux-persist';
 import rootReducer from './reducers/index';
 import rootSaga from './sagas/index';
-
 
 const sagaMiddleware = createSagaMiddleware();
 
@@ -10,10 +10,21 @@ const sagaMiddleware = createSagaMiddleware();
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 /* eslint-enable */
 
+const initialState = {
+  rehydrated: false,
+};
+
 const store = createStore(
   rootReducer,
-  composeEnhancers(applyMiddleware(sagaMiddleware)),
+  initialState,
+  composeEnhancers(
+    applyMiddleware(sagaMiddleware),
+    autoRehydrate(),
+  ),
 );
 sagaMiddleware.run(rootSaga);
+persistStore(store, { whitelist: ['recipes', 'recipePage'] }, () => {
+  this.setState({ rehydrated: true });
+});
 
 export default store;
--- a/core/actions/inputs.js
+++ b/core/actions/inputs.js
@@ -1,0 +1,74 @@
+// gets picked up by recipes saga
+export const FETCHING_INPUTS = 'FETCHING_INPUTS';
+export const fetchingInputs = (filter, selectedInputPage, pageSize) => ({
+  type: FETCHING_INPUTS,
+  payload: {
+    filter,
+    selectedInputPage,
+    pageSize,
+  },
+});
+
+export const FETCHING_INPUTS_SUCCEEDED = 'FETCHING_INPUTS_SUCCEEDED';
+export const fetchingInputsSucceeded = (filter, selectedInputPage, pageSize, inputComponents) => ({
+  type: FETCHING_INPUTS_SUCCEEDED,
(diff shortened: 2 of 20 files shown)


jgiardino:
This is great! I just took a look at the Recipe page, and the UI looks as expected, with exception of the following note about tabs. 

Also, I noticed that there's an action SET_ACTIVE_TAB for the primary tabs, but not for the secondary tabs "Selected Components" and "Dependencies" on the "Components" tab. Is there a reason for that?

Regarding the issue with tabs, we had discussed an issue that was only visible on FF. I created issue #129 to capture this issue, and then submitted PR #130 as a fix just now. I'm currently seeing this same issue while reviewing this PR in Chrome. Just something to check after these PRs are merged, I think.

core/store.js
```diff
@@ -1,19 +1,30 @@
...
-  composeEnhancers(applyMiddleware(sagaMiddleware)),
+  initialState,
+  composeEnhancers(
+    applyMiddleware(sagaMiddleware),
+    autoRehydrate(),
+  ),
 );
 sagaMiddleware.run(rootSaga);
+persistStore(store, { whitelist: ['recipes', 'recipePage'] }, () => {
+  this.setState({ rehydrated: true });
```
henrywang: This line of code will not work here. "this" here will be undefined.

henrywang:
The browser console report the following error when go to recipe page.
> Warning: Failed prop type: Invalid prop `recipePage` of type `array` supplied to `RecipePage`, expected `object`.
    in RecipePage (created by Connect(RecipePage))
    in Connect(RecipePage)
    in Provider

This error is caused by init value of recipePage. The init value of recipePage is an empty array.

henrywang:
When access a recipe page at the first time by following a recipe name, the Details tab will be the first active tab. Then active the Components tab automatically.
That causes the old Provider node was removed and a new Provider node was generated.

deleted node

henrywang:
On view recipe page and when Components tab is active, click Create Composition button. Composition is still at Creating status on toast notification. It cannot go to Complete status.
creating composition

henrywang:
Refresh one of view recipe page, the page will go into a wrong page layout.
If go to this view recipe page by click recipe name from recipe list, no this kind of issue happened.
wrong view recipe page

henrywang:
Took too much time on fetch /api/v0/compose/types (5s) and /api/v0/projects/info/<projects>(10s). That why some of case failed by timeout.
too much time on fetch

jgiardino:
Regarding this comment from @henrywang:

> Refresh one of view recipe page, the page will go into a wrong page layout.
If go to this view recipe page by click recipe name from recipe list, no this kind of issue happened.
wrong view recipe page
wrong view recipe page

I think this is a separate issue that should be handled in another PR. I create issue #129 to capture a similar issue that was only happening in FF prior to this PR. I then submitted PR #130 (still open) as a fix. It's possible this issue will be resolved when #130 is merged, so I'd suggest we test for this issue again after both PRs are merged

henrywang:
I agree with you @jgiardino.

henrywang:
I agree with you @jgiardino. I'll keep track the PR merge and give this issue a test when everything is ready.

jgiardino:
Regarding this comment from @henrywang:

> On view recipe page and when Components tab is active, click Create Composition button. Composition is still at Creating status on toast notification. It cannot go to Complete status.

I'm not able to reproduce this issue. Additionally, notifications are not currently included in the redux store but eventually should be added. Should we wait to test notifications until then?

jgiardino:
Thanks for reviewing this, @henrywang!
bitmoji

henrywang:
The toast notification issue cannot be reproduced every time. I'm trying to find a common step to reproduce, not find one so far.

core/store.js
jkozol: I will remove this and replace it with a reducer.

jkozol:
Thanks for reviewing this @henrywang and @jgiardino. I will change the store to no longer setState and instead just rely on a reducer. I will also fix the prop types. In regards to:

> When access a recipe page at the first time by following a recipe name, the Details tab will be the first active tab. Then active the Components tab automatically.
That causes the old Provider node was removed and a new Provider node was generated.

Should we initially direct the user to the Components page or would it be fine to direct them to the Details tab?

In regards to:

> On view recipe page and when Components tab is active, click Create Composition button. Composition is still at Creating status on toast notification. It cannot go to Complete status.

I have been able to reproduce this but not reliably. I will look into the toast notifications and see what causes this to occur. If I had to guess it would be due to the slow fetches as pointed out by:

> Took too much time on fetch /api/v0/compose/types (5s) and /api/v0/projects/info/(10s). That why some of case failed by timeout.

jkozol:
I believe that the Provider node being deleted is causing some issues. When I go backwards and forwards between the recipes page and the recipe page too quickly I get sent to the Error page. I am looking into what causes this.

errorrecipepage

jkozol:
This is being closed because it is covered by PR 135. Some of the issues brought up such as the Provider have been fixed.

PR #125, opened 2017-07-31
Optimize environment setup script to speed up test time.

henrywang:
1. Enable docker cache to speed up container build time.

PR #132, opened 2017-08-04
Try to optimize e2e tests by re-using images from Docker hub

atodorov:
This is an updated version of #125 which pulls the latest images from Docker hub and tries to reuse them as cache. 

NOTE: as in my PR for bdcs-api-rs here we see the npm install happening again inside the image although there's no need for that. I'm not quite certain what causes this but I think it will be resolved once the changes get merged to master and we start pushing to Hub from CI.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+test
+metadata.db
+.travis.yml
+Makefile
+node_modules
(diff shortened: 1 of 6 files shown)


atodorov:
[Optimize welder-web e2e build and execution times](https://trello.com/c/g4qQmiIa/397-optimize-welder-web-e2e-build-and-execution-times)

PR #133, opened 2017-08-04
Cockpit Integration and make rpm functionality

jkozol:
Added the ability to easily make an rpm/srpm that will install into cockpit.

### DIFF
--- a/welder-web.spec
+++ b/welder-web.spec
@@ -1,0 +1,25 @@
+Name: welder-web
+Version: 0
+Release: 0
+Summary: Welder UI
+License: MIT
+
+Source: welder-web.tar.gz
+BuildArch: noarch
+
+%define debug_package %{nil}
+
+%description
+Welder UI
+
+%prep
+
+%build
+
+%install
+mkdir -p %{buildroot}
+tar --strip-components=1 -xzf %{sources} -C %{buildroot}
+find %{buildroot} -type f >> files.list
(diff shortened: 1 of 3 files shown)


welder-web.spec
```diff
@@ -0,0 +1,25 @@
+Name: welder-web
+Version: 0
+Release: 0
```
atodorov: LGTM but I'd like to have the version read from the same place where the web code reads it, not duplicate the information inside the spec file.

jkozol: Okay. I am looking into how to do this in a manner compatible with copr. You want it to set the version based on the version in the package.json correct? Hopefully I'll have a fix sometime this week.

atodorov: > You want it to set the version based on the version in the package.json correct

Yes. Is it possible to just execute package.json with nodejs and pass something like --version?

jkozol: It is possible to set the version of package.json with `npm version 1.1.1` but I'm not sure how this would help with importing it into the spec file.

shaded-enmity: Why not use Tito to take care of this, since you probably want to publish changelogs and tag releases as well.
AFAIK RPM doesn't allow you to set the Version tag dynamically so the value has to be changed externally.

atodorov:
Merging so we can move ahead. .spec vs. npm version will be fixed separately.

PR #135, opened 2017-08-10
Final Redux Integration

jkozol:
I have integrated redux into the recipe page, editRecipe page, and modals. Many of the tests are failing so I am looking for some assistance in fixing them. When run locally, the recipes, createRecipe, and viewRecipe (excluding some of the exportRecipes) tests all pass. Some of the unit tests are outdated since the CreateComposition modal no longer handles its own state. How would you recommend refactoring these? Also, if you notice any functionality that I failed to integrate into redux please let me know. One of the areas that is performing questionably is the persistence. Fixes will be appended to this PR over the next few days as I continue to debug it.


jgiardino:
I was able to reproduce the issue that @henrywang describes. 

When I'm on the Edit Recipe page, and hit refresh, it appears that `FETCHING_RECIPES_SUCCEEDED` runs for each recipe (which seems a little odd for this page), but not actually `FETCHING_RECIPE_CONTENTS`. This only appears to run when I navigate to the page from another page.

Also, this can be included in a later PR for changing sort order, but when I navigate to the Edit Recipe page, all recipes are stored in the store, and the order is modified so that the recipe that I'm viewing on the Edit Recipe page is listed as the last recipe in the 'recipes' array in the store. If I then navigate back to the Recipes page, the sort order is unchanged in this array, with the recipe I had open listed out of order as the last recipe on the page.

jkozol:
I have fixed this functionality. It appears that calling yield* fetchRecipes() in the recipes saga was causing all of the following yield calls to be ignored. I fixed this by moving the yield* call to the bottom of the  default recipes saga. I also updated the fetchingRecipesSucceeded reducer case so that it does not overwrite the  fetchingRecipesContentsSucceeded reducer case. At this point all tests excluding the editRecipe tests should pass.

jkozol:
I have updated the unit tests. I removed the tests for CreateComposition's componentWillMount() and getCompTypes() since these no longer exist.

henrywang:
I went through the failed cases, all of them are caused by the issue mentioned by me above.
Because all of cases for edit recipe page are waiting for components fully loaded with a command
```javascript
.wait(editRecipePage.componentListItemRootElement)
```
If there's no any selected component loaded here, the case will wait until timeout.

jkozol:
So it appears that the list of input components do not load for nightmare. I have been unable to replicate this particular bug in my browser, however. Are you able to go to a recipe's editRecipe page and have the input components fail to load? Also, all of the tests pass without that wait except for the composition toast notification test.

henrywang:
Hi @jkozol, I found there's no this API fetch in e2e test.That caused empty Available Components.
screenshot_20170815_183554

jkozol:
If we had a version of the recipe already in our state the RecipeApi wouldn't actually call the api so I removed that. Also, if I loaded directly into the editRecipe page after clearing my cache the inputs wouldn't load. Finally, I made a minor fix to the props we are passing into CreateComposition from the editRecipe page. The tests all pass now except the save recipe test. I am unsure why that test fails since it passes when run locally.

henrywang:
@jgiardino, feel free to merge it if you don't have any comment. :100:

jgiardino:
This is amazing, @jkozol! Thanks so much for working on this!

PR #136, opened 2017-08-11
Enable end-to-end test coverage

atodorov:
This is all the works necessary to support this. See the commit logs for more details but it boils down to:

1) Instrument the application code (main.js, pages, etc) with babel-plugin-istanbul. Only do this if --with-coverage has been passed on the command line otherwise we mess up unit tests (more precisely coverage from them)

2) In e2e tests make a slight modification how we chain and terminate Nightmare calls. This is necessary b/c we have to bring that coverage information from the browser scope (the page under test) to nodejs scope (the code doing the testing) and finally close the electron instance running the application.

**WARNING**: the above requires a bit of change in the way tests are written, in particular don't call `.end()` and `done()` but instead call the coverage handling code inside the last `.then()` method after all assertions have been executed. This is described in README but fyi <--- @jgiardino @henrywang @jkozol 

**NOTE** I do use a soft of ugly and hackish `eval()` to execute my coverage helper. I've tried splitting this out into a proper module and instead calling a function from the module but for some reason it doesn't work. I've asked around several experienced nodejs developers and they couldn't figure it out. It works for now and is only a one line, which is easy to copy&paste. If necessary we can change it in the future, provided somebody figures out how. Honestly I've spent too much time on this just trying to figure out why it doesn't work when the code is split into a helper function instead of copying the same code into the last `.then()` method body.

3) e2e coverage reports are stored under `.nyc_output` on the host and the nyc tool can be used to generate the reports.

4) Finally I've switched from Coveralls to CodeCov in order to be able to unify the two coverage reports. This is the easiest way given that Travis executes everything in a separate environment and doesn't allow sharing of artifacts between them

You can see an example report against my fork at:
https://codecov.io/gh/atodorov/welder-web/commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,13 @@ cache:
   - node_modules # NPM packages
 before_install:
   - |
+        npm install coveralls
+        npm install nyc
+
         if [ "$COMMAND" == "end-to-end-test" ]; then
             sudo docker pull welder/web-nodejs:latest
             sudo docker pull welder/web:latest
+            sudo docker pull welder/web-with-coverage:latest
             sudo docker pull welder/web-e2e-tests:latest
         else
             npm install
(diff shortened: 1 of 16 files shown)


atodorov:
[Produce coverage from welder-web e2e tests](https://trello.com/c/HKHpt1Nb/326-produce-coverage-from-welder-web-e2e-tests)

.travis.yml
```diff
@@ -10,9 +10,13 @@ cache:
   - node_modules # NPM packages
 before_install:
   - |
+        npm install codecov
+        npm install nyc --safe-dev
```
henrywang: npm install nyc --save-dev or just npm install nyc. No difference here.

henrywang:
:100: :+1:

test/end-to-end/README.md
```diff
@@ -128,6 +128,34 @@ describe('View Recipe Page', function () {             // Which page does this s
...
+and made available for the reporting tools to use. The helper code inside
+`utils/coverage.js` already to this and store the report files inder
+`/tmp/coverage-<HASH>.json`. The hash value is the sha256 sum of the coverage
+report itself.
+
+There are several things to be noted when writing and executing the tests:
+
+1. Some cases may have identical coverage so the number of json files will be
+   equal or less to the number of test cases;
+2. When constructing the Nightmare sequence makde sure you **DO NOT** call the
```
henrywang: make sure.

PR #137, opened 2017-08-27
Add end-to-end test against welder-web inside Cockpit

henrywang:


.travis.yml
```diff
@@ -18,9 +18,17 @@ before_install:
             sudo docker pull welder/web:latest
             sudo docker pull welder/web-with-coverage:latest
             sudo docker pull welder/web-e2e-tests:latest
-        else
+        fi
+install:
+  - |
+        if [ "$COMMAND" != "end-to-end-test" ]; then
             npm install
```
atodorov: This is not needed, the previous version in before_install was doing the same thing. Please revert.

.travis.yml
```diff
@@ -18,9 +18,17 @@ before_install:
...
         fi
+        if [ "$COMMAND" == "cockpit-test" ]; then
+            node run build --with-coverage
+            sed -i "s,var welderApiHost=.*,var welderApiHost=\"api\";," ./public/js/config.js
+            sudo docker pull welder/web-e2e-tests:latest
```
atodorov: Move the `docker pull` in before_install section.

Dockerfile.cockpit
```diff
@@ -0,0 +1,9 @@
+FROM fedora:latest
```
atodorov: Use `FROM welder/fedora:latest` this should already be downloaded as an intermediate layer of the e2e image. This way you don't have to pull the latest container of the upstream Fedora image.

test/end-to-end/entrypoint.sh
```diff
@@ -1,11 +1,17 @@
 #!/bin/bash
+#!/
```
atodorov: `/` is not an interpreter. Why do we need this change?

test/end-to-end/test/test_createRecipe.js
```diff
@@ -20,16 +22,28 @@ describe('Create Recipe Page', () => {
   const createRecipePage = new CreateRecipePage(pageConfig.recipe.simple.name
     , pageConfig.recipe.simple.description);
 
+// Switch to welder-web iframe if welder-web is integrated with Cockpit.
+// Cockpit web service is listening on TCP port 9090.
+  beforeEach(() => {
```
atodorov: I see that we have the same `beforeEach` block in several test files. Can you instead make it a helper function and only call the function in each test file?

test/end-to-end/utils/coverage.js
```diff
@@ -7,18 +7,35 @@
 // collects coverage from the tested page,
 // calls nightmare.end() and calls done()
 // executed via eval() as a last step of every test!
+if (pageConfig.root.includes('9090')) {
```
atodorov: I don't like this high-level if statement and then repeating the same code block twice.

test/end-to-end/utils/coverage.js
```diff
@@ -7,18 +7,35 @@
...
+  nightmare
+    .exitIFrame()
+    .evaluate(() => document.getElementsByClassName('container-frame')[0].contentWindow.__coverage__)
```
atodorov: IMO the check whether or not we are running inside Cockpit should be done inside the evaluate method and return the appropriate variable for coverage.

Questions:

1) From inside the iframe can we access the URL of the browser and check if we are running in Cockpit?
2) Why do we have to exit the iframe and then access its `.__coverage__` object from the parent? Is there any way to access this object even if we are inside the iframe (even if you have to access the parent from inside the iframe) ?

Makefile
```diff
@@ -58,3 +58,18 @@ end-to-end-test:
...
 	    xvfb-run -a -s '-screen 0 1024x768x24' npm run test -- --verbose
+
+cockpit-test:
+	sudo docker network create welder
+	sudo docker build -f Dockerfile.cockpit --cache-from welder/web-cockpit:latest -t welder/web-cockpit:latest .
+	sudo docker build -f ./test/end-to-end/Dockerfile --cache-from welder/web-e2e-tests:latest -t welder/web-e2e-tests:latest ./test/end-to-end/
+
+	wget --progress=dot:giga https://s3.amazonaws.com/weldr/metadata.db
+	sudo docker run -d --name api --restart=always -p 4000:4000 -v bdcs-recipes-volume:/bdcs-recipes -v `pwd`:/mddb --network welder --security-opt label=disable welder/bdcs-api-rs:latest
+	sudo docker run -d --name web --restart=always -p 9090:9090 -v `pwd`/public:/usr/share/cockpit/welder --network welder welder/web-cockpit:latest
```
atodorov: The above few lines are very similar to what the `ent-to-end` test target does and I don't like the repetition of setup steps. Instead can we move the setting up of api container, download of metadata, etc (all common steps) into a separate make target and have both e2e and cockpit tests depend on it. Then only setup whatever specific contaienrs are needed.

.travis.yml
atodorov: Also `docker pull welder/web-cockpit:latest` because this is very unlikely to change often and `docker pull` inside Travis is faster than `dnf install`.

.travis.yml
```diff
@@ -57,6 +66,17 @@ after_success:
             docker push welder/web-with-coverage
             docker push welder/web-e2e-tests
         fi
+  - |
+        if [ "$COMMAND" == "cockpit-test" ]; then
```
atodorov: Move above and make the if statement something like ` if [ "$COMMAND" == "end-to-end-test"  ] || [ "$COMMAND" == "cockpit-test"] ; then`

.travis.yml
```diff
@@ -57,6 +66,17 @@ after_success:
...
+            nyc report
+            nyc report --reporter=lcov | codecov
+        fi
+
+        if [ "$COMMAND" == "cockpit-test" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
```
atodorov: Same comment as above. Check `TRAVIS_BRANCH` & `TRAVIS_PULL_REQUEST` first; `docker login` and then based on which command we are executing `docker tag && push` the resulting images.

Makefile
atodorov: Also I think `docker run --rm` for web and api containers will not hurt here and better keep the test environment clean.

Makefile
```diff
@@ -58,3 +58,18 @@ end-to-end-test:
...
+
+	sudo docker run --rm --name welder_end_to_end --network welder \
+	    -v `pwd`/.nyc_output/:/tmp -v `pwd`:/mddb -e MDDB='/mddb/metadata.db' \
+	    -e COCKPIT_TEST=1 \
```
atodorov: If you move this into the helper as well you can use the value of a global variable to pass to docker.

.travis.yml
```diff
@@ -18,9 +18,17 @@ before_install:
...
-        else
+        fi
+install:
+  - |
+        if [ "$COMMAND" != "end-to-end-test" ]; then
             npm install
         fi
+        if [ "$COMMAND" == "cockpit-test" ]; then
+            node run build --with-coverage
+            sed -i "s,var welderApiHost=.*,var welderApiHost=\"api\";," ./public/js/config.js
```
atodorov: FYI I don't like this very much but I don't have another proposal ATM. It bugs me that for e2e testing we build the app inside the docker image and for testing with Cockpit we build the app outside and just mount it under the appropriate directory. However this can be resolved some other time.

test/end-to-end/utils/coverage.js
henrywang: To exit iframe first is window.__coverage__ cannot be fetched inside iframe. So I have to exit, then get coverage from global.

.travis.yml
henrywang: It's OK, the downside is there'll be an useless docker login for tests other than e2e and cockpit.

henrywang:
[[spike] UI e2e testing in Cockpit environment](https://trello.com/c/8lleezpR/331-spike-ui-e2e-testing-in-cockpit-environment)

.travis.yml
atodorov: yeah but simpler and cleaer code trumps that downside.

test/end-to-end/utils/coverage.js
atodorov: Can you get access to the parent page without exiting the iframe?

.travis.yml
```diff
@@ -18,9 +18,20 @@ before_install:
             sudo docker pull welder/web:latest
             sudo docker pull welder/web-with-coverage:latest
             sudo docker pull welder/web-e2e-tests:latest
-        else
```
atodorov: `npm install` inside the else block, we don't need a second if statement with a negative condition to do that.

Dockerfile.cockpit
```diff
@@ -0,0 +1,9 @@
+FROM welder:latest
```
atodorov: `welder/fedora:latest`

test/end-to-end/utils/coverage.js
henrywang: Accessing parent page works, but still cannot get __coverage__ value.

atodorov: if you can access the parent object can't you do something like `document.parent.getElementsByClassName('container-frame')[0].contentWindow.__coverage__` ? What does "Accessing parent page works, but still cannot get coverage value." mean? If you can access the parent object what is preventing you from accessing the coverage object as well ?

Dockerfile.cockpit
henrywang: The welder/fedora image as base image will cause login page issue. The page will be kept in login page.

test/end-to-end/utils/coverage.js
henrywang: That'll report `document.parent is undefined` error.

test/end-to-end/test/test_createRecipe.js
henrywang: Added a helper, and call helper inside beforeEach.

test/end-to-end/utils/coverage.js
atodorov: OK, don't take what I write verbatim. Show me how do you access the parent page from within the iframe ?

Dockerfile.cockpit
atodorov: There's no `welder:latest` image either. The format of these images is repo/image:tag. What is your intended base image  then ? Where does it come from?

atodorov: found the latest implementation. lgtm.

test/end-to-end/test/test_editRecipe.js
```diff
@@ -16,12 +19,18 @@ describe('Edit Recipe Page', () => {
     apiCall.serviceCheck(done);
   });
 
+  const editRecipePage = new EditRecipePage(pageConfig.recipe.simple.name);
+
+  // Switch to welder-web iframe if welder-web is integrated with Cockpit.
+  // Cockpit web service is listening on TCP port 9090.
+  beforeEach(() => {
+    helper.gotoURL(nightmare = new Nightmare(), editRecipePage);
```
atodorov: This comment applies everywhere where you call the helper method. As far as I can see everywhere you declare the `nightmare` variable and then call `helper.gotoURL` and create a new Nightmare object while at the same time passing this as parameter to the function. I think you can instead use something like:

```

beforeEach(() => {
    helper.gotoURL(nightmare, editRecipePage);
});
```

.travis.yml
```diff
@@ -18,9 +18,20 @@ before_install:
...
+        fi
+        if [ "$COMMAND" == "cockpit-test" ]; then
+            sudo docker pull welder/web-e2e-tests:latest
+            sudo docker pull welder/web-cockpit:latest
+        fi
+        if [ "$COMMAND" != "end-to-end-test" ]; then
```
atodorov: `npm install` inside the (deleted) else block above, we don't need a second if statement with a negative condition to do that.

test/end-to-end/test/test_editRecipe.js
henrywang: The nightmare instance has to be scoped inside each test. That means the nightmare will be destroy after each case. If nightmare instance locates in test suite scope, only the first case will be run without issue. The other cases after the first one will be failed because nightmare is not clear and inherit lots of settings from the first case. So nightmare has to be declared inside beforeEach().

atodorov: ok then. +1 on this

Makefile
henrywang: If you use -d with --rm, the container is removed when it exits or when the daemon exits, whichever happens first. In this case, both of api and web will not exit during testing.

.travis.yml
henrywang: updated.

.travis.yml
henrywang: updated.

Dockerfile.cockpit
henrywang: The cockpit image built from welder/fedora:latest does not work in my local env and Travis. All of cases are failed. In my local env, the welder-web is not able to be loaded in iframe. Travis result: https://travis-ci.org/henrywang/welder-web/jobs/270048364
Since the web-cockpit image is built from cache almost all the time, IMO, we just keep fedora:latest in this PR and add a Trello card to investigate the root cause.

atodorov: ok, leave it as-is for now and create a new card to tackle this later

test/end-to-end/test/test_createRecipe.js
atodorov: +1 on this

.travis.yml
```diff
@@ -30,12 +39,13 @@ env:
...
+    - COMMAND=cockpit-test
 script:
   - make "$COMMAND"
 after_success:
   - if [ "$COMMAND" == "unit-test" ]; then cat ./coverage/lcov.info | codecov; fi
   - |
-        if [ "$COMMAND" == "end-to-end-test" ]; then
+        if [ "$COMMAND" == "end-to-end-test" ] || [ "$COMMAND" == "cockpit-test" ]; then
             nyc report
             nyc report --reporter=lcov | codecov
```
atodorov: && codecov

.travis.yml
```diff
@@ -58,6 +68,11 @@ after_success:
             docker push welder/web-e2e-tests
         fi
 
+        if [ "$COMMAND" == "cockpit-test" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
```
atodorov:
```
if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
    docker login ...

    if [ "$COMMAND" == "end-to-end-test" ] ; then
       ... tag & push e2e related images
    fi

    if [ "$COMMAND" == "cockpit-test" ] ; then
       ... tag & push cockpit related images
    fi
```

Also note: docker push welder/web-with-coverage is not needed b/c the coverage image shouldn't be cached. This has been a mistake on my part.

Makefile
```diff
@@ -42,19 +42,29 @@ stylelint:
...
-	sudo docker build -f Dockerfile.with-coverage --cache-from welder/web-with-coverage:latest -t welder/web-with-coverage:latest .
 	sudo docker build -f ./test/end-to-end/Dockerfile --cache-from welder/web-e2e-tests:latest -t welder/web-e2e-tests:latest ./test/end-to-end/
-
+	sudo docker network create welder
 	wget --progress=dot:giga https://s3.amazonaws.com/weldr/metadata.db
 	sudo docker run -d --name api --restart=always -p 4000:4000 -v bdcs-recipes-volume:/bdcs-recipes -v `pwd`:/mddb --network welder --security-opt label=disable welder/bdcs-api-rs:latest
+
+end-to-end-test: shared
+	sudo docker build --cache-from welder/web:latest -t welder/web:latest .
+	sudo docker build -f Dockerfile.with-coverage --cache-from welder/web-with-coverage:latest -t welder/web-with-coverage:latest .
```
atodorov: Remove `--cache-from welder/web-with-coverage:latest` b/c this base image isn't downloaded in CI. This has been a mistake in the previous implementation (copy & paste programming).

PR #138, opened 2017-08-28
Fix coverage issue.

henrywang:
1. Fix coverage issue.
2. Resolve one case stability issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ after_success:
   - |
         if [ "$COMMAND" == "end-to-end-test" ]; then
             nyc report
-            nyc report --reporter=lcov | codecov
+            nyc report --reporter=lcov && codecov
         fi
 
         if [ "$COMMAND" == "end-to-end-test" ] && [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then


.travis.yml
```diff
@@ -37,7 +37,7 @@ after_success:
   - |
         if [ "$COMMAND" == "end-to-end-test" ]; then
             nyc report
-            nyc report --reporter=lcov | codecov
+            nyc report --reporter=lcov > coverage.lcov && codecov
```
atodorov: We need to know what the root cause of the problem is. As you can see above on line 36 `codecov` can easily read the lcov information from its stdin. It doesn't need to read that information from a file. And if you inspect the CI build logs for the various test jobs you can see that for the unit-test job it can happily read this information. Also codecov was able to happily read the lcov information from stdin prior to PR #135. 

The problem lies elsewhere and until we know where I don't feel comfortable accepting this patch. Things should either work or not work. Instead we have the case of "was working" and "stopped working without any related changes".

test/end-to-end/test/test_editRecipe.js
```diff
@@ -325,8 +325,7 @@ describe('Edit Recipe Page', () => {
         const toastNotifPage = new ToastNotifPage(pageConfig.recipe.simple.name);
```
atodorov: This is unrelated, move it out of here.

atodorov:
[welder-web: Merging PR#135 decreased e2e coverage](https://trello.com/c/dRXoVq6G/416-welder-web-merging-pr135-decreased-e2e-coverage)

test/end-to-end/test/test_editRecipe.js
henrywang: This case will fail sometimes, and there'll not be coverage result. So I put this improvement here.

.travis.yml
henrywang: This is the log from PR#136 which got a right coverage result.
```shell
==> Generating gcov reports (skip via --disable=gcov)
    $ find /home/travis/build/weldr/welder-web -type f -name '*.gcno'  -exec gcov  {} +
==> Scanning for reports
    + /home/travis/build/weldr/welder-web/coverage/lcov.info
==> Uploading reports
    Success!
```
Here's the correct log:
```shell
==> Generating gcov reports (skip via --disable=gcov)
    $ find /home/travis/build/weldr/welder-web -type f -name '*.gcno'  -exec gcov  {} +
==> Scanning for reports
    + /home/travis/build/weldr/welder-web/coverage.lcov
    + /home/travis/build/weldr/welder-web/coverage/lcov.info
==> Uploading reports
```
The only thing codecov did is scanning file which match the pattern and upload the matched result file into server. The code can be found from https://github.com/codecov/codecov-node/blob/master/lib/codecov.js
So even the stdin was passed to codecov, it'll ignore it and does scanning result file according to pattern. From the codecov.js, it only accepts the following types of files:
```shell
'*coverage.*' + 'nosetests.xml' + 'jacoco*.xml' + 'clover.xml' + 'report.xml' + 'cobertura.xml' +'luacov.report.out' + 'lcov.info' +'*.lcov' +'gcov.info' +'*.gcov' +'*.lst'
```
So the .json results has to be converted into one of above file type. Then the codecov will find it and upload it to server.

henrywang: For the Line 36, if removing `cat ./coverage/lcov.info | `, it still works. 
The only thing codecov did is scanning file which match the pattern and upload the matched result file into server. The codecov source code can be found from https://github.com/codecov/codecov-node/blob/master/lib/codecov.js
So even the stdin was passed to codecov, codecov'll ignore it and does scanning result file according to pattern. From the codecov.js, it only accepts the following types of files:
```shell
'*coverage.*' + 'nosetests.xml' + 'jacoco*.xml' + 'clover.xml' + 'report.xml' + 'cobertura.xml' +'luacov.report.out' + 'lcov.info' +'*.lcov' +'gcov.info' +'*.gcov' +'*.lst'
```

henrywang:
For the Line 36, if removing `cat ./coverage/lcov.info | `, it still works. 
The only thing codecov did is scanning file which match the pattern and upload the matched result file into server. The codecov source code can be found from https://github.com/codecov/codecov-node/blob/master/lib/codecov.js
The `nyc 

Here's the lcov.info from this PR: (`nyc report --reporter=lcov > coverage.lcov && codecov`)
```shell
# path=/home/travis/build/henrywang/welder-web/coverage.lcov

<<<<<< EOF
# path=/home/travis/build/henrywang/welder-web/coverage/lcov.info
TN:
SF:/welder/main.js
FN:16,renderComponent
```
From this lcov.info, on content in coverage.lcov, that proves the `nyc report --reporter=lcov` does not generate stdout. It just scan from ./nyc_output and generate coverage result into coverage folder.
From source code https://github.com/istanbuljs/nyc/blob/master/index.js, the ./nyc_output and coverage folders are default parameters.

henrywang:
The fix is remove pipe line, just leave them alone.
```shell
        if [ "$COMMAND" == "end-to-end-test" ]; then
            nyc report
            nyc report --reporter=lcov
            codecov
        fi
```

henrywang:
The issue for PR#135 is no lcov.info generated by ` nyc report --reporter=lcov | codecov`. But the coverage info is right, because the `nyc report` generated a correct output.
Here's the log:
```shell
==> Building file structure
==> Generating gcov reports (skip via --disable=gcov)
    $ find /home/travis/build/weldr/welder-web -type f -name '*.gcno'  -exec gcov  {} +
==> Scanning for reports
    X Failed to read file at 
==> Uploading reports
    Success!
```

atodorov:
OK, I think I know what happened. 
`nyc report --reporter=lcov | codecov` produced a lcov.info file and that completed before the empty output was sent down the pipe and codecov started. So codecov was able to find the lcov.info file and submit it. 

But after merging PR #135 this is not happening so maybe it's a race condition. Eitherway I agree the pipe isn't appropriate there so maybe something like `nyc report --reporter=lcov && codecov` instead?

test/end-to-end/test/test_editRecipe.js
atodorov: If this test fails then we don't get "some coverage" but it is not related to the problem of submitting the information to codecov. If you want to keep this change in this PR then at least make 2 separate commits for the two files.

henrywang:
Agree with you. Update to `nyc report --reporter=lcov && codecov`.

PR #140, opened 2017-09-20
Sorting PR

jkozol:
Recipes, components, and dependencies can be sorting by name in either ascending or descending order on the recipes page, recipe page, and editRecipe page. The tests all pass locally and I am unsure why they fail on travis.


jgiardino:
This is great! The sorting behaves as I would expect it to on each page. 

Since I don't know yet what other features are needed regarding filters and 'sort by' fields, I'll create separate work items to capture that as I have more details, whether it's to remove non-functional UI or to implement it.

henrywang:
Test got passed with new db file. @jgiardino feel free to merge it.

PR #142, opened 2017-09-22
Adjust Cockpit manifest order

martinpitt:
Increase order priority so that Welder does not appear as the first
thing in the top menu bar, even before Cockpit's own dashboard.

Also drop erroneous tab character.

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,13 +1,13 @@
 {
     "version": "135",
     "requires": {
-	"cockpit": "134"
+        "cockpit": "134"
     },
 
     "dashboard": {
         "index": {
             "label": "Welder",
-            "order": 8
+            "order": 30
         }
     },
     "content-security-policy": "default-src 'self' code.jquery.com maxcdn.bootstrapcdn.com  cdnjs.cloudflare.com 'unsafe-inline' 'unsafe-eval'"


martinpitt:
At some point we also need to decide about the "Welder" name, which doesn't appear anywhere else in the UI; maybe change it to "Composer"?

martinpitt:
This test failure is certainly unrelated to my change here:
```
 FAIL  test/test_editRecipe.js (64.333s)
  ● Edit Recipe Page › Single Word Recipe Name Scenario › Save Recipe Test #acceptance › should have toast notification pop up when Save button clicked @edit-recipe-page

    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
```
Is that "expected"? Can the test be retried?

martinpitt:
This happened a second time, thus for some curious reason this test is influenced by the order of the dashboards. I'll need to investigate this, setting as "work in progress" for now.

martinpitt:
I actually do get a crash when I try to save a recipe, I filed this as issue #145. Just not sure why this doesn't happen in my other PRs.

martinpitt:
I can't reproduce this crash locally, and it indeed seems to be a flake - a test retry (via force push) now worked.

PR #143, opened 2017-09-22
Drop references to external code

martinpitt:
Pull in jQuery, jquery.matchHeight, bootstrap, and bootstrap-select via
npm and webpack instead of external URLs. Drop these from
Content-Security-Policy.

---

Official cockpit modules must have a "strict" C-S-P without external code and also without `unsafe-inline` and `unsafe-eval`. This PR gets rid of the former.

I kept jQuery as separate .js files, I think (for some reason, not sure why) that is a common pattern. However, bootstrap should be fine to get pulled in via `import` and bundled via webpack (so it ends up in the bundled main.js), and in my manual tests that seems to work.

### DIFF
--- a/main.js
+++ b/main.js
@@ -5,6 +5,8 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import FastClick from 'fastclick';
 import { Provider } from 'react-redux';
+import 'bootstrap';
+import 'bootstrap-select';
 
 import store from './core/store';
 import router from './core/router';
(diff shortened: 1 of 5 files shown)


martinpitt:
@petervo, @stefwalter: I'd appreciate if you could review this, as my webpack skills are still very weak and I'm not sure about the jquery bundling.

jgiardino:
Thanks for making this update! Looks good to me!

PR #144, opened 2017-09-22
Drop unsafe-inline from Cockpit Content-Security-Policy

martinpitt:
There are on inline HTML `on*=""` event handlers and no `<script>` tags.
Inline `style=` was only being used in CardView.js where it does not
actually make any visual difference, so just drop that.

### DIFF
--- a/components/CardView/CardView.js
+++ b/components/CardView/CardView.js
@@ -30,7 +30,7 @@ class CardView extends React.Component {
         {users.map((user, i) =>
           <div className="col-xs-12 col-sm-6 col-md-4 col-lg-3" key={i}>
             <div className="card-pf card-pf-view card-pf-view-select card-pf-view-single-select">
-              <div className="card-pf-body" style={{ height: '260px' }}>
+              <div className="card-pf-body">
                 <div className="card-pf-top-element">
                   <span className="fa fa-birthday-cake card-pf-icon-circle"></span>
                 </div>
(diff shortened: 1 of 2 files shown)


martinpitt:
Keeping notes here.

I compared the test logs, and there are no further failures due to the unsafe-inline fix. So it's "only" the unsafe-eval removal which is causing trouble somehow. Unfortunately I can't currently reproduce this as trying to build docker containers causes my Fedora 26 to freeze, but I'll keep experimenting.

E. g. this failure:
```
 FAIL  test/test_viewRecipe.js (376.333s)^M
  ● View Recipe Page › Single Word Recipe Name Scenario › Menu Nav Bar Check #acceptance › should show a recipe name @view-recipe-page^M
```
This doesn't reproduce in interactive testing. `viewRecipePage.labelRecipeName` is `ol[class="breadcrumb"] li strong`, and that exists both in the Inspector and on screen:

breadcrumb

So I don't think dropping the `unsave-eval` actually broke the code, but it apparently does break the test. Apparently the `evals()` from test/end-to-end/test/test_*.js are running *inside* the cockpit welder module, not outside of the test container?

martinpitt:
I updated the PR to just drop `unsafe-inline` for now. `unsafe-eval` has to wait for either fixing or porting the tests.

martinpitt:
Closing this for now, this is too hard to fix with the current tests, and would most probably be in vain due to porting the tests.

PR #146, opened 2017-09-25
Fix running integration tests multiple times

martinpitt:
The tests don't clean up behind themselves very well and cannot currently be run more than once without manual steps. There are still running containers, a docker network,  and the db download file handling is broken.

This isn't perfect yet, but at least makes things better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ metadata.db:
 shared: metadata.db
 	sudo docker build -f Dockerfile.nodejs --cache-from welder/web-nodejs:latest -t welder/web-nodejs:latest .
 	sudo docker build -f ./test/end-to-end/Dockerfile --cache-from welder/web-e2e-tests:latest -t welder/web-e2e-tests:latest ./test/end-to-end/
-	sudo docker network create welder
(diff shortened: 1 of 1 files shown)


martinpitt:
```
sudo docker network rm welder
Error response from daemon: network welder has active endpoints
```

Need to clean up the running containers, too.

martinpitt:
Now `make end-to-end-test` (or `make cockpit-test`) actually works multiple times in a row.

jgiardino:
@henrywang and @atodorov - Will you please review and merge this PR if you have no suggestions?

Makefile
```diff
@@ -48,7 +48,7 @@ metadata.db:
 shared: metadata.db
 	sudo docker build -f Dockerfile.nodejs --cache-from welder/web-nodejs:latest -t welder/web-nodejs:latest .
 	sudo docker build -f ./test/end-to-end/Dockerfile --cache-from welder/web-e2e-tests:latest -t welder/web-e2e-tests:latest ./test/end-to-end/
-	sudo docker network create welder
+	sudo docker network inspect welder >/dev/null 2>&1 || sudo docker network create welder
```
atodorov: +1 this is exactly what I've been looking for

PR #147, opened 2017-09-25
Drop users page

martinpitt:
It is unrelated to recipes, and just duplicates the existing Unix users
page in Cockpit.


jgiardino:
Looks good to me!

PR #148, opened 2017-09-26
Replace Overview page with blank slate pattern on Recipes

martinpitt:
The Overview page is very empty and is mostly just tailored for
usability testing with pre-existing recipes. For real deployments with
no pre-existing recipes it's a more common workflow to put the
instructions as a "blank slate pattern" on the actual recipes page.

This works towards avoiding the navigation menu completely (together
with dropping the Users page).

Rename home.index.spec.js unit test to index.norecipes.spec.js and
verify that blank slate is generated with zero recipes.


martinpitt:
As discussed in yesterday's meeting. Currently the empty recipes page looks like this:
welder-empty-current

Wit this PR, it looks like this, and the separate "Overview" menu enty is gone:
welder-empty-this

We can/will surely adjust the text in the box further (that requires some designing), but this at least works towards cleaning up the navigation, together with PR #147.

jgiardino:
Thanks for the update! I'm good with finalizing the empty state message in another PR.

PR #149, opened 2017-09-26
Only install firebase-tools on demand

martinpitt:
The firebase-tools NPM module and its dependendies take over 230 MB, and
for normal building, development, and CI it's not necessary. Thus move
it out of package.json and install it on demand for "npm run publish".

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "eslint-plugin-redux-saga": "^0.4.0",
     "faker": "^4.1.0",
     "file-loader": "^0.9.0",
-    "firebase-tools": "^3.0.4",
     "front-matter": "^2.1.0",
     "highlight.js": "^9.5.0",
     "jest": "^20.0.4",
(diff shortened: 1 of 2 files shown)


martinpitt:
Before:
```
❱❱❱ du -hsc node_modules/
835M	node_modules/
```

After:
```
❱❱❱ du -hsc node_modules/
599M	node_modules/
```
I also verified that `npm run publish` correctly installs it.

jgiardino:
@henrywang will you review and merge if you have no objections?

henrywang:
Thanks @martinpitt. All changes work for me.
One more question to discuss, do we really need firebase in this case? @jgiardino @jkozol @martinpitt ?

martinpitt:
@henrywang, I'm not sure. I don't know what firebase even is. We can drop that part once we move the web UI to cockpit, but I didn't want to break existing functionality in this repo, so I rather left it intact.

henrywang:
@martinpitt, got you. Thanks.

PR #150, opened 2017-09-27
Drop nav bar, and hide header bar in Cockpit

martinpitt:
Now that the Overview and Users menu entries are gone, "Recipes" is the only one left and thus the navigation bar is currently unused. I left it in the code for now, but not show it. (We can clean up the unused code later).

Also, the header bar is redundant in Cockpit, so hide it there.

### DIFF
--- a/components/Layout/Header.js
+++ b/components/Layout/Header.js
@@ -4,12 +4,6 @@ function Header() {
   return (
     <nav className="navbar navbar-pf-vertical">
       <div className="navbar-header">
-        <button type="button" className="navbar-toggle">
-          <span className="sr-only">Toggle navigation</span>
-          <span className="icon-bar"></span>
-          <span className="icon-bar"></span>
-          <span className="icon-bar"></span>
-        </button>
         <a href="/" className="navbar-brand">
           <img className="navbar-brand-name" src="/logo-header.svg" alt="" width="220" />
         </a>
--- a/components/Layout/Layout.js
+++ b/components/Layout/Layout.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Header from './Header';
-import Navigation from './Navigation';
 import Notification from '../../components/Notifications/Notification';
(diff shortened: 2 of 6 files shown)


jgiardino:
Looks good. The only thing I notice is when running the app on it's own, the class `container-pf-nav-pf-vertical` results in a 200px empty margin where the navigation would normally display. Will that class be necessary in the context of Cockpit, or would the content margin for the navigation in Cockpit be handled in some other way?

martinpitt:
@jgiardino : Indeed, I failed to notice that. I guess maybe because it doesn't actually look that bad on a wide screen (otherwise the table becomes too wide  for comfortable reading), but it really gets in the way on narrow browser windows. If we want to limit the width of the table, we should do that with proper CSS.

I'll fix that, thanks for spotting! No, this class/free space for menu won't be necessary for cockpit either. There you will now also not see a vertical menu, the Cockpit "chrome" is just on top:

welder-chrome

martinpitt:
I dropped the `container-pf-nav-pf-vertical` classes everywhere, now it looks as it should. It's still a reasonably small patch, in case we have to revert it again when bringing back menus.

henrywang:
@martinpitt, There is an empty margin where the title bar would normally locate. Can we remove that empty margin when title bar is at hide status?
hide-title_bar

martinpitt:
@henrywang: You have a curious view - even if I uninstall cockpit-kubernetes and cockpit-dashboard (which were responsible for two of the horizontal header menu entries in my screenshot above), I still keep "Welder":

welder-header

How did you get rid of the "Welder" dashboard view?

That said, this is only tangential. In my view there is some empty space on top of the page too, just a little less than on your's, so here it still looks pleasant. But again, if we want that we should do that through explicit CSS instead of relying on empty padding for the hidden top menu indeed. I'll look into this.

martinpitt:
Fixed. Now it looks correct in Cockpit mode:
cockpit

... and in standalone mode:
standalone

Issue #151, opened 2017-09-27
core/sagas/inputs.js updateInputComponentData() is missing `yield`


martinpitt:
Cockpit has a slightly different eslint configuration, which picked up several issues. I can resolve most of them (will send a PR soon), but one is more tricky:
```
ERROR in ./pkg/welder/core/sagas/inputs.jsx

/home/martin/upstream/c-welder/pkg/welder/core/sagas/inputs.jsx
  5:1  error  This generator function does not have 'yield'  require-yield
```

And indeed [updateInputComponentData()](https://github.com/weldr/welder-web/blob/master/core/sagas/inputs.js#L5) does not call `yield` anywhere. I'm not deep enough into the code yet to be able to say what this does exactly, so could you have a look, please?

martinpitt:
@jkozol , can you please have a look at this and tell me how `updateInputComponentData()` is supposed to work? (sync or async, with yield or return). It's being called asynchronously from `fetchInputs()` right now, so I guess "async", but it's written in a sync way.

martinpitt:
Seems it was intended to be sync after all. This was fixed as a by-product in https://github.com/weldr/welder-web/commit/50630524afbc1c03eb06204a44240edceea45498#diff-9e47aee6cedcc9f80cf6432dc6e63aff and without also dropping the eslint magic comment. I'll send a PR to drop it to avoid confusion and regressions. Thanks!

PR #152, opened 2017-09-27
Enable recommended eslint checks and fix resulting errors

martinpitt:
This triggers a few errors for unused variables, trailing commas in
function arguments, but most importantly a missing `yield` in
core/sagas/inputs.js (see issue #151). Tone down that error into a
warning for the time being, until this gets fixed properly.

---
I stumbled over these when trying to build welder-web in Cockpit. That has these eslint errors on, so the build failed due to these.

### DIFF
--- a/core/constants.js
+++ b/core/constants.js
@@ -1,4 +1,3 @@
-/* global welderApiHost */
 /* *
   Application Constants
  */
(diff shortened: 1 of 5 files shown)


jgiardino:
@henrywang or @jkozol, do you have any comments for these changes?

PR #154, opened 2017-09-28
Lower required jquery version to 2.1+

martinpitt:
The dependency was added in 71a63b230d51 without a particular reason,
thus presumably to just whichever version was current at that time.
However, Cockpit currently uses (and exposes as part of its API) jQuery
2.x only, so lower the minimum version.

----

I went through all the pages, and everything seems to work normally. I also locally ran the tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "classnames": "^2.2.5",
     "fastclick": "^1.0.6",
     "history": "^3.0.0",
-    "jquery": "^3.1.1",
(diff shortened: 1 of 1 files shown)


henrywang:
It works for me.

henrywang:
@jgiardino, do you have any concern about lower version of jQuery?

jgiardino:
No concerns from me. Thanks!

PR #155, opened 2017-10-03
Avoid using non-standardized ECMAScript features

martinpitt:
We don't want to condone arbitrary JavaScript syntax proposals which
haven't been standardized yet. So drop the stage-1 babel preset and
replace it with the explicit two features that the code currently uses:
object spreads and class properties.

### DIFF
--- a/components/Layout/Layout.js
+++ b/components/Layout/Layout.js
@@ -8,16 +8,15 @@ import utils from '../../core/utils';
 class Layout extends React.Component {
   constructor() {
     super();
+    this.state = { notifications: [] };
     this.setNotifications = this.setNotifications.bind(this);
   }
-
-  state = { notifications: [] }
 
   componentWillMount() {
     this.setNotifications();
   }
 
-  setNotifications = () => {
+  setNotifications() {
     this.setState({ notifications: NotificationsApi.getNotifications() });
   }
 
--- a/components/Link/Link.js
+++ b/components/Link/Link.js
@@ -33,8 +33,9 @@ class Link extends React.Component {
   };
 
   render() {
-    const { to, ...props } = this.props; // eslint-disable-line no-use-before-define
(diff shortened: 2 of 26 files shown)


jgiardino:
No objections here. @jkozol will you take a look at this too?

martinpitt:
object spreads are currently only [at stage 3](http://2ality.com/2016/10/rest-spread-properties.html), thus ideally it wouldn't be used yet in production code. I might look into porting the code away from it, to `Object.assign()`.

Class properties are also [only at stage 3](https://github.com/tc39/proposal-class-fields). They are reasonably easy to avoid, but this needs some care as e. g. replacing a `foo = () => {}` method definition with `foo() {}` will require an additional `.bind()` call to make sure `this` is correct.

martinpitt:
I pushed a followup commit which ports the code away from the object spread operator. The first commit [passed tests](https://travis-ci.org/weldr/welder-web/builds/282747043?utm_source=github_status&utm_medium=notification) , so if you prefer I can propose the porting as a separate PR.

jkozol:
This looks good to me. Is there any issue with us using generator functions?

martinpitt:
@jkozol: No, [generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) are  part of es2015, so they are fine.

martinpitt:
I now ported the other non-standard extension (class properties) to ES2015. IMHO it does not really improve readability and was not consistently used anyway, so this makes the code more consistent.

martinpitt:
Any reservations about the third commit? Or can this be merged? Thanks in advance!

jgiardino:
No objections from me. @jkozol will you give this one last look and merge if you're ok with it?

jkozol:
This looks good to me too. Thanks for making the code more consistent.

PR #156, opened 2017-10-04
Explicitly eslint-declare jQuery where needed

martinpitt:
Don't ignore undefined "$" references (jQuery) globally in eslint, as
that cannot be done when building from Cockpit. Declare it per-file
instead.

----

In Cockpit we avoid using jQuery in new code, and it is really discouraged to use *both* React and jQuery in the same module. Thus we don't want a blanket exception for it in eslint, and this makes it easier to track the jQuery removal from welder-web code.

### DIFF
--- a/components/CardView/CardView.js
+++ b/components/CardView/CardView.js
@@ -1,3 +1,5 @@
+/* global $ */
+
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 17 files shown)


martinpitt:
Fails tests for some reason, marking "work in progress".

martinpitt:
I reworked this to only be declarative for eslint, it doesn't change runtime behaviour. This isn't strictly necessary for welder-web itself, but will help the cockpit integration.

PR #159, opened 2017-10-06
Move/remove actions on Recipe page

jgiardino:
The actions Edit Recipe, Create Composition, and Export should not display
below the tabs as tab-level actions, but instead display in the page header as
page-level actions. The Search icon that displays in the toolbar for some of
these tabs is non-functional and should be removed. Finally, if these changes
result in any of the tabs having an empty toolbar, then the toolbar should be
removed.


henrywang:
The failed case is due to issue 145. Could you please take a look at issue 145 if you have time? Thanks. @jgiardino

PR #160, opened 2017-10-06
Update Recipes page empty state and non-functional actions

jgiardino:
Update Empty State message and include Create Recipe action
to be more consistent with the patternfly pattern. Remove actions
from the toolbar kebab that are not functional and have no plans
to be implemented in the near future.

### DIFF
--- a/pages/recipes/index.js
+++ b/pages/recipes/index.js
@@ -117,51 +117,11 @@ class RecipesPage extends React.Component {
                 </button>
               }
               </div>
-
               <div className="toolbar-pf-action-right">
                 <div className="form-group">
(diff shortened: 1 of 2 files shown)


henrywang:
All the things work for me.
One question about this feature. There're two Create Recipe button in the page. It's a little bit confused for user. Can we remove one of them? Thanks.
screenshot-2017-10-9 recipes

jgiardino:
Thanks for reviewing, @henrywang!

The layout of the empty state is based on the [patternfly design pattern](http://www.patternfly.org/pattern-library/communication/empty-state/#/overview), where the main action on the page, usually the action that would create items in the list, is duplicated in the empty state message.

PR #162, opened 2017-10-15
Save Recipe Workflow

jkozol:
This PR meets the UI Acceptance Criteria of:
    
- When a change is made to a recipe, then the number of pending changes displayed next to the Save button is incremented.
- When no pending changes exist, then actions for Save, Discard, View and Comment are not available (buttons are disabled, pending changes text and link are not visible).
- Given one or more pending changes exist, when the user clicks the link "View and Comment", the pending changes are listed in the modal.
- When the user hits Save, then pending changes are committed to the recipe
- The modal dialog includes a comment field
- Undo and Redo actions

In regards to the failed test, I am uncertain why it is failing on travis. It passes locally.

### DIFF
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -10,6 +10,7 @@ class PendingChanges extends React.Component {
       comment: ""
     };
     this.handleChange = this.handleChange.bind(this);
+    this.handleSaveChanges = this.handleSaveChanges.bind(this);
   }
 
   componentWillMount() {
@@ -85,7 +86,7 @@ class PendingChanges extends React.Component {
                 </div>
                 <strong>Pending Changes</strong><span className="text-muted"> (most recent first)</span>
                 <ul className="list-group">
-                  {this.props.componentUpdates.map((componentUpdated, index) => (
+                  {this.props.recipe.pendingChanges.map((componentUpdated, index) => (
                     <li className="list-group-item" key={index}>
                       {componentUpdated.componentNew && componentUpdated.componentOld &&
                         <div className="row">
@@ -129,16 +130,15 @@ PendingChanges.propTypes = {
   recipe: PropTypes.object,
   contents: PropTypes.array,
   handleHideModal: PropTypes.func,
-  componentUpdates: PropTypes.array,
   setRecipeComment: PropTypes.func,
   handleSave: PropTypes.func,
+  modals: PropTypes.object,
 };
-
 const mapStateToProps = state => ({
-  componentUpdates: state.modals.pendingChanges.componentUpdates.present,
+  modals: state.modals,
 });
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch) => ({
   setRecipeComment: (recipe, comment) => {
     dispatch(setRecipeComment(recipe, comment));
   },
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -23,6 +23,7 @@ class PendingChanges extends React.Component {
 
   handleSaveChanges() {
     $('#cmpsr-modal-pending-changes').modal('hide');
+    this.props.handleSave();
   }
 
   handleChange(e) {
@@ -130,6 +131,7 @@ PendingChanges.propTypes = {
   handleHideModal: PropTypes.func,
   componentUpdates: PropTypes.array,
   setRecipeComment: PropTypes.func,
+  handleSave: PropTypes.func,
 };
 
 const mapStateToProps = state => ({
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -29,7 +29,6 @@ class PendingChanges extends React.Component {
     this.setState({comment: e.target.value});
   }
 
-
   render() {
     return (
       <div
@@ -134,7 +133,7 @@ PendingChanges.propTypes = {
 };
 
 const mapStateToProps = state => ({
-  componentUpdates: state.modals.pendingChanges.componentUpdates,
+  componentUpdates: state.modals.pendingChanges.componentUpdates.present,
 });
 
 const mapDispatchToProps = dispatch => ({
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -1,17 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 
 class PendingChanges extends React.Component {
   constructor() {
     super();
-    // dummy data
-    this.state = {
-      componentUpdates: [
-        { componentOld: null, componentNew: 'rsync-3.0-17.317' },
-        { componentOld: 'httpd-2.3-45.el7', componentNew: 'httpd-2.4-45.el7' },
-        { componentOld: 'basesystem-10.0-7.el7', componentNew: null },
-      ],
-    };
   }
 
   componentDidMount() {
@@ -76,8 +69,8 @@ class PendingChanges extends React.Component {
                 </div>
                 <strong>Pending Changes</strong><span className="text-muted"> (most recent first)</span>
                 <ul className="list-group">
-                  {this.state.componentUpdates.map((componentUpdated) => (
-                    <li className="list-group-item">
+                  {this.props.componentUpdates.map((componentUpdated) => (
+                    <li className="list-group-item" key={componentUpdated.componentNew}>
                       {componentUpdated.componentNew && componentUpdated.componentOld &&
                         <div className="row">
                           <div className="col-sm-3">Updated</div>
@@ -120,6 +113,12 @@ PendingChanges.propTypes = {
   recipe: PropTypes.string,
   contents: PropTypes.array,
   handleHideModal: PropTypes.func,
+  componentUpdates: PropTypes.array,
 };
 
-export default PendingChanges;
+
+const mapStateToProps = state => ({
+  componentUpdates: state.modals.pendingChanges.componentUpdates,
+});
+
+export default connect(mapStateToProps)(PendingChanges);
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -1,10 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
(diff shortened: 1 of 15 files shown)


jgiardino:
Thanks for working on this! I did a quick review of the UI and the state. There are a few user task flows that don’t behave the way I expect them to. They are outlined below. Please let me know if you have any questions about these.

Also, when I look at the state for `recipes` it doesn’t match what I would expect. I would assume that `recipes` is an array of recipes, and that within a recipe, you would have past, present and future per recipe, which captures only changes made to that recipe. But currently, `recipes` uses the structure shown below, which seems like it wouldn't work for managing changes per recipe, and seems like it would grow unnecessarily based on the number of recipes and the number of changes pending save just based on how the data is stored under `past`.

```
recipes: {
	past: [
		[array of recipes],
		[array of recipes],
		* this array is listed per change I made in one of the recipes
	],
	present: [
		[array of recipes]
	],
	future: [
		[array of recipes]
	],
},
```


## User task flows:

### The dialog

#### Task flow 1 - Pending Changes sort order
1. Make changes to the recipe
2. Click the link labeled “View and Comment” to show the Changes Pending Save dialog
**Expected result:** the list of Pending Changes are sorted by most recent first
**Actual result:**  the list of Pending Changes are sorted by oldest first

#### Task flow 2 - Persistent comment
1. Make changes to the recipe
2. Click the link labeled “View and Comment” to show the Changes Pending Save dialog
3. enter a comment
4. click Close
5. Click the link labeled “View and Comment”
**Expected result:** the comment that was entered remains in the input field
**Actual result:** the comment is cleared from the input field
Additional notes: At a minimum, the comment should be persistent as long as the user is on the Edit Recipe page. Beyond that, either result is fine (i.e. clearing or persisting the comment) as long as the comment only displays for the recipe for which it was entered.

#### Task flow 3 - Add then Remove
1. Add a package, e.g. 389-ds-base
2. Click “View and Comment” to see the package listed under Pending Changes
3. Remove the package
4. Click “View and Comment”
**Expected result:** The package that I added in step 1 should not be listed
**Actual result:** The package that I added in step 1 is listed
**Additional notes:** The list of Pending Changes should reflect the difference in selected components between the current state and the previous commit. If I add a package then remove that package, then there is no difference and this should not be listed.

#### Task flow 4 - Undo
1. Add a package, e.g. 389-ds-base
2. Click “View and Comment” to see the package listed under Pending Changes
3. Click Undo icon
4. Click “View and Comment”
**Expected result:** The package that I added in step 1 should not be listed
**Actual result:** The package that I added in step 1 is listed
**Additional notes:** The list of Pending Changes should reflect the difference in selected components between the current state and the previous commit. If I make changes then undo those changes, then the change should be cleared from the list of Pending Changes.

### Navigation
#### Task flow 5 - navigate away then come back
1. Make changes to the recipe
2. Navigate to another page
3. Come back to Edit Recipe page for the modified recipe
**Expected result:** The changes pending save are visible in the recipe, and are noted in the dialog and counted in the text that precedes “View and Comment”
**Actual result:** The count remains, along with showing the changes in the dialog. But the contents of the recipe do not show the changes pending save.

#### Task flow 6 - navigate to another recipe
1. Make changes to the recipe
2. Navigate to the Edit Recipe page of another recipe
**Expected result:** There should be no indication of changes pending save in the top-right corner of the page (e.g. no text with number of changes, no link to View and Comment, and the buttons Save and Discard Changes should be disabled)
**Actual result:** The changes that were made for the first recipe are listed as pending changes for the second recipe.

henrywang:
The available components, which is not added into this recipe, should be available to be added.
How to reproduce:
1. Add one of available component into a recipe.
2. Save this recipe.
3. Navigate another recipe and edit.
4. The added component in the first step is not available to be use for this recipe.
selected component issue

henrywang:
When save action complete, the "x change" and "View and Comment" should be dismissed. The Save and Discard Changes buttons should be not available.
For now, the page has to be refreshed to dismiss "x change" and "View and Comment" and disable Save and Discard Changes buttons.
after saved issue

jkozol:
Thank you both for the feedback. I have a pushed a commit that addresses Task Flow 1 and 2. I will push additional commits to address the other issues and can clean up the commit log at the end if you like.

jgiardino:
I did see the sort order in the Pending Changes dialog is fixed, and that the comment is persistent. Another issue I noticed is that if the first action is Add, then I don't see the Remove actions listed as pending changes, and vice versa. Also, if I hit refresh, on the page, the changes are no longer present.

jgiardino:
Thanks for the updates! I now see that I can add and remove in any order, and the list of changes is updated as expected.

Per a separate discussion with @bcl about using this api call `/api/v0/recipes/diff/http-server/NEWEST/WORKSPACE` to get the difference between the workspace and the latest commit (i.e. when the user returns to the Edit page after not saving changes to see there are changes pending save), he had mentioned that these changes won't be returned in any specific order. I'm interested in your thoughts on whether we continue to sort them during the current editing session or not. I think it would be fine to display the list sorted for the current session and unsorted for changes made in a previous session. And if we do handle these lists differently, we can alter how these are listed in the dialog so that it's clear what edits were made during the current session and a previous one (which might actually be really good to know, now that I'm thinking about it). Let me know what you think and I can capture these details in a separate issue.

Some other things I noticed... (I will add these to my original comment with the list of task flows, to make it easier to track these):
- Save button on dialog currently doesn't appear to be hooked up to the Save action
- Task flows 3 and 4 appear to work now, but in the case where one change is made, and that change is undone either manually or via Undo, the Save and Discard Changes buttons remain enabled and "0 changes" displays. Instead, the buttons should be disabled, and the text and link should be removed from the UI

jkozol:
Will the pending changes for a previous session be stored locally or fetched from through the API?

jgiardino:
For reference, [here are some notes](https://docs.google.com/presentation/d/1KaZD4nmPsPhGF2pafoeDMmqFfDnOIzSpKNmZ9O-g5H4/edit#slide=id.g1e3035a780_0_50) about pending changes that I made when we chatted about this earlier this year.

To answer your question, this is what I'm thinking...
Given a user has changes pending save, when a user navigates to the Edit Recipe page, the changes should be pulled both from the local storage and from the workspace. Because depending on the circumstances of what browser they're using and what happened while they were making changes, it's possible that the there are changes in the local storage that are not reflected in the workspace (i.e. the browser crashed before we saved the changes to the workspace), or it's possible that the changes are in the workspace but not the local storage (i.e. they switched to a different browser -- although I'm assuming that browsers don't have access to the same local storage). 

Although, I'm wondering again, is there a reason why we don't want to save to the workspace immediately? I know we talked about this, and I think the reason is we don't want to frequently post changes, but I also wonder whether that many changes are really going to be happening.

Just throwing that out there. ;-) I'm definitely interested in what you think we should do. I'm also fine with deferring the local storage piece to a separate PR if that's something that you haven't started yet. I do think that if we're separating the work, that the workspace is a higher priority than the local storage. And that's also something that can be deferred to a later PR as well, along with task flows 5, 6, and 7 that I mentioned above.

jgiardino:
@jkozol, I'm not able to review updates for this PR due to errors. In my local build, I don't have any recipes, so I get this error when I try to create a new recipe. I'm not sure if this is an issue with my local build, or if the create recipe flow has issues in your latest changes.

These are some of the errors I see listed in the console: 

- fetchRecipeContents has been cancelled 
- Uncaught TypeError: Cannot read property 'id' of undefined

jkozol:
@jgiardino Can you try clearing your cache and offline website data? Since I changed the recipe state any recipes that you have persisted in your local storage will incorrectly rehydrate into the store. Also, all issues should be resolved except for task flow 7.

jgiardino:
Hi @jkozol, yes, thank you for suggesting that I clear my cache. That was the issue.

I do see that all task flows work as expected except for #7. 

An additional issue that I see is with the Recipe sort order. I'm not sure if this is related or not, but in PR #140 you had fixed the sort order on the Recipes page, however with this PR, I'm not seeing that update. For example, when I return to the Recipes page, I see the most recently modified recipe display last. I went ahead and created an issue for this: #174

jgiardino:
Oh, forgot to mention, I am fine with merging this, and addressing the issue with refresh in a separate PR.

@henrywang could you test this one more time?

Thanks!

henrywang:
@jkozol, there's an issue in create recipe page. I cannot input any letter in Name and Description input box. The letter will be deleted when I press any letter. Could you please take a look at this issue? Thanks.

jkozol:
@henrywang I am not able to replicate this issue. Are you getting any errors in the console?

henrywang:
@jkozol, the console reports error TypeError
```shell
Uncaught TypeError: Cannot read property 'name' of undefined
```
This line of code of function getSortedRecipes caused this error:
```javascript
if (a.present[key] > b.present[key]) return value === 'DESC' ? 1 : -1;
```

jkozol:
@henrywang Can you try clearing your cache and offline website data? Since I changed the recipe state, any recipes that you have persisted in your local storage will incorrectly rehydrate into the store without past, present, or future sub-objects.

henrywang:
@jkozol, I cleared my cache, the same issue was reproduced. The main page (recipes page) did not show the recipe. No any recipe in main page. I roll back to master branch, everything works well.
Firefox console reported the other errors:
```shell
TypeError: a.present is undefined
```
and
```shell
uncaught at _callee recipes/<@http://localhost:3000/dist/main.js
uncaught at _callee getSortedRecipes/<@http://localhost:3000/dist/main.js
```

jkozol:
@henrywang I believe your issue is caused by the way we are persisting the recipes. I am going to remove this persistence since we will be using the workspace to save changes in between sessions anyways. While testing this, however, I began to get errors when adding components to recipes. When I click to add a component I get the console error:

```
TypeError: NetworkError when attempting to fetch resource.
```

and the API error:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidColumnType(4, Text)', /checkout/src/libcore/result.rs:906:4
```

I am getting this error even on master. Do you have any idea what might be causing this?

henrywang:
@jgiardino, feel free to merge this PR. Thanks.

jgiardino:
Looks good on my end. I don't get the errors that you mentioned, @jkozol.

Thanks, @jkozol and @henrywang!

PR #165, opened 2017-10-20
Update wait element to make case more stable.

henrywang:
Add header class before ol class to make finding element with element hierarchy.


jgiardino:
Looks good to me!

henrywang:
Update about failed case:
1. api image is able to be built, but clippy has to be disabled.
2. UI is able to show all of components and Imported Content Sanity Testing passed.
3. the failed case "copy and paste correct components" is because depsolve does not work. There's empty in dependencies tab from UI.

atodorov:
Merging since Jenn has already approved.

PR #170, opened 2017-11-03
Replace eval(coverage.js) in tests with proper import

martinpitt:
Slurping in utils/coverage.js via a manual file read and `eval()` is a
big and unnecessary hack. Turn it into a proper function instead.


martinpitt:
I added another commit "Drop unsafe Content-Security-Policy from Cockpit manifest" as I thought that would work with dropping the `eval()`, but in CI it still fails. Locally the cockpit integration test works fine, so this will be rather ugly to debug. I reverted that second commit for now, but even without that this change still makes sense IMHO.

PR #172, opened 2017-11-07
Remove metadata.db depencence for test test_importSanity.

henrywang:
No database query any more, use API call instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,13 +42,11 @@ stylelint:
 unit-test:
 	npm run test:cov
 
-metadata.db:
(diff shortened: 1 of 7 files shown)


atodorov:
@henrywang can you update both README's which also mention metadata.db

atodorov:
One more change. It looks like metadata.db is only used by the API container so it's no longer shared. Can you update Makefile ?

henrywang:
shared section is used by both end-to-end and cockpit test. metadata.db section is used by shared which run API image, so it needs metadata.db.

PR #173, opened 2017-11-09
Install Welder Web into Cockpit by RPM.

henrywang:


Dockerfile.cockpit
```diff
@@ -5,5 +5,9 @@ RUN dnf install -y cockpit-ws cockpit-kubernetes && dnf clean all
 RUN echo $'[Negotiate]\nCommand = /usr/libexec/cockpit-stub\n[WebService]\nShell = /shell/simple.html' > /etc/cockpit/cockpit.conf
 RUN rm -rf /usr/share/cockpit/kubernetes/
 
+COPY welder-web*.rpm /tmp/
+RUN rpm -ih /tmp/welder-web*.rpm
```
atodorov: dnf install -y so we can pick up dependencies if later we have such

test/end-to-end/entrypoint.sh
```diff
@@ -8,7 +8,7 @@ sed -i "s,\"uri\": \"http://localhost:4000/\",\"uri\": \"http://api:4000/\"," co
 if [ -z "$COCKPIT_TEST" ]; then
     sed -i "s,\"root\": \"http://localhost:3000/\",\"root\": \"http://web:3000/\"," config.json
 else
-    sed -i "s,\"root\": \"http://localhost:3000/\",\"root\": \"http://web:9090/welder\"," config.json
+    sed -i "s,\"root\": \"http://localhost:3000/\",\"root\": \"http://web:9090/welder-web\"," config.json
```
atodorov: Why is this change necessary ?

test/end-to-end/pages/editRecipe.js
```diff
@@ -32,7 +32,7 @@ module.exports = class EditRecipePage extends MainPage {
     // Recipe Name link
     this.varLinkToViewRec = `${this.mailUrl}#/recipe/${this.recipeName}`;
     if (pageConfig.root.includes('9090')) {
-      this.varLinkToViewRec = `${this.mailUrl.slice(0, -6)}cockpit/@localhost/welder/index.html#/recipe/${this.recipeName}`;
+      this.varLinkToViewRec = `${this.mailUrl.slice(0, -10)}cockpit/@localhost/welder-web/index.html#/recipe/${this.recipeName}`;
```
atodorov: Same question as above, why is this necessary?

test/end-to-end/entrypoint.sh
henrywang: The RPM package installed into /usr/share/cockpit/welder-web directory, so the URL should be changed.

Dockerfile.cockpit
henrywang: In this case, there has to be with COPY instruction. Because for the ADD and COPY instructions, the contents of the file(s) in the image are examined and a checksum is calculated for each file. For the RUN instruction, cache checking will not look at the files in the container to determine a cache match.

test/end-to-end/pages/editRecipe.js
henrywang: The "welder-web" has ten characters, but welder has six.

Dockerfile.cockpit
atodorov: I'm not arguing about the COPY or RUN instructions. Instead of
`rpm -ih /tmp/welder-web*.rpm` (btw a better way is to rpm -Uh) use `dnf -y install /tmp/welder-web*.rpm` which will pull additional RPM requirements if/when the welder-web RPM package introduces them.

test/end-to-end/entrypoint.sh
atodorov: ok then

Dockerfile.cockpit
henrywang: Got it.

test/end-to-end/entrypoint.sh
martinpitt: Still,  this URL hack is rather ugly. You could instead add a `"name": "welder"` into the cockpit manifest, or better yet, harmonize the install path and the name to all be `welder`?

henrywang: @jgiardino and @jkozol, is that possible to change RPM install path to welder? Thanks.

atodorov:
@henrywang this PR has been open for quite a long time. Please resolve the review comments so we can merge it.

atodorov:
Closed via #180

PR #176, opened 2017-11-20
tests: Fix metadata.db download regression

martinpitt:
Commit 147cf6e62 moved the metadata.db from a Makefile dependency to an
unconditional command in the `shared` target. As the latter is phony,
this leads to unnecessarily downloading the metadata for every test run.
Revert that bit.

CC: @henrywang, can you please review/land?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,11 +42,13 @@ stylelint:
 unit-test:
 	npm run test:cov
 
-shared:
+metadata.db:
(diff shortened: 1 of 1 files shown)


henrywang:
@martinpitt, do we need to remove db download in the `shared` target.

martinpitt:
@henrywang , err, of course we do. git pilot error (forgot to --amend). Pushed, thanks for spotting!

PR #178, opened 2017-12-07
Add make targets for Jenkins

atodorov:


atodorov:
@henrywang @martinpitt can you take a look at the failures from `jest --coverage`. Surprisingly they allso appear in the Travis logs but they don't trigger a failure. (note: the failure in Jenkins is triggered later by the --cache-from parameter for docker).

PR #179, opened 2017-12-07
Drop obsolete eslint override

martinpitt:
The function `updateInputComponentData()` was fixed recently [1] to be
synchronous instead of async, so it is correct now and the eslint
error is gone. Drop the annotation to avoid confusion and regression.

[1] https://github.com/weldr/welder-web/commit/50630524afbc1c03eb06204a44240edceea45498#diff-9e47aee6cedcc9f80cf6432dc6e63aff

----

This is cleanup from #151.


atodorov:

atodorov:
@martinpitt FYI martin I've added you to the PR build whitelist for Jenkins so that when you push another PR it will be build automatically.

martinpitt:
Thanks @atodorov

PR #180, opened 2017-12-12
Install Welder Web into Cockpit by RPM with improvements

atodorov:
This PR takes #173, rebases it to the latest changes and harmonizes the RPM installation path.

Opening as a new PR since we're not getting any movement on the original one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ before_install:
 install:
   - |
         if [ "$COMMAND" == "cockpit-test" ]; then
-            node run build --with-coverage
             sed -i "s,var welderApiHost=.*,var welderApiHost=\"api\";," ./public/js/config.js
         fi
 env:
@@ -44,7 +43,7 @@ script:
 after_success:
   - if [ "$COMMAND" == "unit-test" ]; then cat ./coverage/lcov.info | codecov; fi
   - |
-        if [ "$COMMAND" == "end-to-end-test" ] || [ "$COMMAND" == "cockpit-test" ]; then
+        if [ "$COMMAND" == "end-to-end-test" ]; then
             nyc report
             nyc report --reporter=lcov && codecov
         fi
--- a/Dockerfile.cockpit
+++ b/Dockerfile.cockpit
@@ -5,5 +5,8 @@ RUN dnf install -y cockpit-ws cockpit-kubernetes && dnf clean all
 RUN echo $'[Negotiate]\nCommand = /usr/libexec/cockpit-stub\n[WebService]\nShell = /shell/simple.html' > /etc/cockpit/cockpit.conf
 RUN rm -rf /usr/share/cockpit/kubernetes/
 
+COPY welder-web*.rpm /tmp/
+RUN dnf -y install /tmp/welder-web*.rpm && rm -f /tmp/welder-web*.rpm
+
 CMD ["/usr/libexec/cockpit-ws", "--no-tls"]
 EXPOSE 9090
--- a/Makefile
+++ b/Makefile
@@ -77,16 +77,29 @@ end-to-end-test: shared
 	    xvfb-run -a -s '-screen 0 1024x768x24' npm run test -- --verbose
 	sudo docker ps --quiet --all --filter 'ancestor=welder/web-with-coverage' | sudo xargs --no-run-if-empty docker rm -f
 
-cockpit-test: shared
+build-rpm:
+	if [ -n "$$TRAVIS" ]; then \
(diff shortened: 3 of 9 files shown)


atodorov:
@martinpitt can you review please? This has been open for too long and I'd like to be done with it.

Dockerfile.cockpit
```diff
@@ -5,5 +5,9 @@ RUN dnf install -y cockpit-ws cockpit-kubernetes && dnf clean all
 RUN echo $'[Negotiate]\nCommand = /usr/libexec/cockpit-stub\n[WebService]\nShell = /shell/simple.html' > /etc/cockpit/cockpit.conf
 RUN rm -rf /usr/share/cockpit/kubernetes/
 
+COPY welder-web*.rpm /tmp/
+RUN dnf -y install /tmp/welder-web*.rpm
+RUN rm -f /tmp/welder-web*.rpm
```
martinpitt: I suggest to put that into one `RUN` command (`dnf && rm`), otherwise the `rm` is pointless. Remember that each RUN creates a new docker layer, so effectively you can't remove files in separate steps if you actually want to avoid wasting space.

.travis.yml
```diff
@@ -44,7 +43,7 @@ script:
 after_success:
   - if [ "$COMMAND" == "unit-test" ]; then cat ./coverage/lcov.info | codecov; fi
   - |
-        if [ "$COMMAND" == "end-to-end-test" ] || [ "$COMMAND" == "cockpit-test" ]; then
```
martinpitt: This affects code coverage report, right? I. e. this seems unrelated to how welder gets installed into the container. What's the rationale for this? (NB: This may well be correct, I'm not very familiar with the docker-based tests here)

Makefile
```diff
@@ -77,16 +77,29 @@ end-to-end-test: shared
 	    xvfb-run -a -s '-screen 0 1024x768x24' npm run test -- --verbose
 	sudo docker ps --quiet --all --filter 'ancestor=welder/web-with-coverage' | sudo xargs --no-run-if-empty docker rm -f
 
-cockpit-test: shared
+build-rpm:
+	if [ -n "$$TRAVIS" ]; then \
+	    sudo docker build -f Dockerfile.buildrpm --cache-from welder/buildrpm:latest -t welder/buildrpm:latest .; \
```
martinpitt: Does it hurt to have the `--cache-from` on a local build? (I don't see why it would)

Makefile
```diff
@@ -77,16 +77,29 @@ end-to-end-test: shared
...
 	else \
 	    sudo docker build -f Dockerfile.cockpit -t welder/web-cockpit:latest .; \
 	fi;
 
-	sudo docker run -d --name web --restart=always -p 9090:9090 -v `pwd`/public:/usr/share/cockpit/welder --network welder welder/web-cockpit:latest
+	sudo docker run -d --name web --restart=always -p 9090:9090 --network welder welder/web-cockpit:latest
+
+# Clean generated intermidiate tar file and useless RPM file
+# RPM file is inside docker image already
+	sudo rm -f welder-web*.rpm welder-web.tar.gz
```
martinpitt: Let's please not remove local files as root :-)

.travis.yml
atodorov: The RPM content is not instrumented for code coverage so we can't gather coverage data once this change is applied.

Makefile
atodorov: It does not but the version of Docker on the RHEL builders is older and doesn't support that flag. OTOH we're migrating from Travis (hopefully) and at some point should have docker caching enabled locally so this flag will not be necessary.

.travis.yml
martinpitt: Makes sense, thanks.

Makefile
martinpitt: Ah, that's where it comes from, thanks.

Dockerfile.cockpit
```diff
@@ -5,5 +5,8 @@ RUN dnf install -y cockpit-ws cockpit-kubernetes && dnf clean all
 RUN echo $'[Negotiate]\nCommand = /usr/libexec/cockpit-stub\n[WebService]\nShell = /shell/simple.html' > /etc/cockpit/cockpit.conf
 RUN rm -rf /usr/share/cockpit/kubernetes/
 
+COPY welder-web*.rpm /tmp/
+RUN dnf -y install /tmp/welder-web*.rpm && rm -f /tmp/welder-web*.rpm
```
martinpitt: I just realized that this won't actually save any space as it's already the COPY layer that has the file. But nevermind, it still saves an extra layer (even though not  the actual space for the .rpm, but that's small enough).

PR #181, opened 2017-12-13
Bind-mount .nyc_output/ to /tmp/.nyc_output inside docker

atodorov:
this leaves a functional /tmp in the container and solves issues
with mktemp (called from xvfb-run), possibly triggered by SELinux


atodorov:
OK, got EACCESS in Jenkins, creating a coverage file inside the bind-mount directory. I'm going to merge and probably disable SELinux altogether on the host executing this job

PR #185, opened 2018-01-08
Import minimal metadata.db instead of downloading it

atodorov:
we want to get rid of the humongous CentOS 7 metadata.db and use
a small file for testing. Here metadata.db isn't really used for
anything and is only a requirement so we can run the api
container.


atodorov:
The failures are when instlling rpm-build before running Cockpit tests. Reported against Fedora here: https://bugzilla.redhat.com/show_bug.cgi?id=1532293

Issue #49, opened 2017-03-09
Save recipe needs to save both version and release


jgiardino:
### Description of problem
Currently, the information that is provided on save is name and version, but release is not included. This means that the selected release version will not be saved with the recipe.

### Component (web, API, etc)
web

### Steps to Reproduce

1. Navigate to the Edit Recipe page
2. Find a component that has more than one version (click the component name, click the Version Release select menu to see number of available versions)
3. Click the component name to view details.
4. Click the Version Release select menu to select a different version and release value. (The default one is the one that displays in the popover when clicking the list item on the left, and is also most likely selected by default)
5. Save the recipe.
6. Reload the page.

### Actual results
The default values for version and release will display.

### Expected results
The selected values for version and release should display.

jkozol:
This issue is outdated.

PR #186, opened 2018-01-12
Load Pending Changes from Workspace

jkozol:
A user should now be able to view a list of pending changes with those from the current session listed in the order they were made (newest first) and the unsaved updates from prior sessions shown separately. The user can discard these pending changes and can undo/redo both local and workspace changes. This PR does not include the discard changes modal, the pending changes modal info tooltip, nor an alert when undoing workspace changes. Also, when loading components from the workspace the only dependencies loaded are the components themselves.

### DIFF
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -133,7 +133,7 @@ class PendingChanges extends React.Component {
                     } {this.props.recipe.workspacePendingChanges.deletedChanges.length !== 0 &&
                       <li className="list-group-item">
                         <div className="row">
-                          <div className="col-sm-3">Deleted</div>
+                          <div className="col-sm-3">Removed</div>
                           <div className="col-sm-9">
                             <ul className="list-unstyled">
                               {this.props.recipe.workspacePendingChanges.deletedChanges.map((componentUpdated, index) => (
--- a/components/Modal/PendingChanges.js
+++ b/components/Modal/PendingChanges.js
@@ -88,7 +88,7 @@ class PendingChanges extends React.Component {
                 </div>
                 <strong>Pending Changes</strong><span className="text-muted"> (most recent first)</span>
                 <ul className="list-group">
-                  {this.props.recipe.pendingChanges.map((componentUpdated, index) => (
+                  {this.props.recipe.localPendingChanges.map((componentUpdated, index) => (
                     <li className="list-group-item" key={index}>
                       {componentUpdated.componentNew && componentUpdated.componentOld &&
                         <div className="row">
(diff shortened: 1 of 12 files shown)


atodorov:

jgiardino:
Thanks for making these updates! 

When I reviewed this, I noticed a couple of issues:

1. The only updates that are pulled from the workspace are updates where I remove components. When I add components, they are never listed.
2. On the Edit Recipe page, we "add" or "remove" components to/from the recipe. But we don't "create" or "delete" components. On the Pending Changes dialog, where you list changes in the workspace, can you change the string "Deleted" to "Removed"?

I tested using the following steps, and everything else works as expected (except where you note above):

1. Opened one of my recipes for edit
2. Removed one component and added 2 components
3. Closed the tab
    - Are there other situations that result in pulling from the workspace instead of the store? 
4. Opened a new tab and pasted in the url for the Edit Recipe page of the recipe I updated
5. Clicked Undo
6. Clicked Redo
7. Clicked Save

jkozol:
Thank you for your feedback. I believe I fixed the issue of added components not being listed. I also changed the modal to say "Removed". 

> Are there other situations that result in pulling from the workspace instead of the store?

The only situation where we will pull from the workspace is when we fetch our recipe contents and the workspace version is the most up to date version. We fetch the recipe contents on page load for the edit recipe page and the recipe page.

jgiardino:
Thanks for making these updates!

The only other issue I notice is that there's a difference regarding when version displays. There has also been an issue related to displaying release, but this is a known issue (#49). Do you know why this is happening? 

When the Selected Components are loaded, they display the version:
image 
When they're pulled from the workspace, the version is empty:
image
But the Pending Changes dialog includes this information:
image

Also, the reason that we the only dependencies listed are the selected components themselves when pulling from the workspace is because the api can't depsolve the workspace version of a recipe. Is that correct? Is the expectation that we would be able to do this in the future?

jkozol:
The version and release issue should be fixed now. I was also able to get the workspace dependencies working. The dependencies are still a little weird when adding and removing components but I feel that fixing this is best done when I rework the data structure into a single list of components.

jgiardino:
I don't see the tests listed in this PR like they are in other PRs.
@henrywang @atodorov Do you know why the tests aren't running or listed here?

atodorov:
@jgiardino do you mean all the "Check jobs" that are shown automatically in GitHub's interface? If this is what you mean the ones coming from Jenkins were probably never kicked off for some reason. You can always force push to this PR to cause all the jobs to restart.

OTOH Jenkins is duplicating what Travis does so you can merge as-is.

jgiardino:
LGTM
Thanks!

PR #187, opened 2018-01-18
Prepare for automatic releases

martinpitt:
As discussed on the mailing list, I am working on automating welder-web releases with [cockpituous](https://github.com/cockpit-project/cockpituous). This fixes the tarball and spec file to actually get an idea of releases and versions, and provides an initial cockpituous release script.

I tested this on my own welder-web fork on https://github.com/martinpitt/welder-web, and ran:

in my welder-web branch:

    git tag -s 0.0.1 -m "$(printf '0.0.1\n\n- test release\n')"
    git push --tags martin

in an empty temporary directory:

    ~/upstream/cockpituous/release/release-runner -r https://github.com/martinpitt/welder-web utils/cockpituous-release

The latter command would be done by a bot from a cockpituous container eventually, but this makes it easy to test locally.

This produced a first release: https://github.com/martinpitt/welder-web/releases, with correct release notes (from the tag) attached) and the produced tarball. It also produced a correct srpm with adjusted %changelog and Version:. This does not currently  get uploaded anywhere, but that (COPR) will be next step after this.

As I will keep removing and re-adding the tag/release for testing, here is a screenshot:

release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+# generate version number from the latest git tag; if the topmost commit is
+# tagged (at release time), just use that; if there are commits after that,
+# append a ".x" suffix to indicate that's a development snapshot
+# if there is no tag (yet), use 0 as version
+RELEASEVER=$(shell (git describe --exclude '*jenkins*' || echo 0) | sed 's/-[0-9]\+-g.*/.x/')
+
 all:
 	NODE_ENV=$(NODE_ENV) npm run build
 
@@ -15,13 +21,14 @@ dist-gzip: all
 	# remove evil file name that breaks rpmbuild (https://github.com/patternfly/patternfly/issues/917)
 	find _install -name 'Logo_Horizontal_Reversed.svg alias' -delete
 	cp welder-web.spec _install/
-	tar -C _install/ -czf welder-web.tar.gz .
+	tar -C _install/ -czf welder-web-$(RELEASEVER).tar.gz .
 	rm -rf _install
 
 srpm: dist-gzip
 	rpmbuild -bs \
 	  --define "_sourcedir `pwd`" \
 	  --define "_srcrpmdir `pwd`" \
+	  --define "releasever $(RELEASEVER)" \
 	  welder-web.spec
 
 rpm: dist-gzip
@@ -34,6 +41,7 @@ rpm: dist-gzip
 	  --define "_srcrpmdir `pwd`" \
 	  --define "_rpmdir `pwd`/output" \
 	  --define "_buildrootdir `pwd`/build" \
+	  --define "releasever $(RELEASEVER)" \
 	  welder-web.spec
 	find `pwd`/output -name '*.rpm' -printf '%f\n' -exec mv {} . \;
 	rm -r "`pwd`/rpmbuild"
@@ -100,9 +108,9 @@ cockpit-test: shared build-rpm
 
 	sudo docker run -d --name web --restart=always -p 9090:9090 --network welder welder/web-cockpit:latest
 
-# Clean generated intermidiate tar file and useless RPM file
+# Clean generated intermediate tar file and useless RPM file
 # RPM file is inside docker image already
-	rm -f welder-web*.rpm welder-web.tar.gz
+	rm -f welder-web*.rpm welder-web*.tar.gz
 
 	sudo docker run --rm --name welder_end_to_end --network welder \
 	    -e COCKPIT_TEST=1 \
--- a/utils/cockpituous-release
+++ b/utils/cockpituous-release
@@ -1,0 +1,25 @@
+# This is a script run to release welder-web through Cockpituous:
+# https://github.com/cockpit-project/cockpituous
+
+# Anything that start with 'job' may run in a way that it SIGSTOP's
+# itself when preliminary preparition and then gets a SIGCONT in
+# order to complete its work
+
+RELEASE_SOURCE="_release/source"
+RELEASE_SPEC="welder-web.spec"
+RELEASE_SRPM="_release/srpm"
+
+release_source() {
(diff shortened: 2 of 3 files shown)


martinpitt:
Note that this also needs a [fix in cockpituous](https://github.com/cockpit-project/cockpituous/pull/169).

martinpitt:
Not sure about the failed travis test:
```
Error: Transaction check error:
  file /usr/lib/rpm/macros.d/macros.cmake from install of cmake-rpm-macros-3.10.1-4.fc26.noarch conflicts with file from package cmake-data-3.10.0-1.fc26.noarch
```
This seems unrelated?

Is that a known thing, or something new? Also, does the "welder-web-jslave-1" thing run the same tests, or different ones?

martinpitt:
Ah, the "welder-web-PR-runtest" one now failed for the exact same reason.

martinpitt:
I have a workaround for the cmake package bug, but building the RPM in the docker container does not yet work. Marking as work in progress.

martinpitt:
rpmbuild was broken due to a PatternFly regression: https://github.com/patternfly/patternfly/issues/917

I added a workaround for that too now.

martinpitt:
rpmbuild and test now succeeds locally and on travis; on Jenkins it fails  rather curiously:
```
rpmbuild -bb \
  --define "_sourcedir `pwd`" \
  --define "_specdir `pwd`" \
  --define "_builddir `pwd`/rpmbuild" \
  --define "_srcrpmdir `pwd`" \
  --define "_rpmdir `pwd`/output" \
  --define "_buildrootdir `pwd`/build" \
  --define "releasever jenkins-welder-web-PR-runtest-19" \
  welder-web.spec
error: line 6: Illegal char '-' (0x2d) in: Version: jenkins-welder-web-PR-runtest-19
make: *** [Makefile:37: rpm] Error 1
```
Looks like jenkins somehow creates a git tag named "jenkins-welder-web-PR-runtest-19" in the repo? why does it modify the git repo that way, I wonder..

Makefile
```diff
@@ -1,3 +1,9 @@
+# generate version number from the latest git tag; if the topmost commit is
+# tagged (at release time), just use that; if there are commits after that,
+# append a ".x" suffix to indicate that's a development snapshot
+# if there is no tag (yet), use 0 as version
+RELEASEVER=$(shell (git describe || echo 0) | sed 's/-[0-9]\+-g.*/.x/')
```
bcl: I think this needs to be a bit more complex (or just don't build without a proper version tag). We should be following semver.org rules for versions.

Makefile
```diff
@@ -8,18 +14,21 @@ install: all
 	mkdir -p /usr/share/cockpit/welder
 	cp -r public/* /usr/share/cockpit/welder
 
+# remove evil file name that breaks rpmbuild (https://github.com/patternfly/patternfly/issues/917)
```
bcl: Comment should be next to the find ... so that when the bug is fixed it's obvious what to revert.

martinpitt: Ack, fixed.

Makefile
martinpitt: Can you please expand what you mean with "more complex"? This doesn't make many assumptions about how release versions look like, other than that they have to start with a number (otherwise rpm complains).

I added the fallback to "0" for the current case where we dont' yet have any release (chicken and egg problem).

martinpitt:
I pushed a commit that shows lots of git state, to debug what's going on in the Jenkins build. Work In Progress until then.

martinpitt:
Indeed:
```
git tag -l
jenkins-welder-web-PR-runtest-20
```
So these need to be filtered out. Fix pushed.

martinpitt:
Yay, green now.

martinpitt:
Note, I created PR #189 to land the CI fixes quickly, as they should be straightforward and unblock other work. Then I can rebase this one to only contain the CD, and also work on other stuff.

martinpitt:
Force-pushed to rebase after #189 landed, so this PR now only has the relevant changes.

Makefile
```diff
@@ -1,3 +1,9 @@
+# generate version number from the latest git tag; if the topmost commit is
+# tagged (at release time), just use that; if there are commits after that,
+# append a ".x" suffix to indicate that's a development snapshot
+# if there is no tag (yet), use 0 as version
+RELEASEVER=$(shell (git describe --exclude '*jenkins*' || echo 0) | sed 's/-[0-9]\+-g.*/.x/')
```
bcl: This still needs to be fixed to conform to semver, if there are new commits it really should bump the patch number and add -rc so that it evaluates after the previous version.

martinpitt: These tarballs are not ever supposed to be public, so how precisely to mark them as "not for public use" is fairly theoretical - However, as you insist: Just incrementing the last digit would be too presumptuous IMHO. After an actual 3.0 release someone might want to release a 3.0.1, so an intermediate version needs to be smaller than that, and not be called 3.1-rc (which would be greater than 3.0.1). Also, these are certainly not release candidates.

So how about appending "-alpha.0" instead? That should be compatible with semver, never sort after the next official version number, and does not make some of the strong semver assumptions, i. e. that should be compatible with rpm version comparison, too. (I hope - e. g. dpkg disallows dashes in version numbers, need to try with rpm).

martinpitt: Yep, same with RPM:
```
rpmbuild -bs \
  --define "_sourcedir `pwd`" \
  --define "_srcrpmdir `pwd`" \
  --define "releasever 0.1-alpha.0" \
  welder-web.spec
error: line 6: Illegal char '-' (0x2d) in: Version: 0.1-alpha.0
```
As this is mostly just an test-internal vehicle, do we really need to bother?

bcl: As long as there's no chance of releasing these I guess it is ok.

PR #188, opened 2018-01-19
Fix timeout issue when running end to end test in local environment.

henrywang:
This issue is caused by saving code coverage result file to a
non-exist folder(/tmp/.nay_output).

To fix this issue, two check points were added.
1. The code coverage result file will not be genreated and saved
if code coverage is not enabeled.
2. If there is no /tmp/.nay_output folder in current system, create
one.

### DIFF
--- a/test/end-to-end/utils/coverage.js
+++ b/test/end-to-end/utils/coverage.js
@@ -12,13 +12,17 @@ module.exports = {
       .evaluate(() => window.__coverage__)
       .end()
       .then((cov) => {
-        const strCoverage = JSON.stringify(cov);
-        const hash = require('crypto').createHmac('sha256', '')
-          .update(strCoverage)
-          .digest('hex');
-        const fileName = `/tmp/.nyc_output/coverage-${hash}.json`;
-        require('fs').writeFileSync(fileName, strCoverage);
-
+        if (cov) {
+          const fs = require('fs');
+          const strCoverage = JSON.stringify(cov);
+          const hash = require('crypto').createHmac('sha256', '')
+            .update(strCoverage)
+            .digest('hex');
+          const covOutDir = '/tmp/.nyc_output/';
+          if (!fs.existsSync(covOutDir)) fs.mkdirSync(covOutDir);
+          const fileName = `${covOutDir}coverage-${hash}.json`;
+          fs.writeFileSync(fileName, strCoverage);
+        }
         done();
       })
     .catch(err => console.log(err));


henrywang:
The failed cockpit-test is due to dnf installation issue. It passed the other tests.

martinpitt:
@henrywang : My fix for the dnf issue landed, so if you rebase and force-push, it should work.

henrywang:
Thanks @martinpitt. Rebased on PR. It's green now!

martinpitt:
@henrywang : Except that it's not - the "welder-web-PR-runtest" one failed in the same way as it used to before.. You might have looked when there was only the "jslave-provision" job?

henrywang:
@martinpitt: I fixed the issue which caused all of cockpit-test cases failed.
Root cause:
There's a substitute running only on Travis, not on Jenkins.
```shell
sed -i "s,var welderApiHost=.*,var welderApiHost=\"api\";," ./public/js/config.js
```
Fix:
Move this sed expression into makfile, then it works on both Jenkins and Travis.

Makefile
```diff
@@ -83,6 +83,7 @@ end-to-end-test: shared
 	sudo docker ps --quiet --all --filter 'ancestor=welder/web-with-coverage' | sudo xargs --no-run-if-empty docker rm -f
 
 build-rpm:
+	sed -i "s,var welderApiHost=.*,var welderApiHost=\"api\";," ./public/js/config.js
```
martinpitt: IMHO this is wrong here. `build-rpm` is not exclusively a test thing, one would expect to use `make srpm` and `make rpm` for RPMs that you can actually install on some machine. Thus this sed should move into `Dockerfile` and become an environment variable so that the `end-to-end` and `cockpit` tests can set that, but "production" usage stays unaffected.

martinpitt: @henrywang : BTW, in #190 I introduce a change to pass a different port to the cockpit and welder-web containers. Something similar can/should be used here (i. e. configure it at container run time, not statically in the rpm or container image).

henrywang: @martinpitt: I created a new PR #191, which removed some lines of ugly sed codes the above included.
I'll make above change out of this PR to make this PR only for timeout issue in local environment.

henrywang:
#188

martinpitt:
Same CI error as in #191 - @atodorov, can you please have a look?

    fatal: unable to access 'https://gitlab.cee.redhat.com/atodorov/ci-ops-composer.git/': Failed connect to gitlab.cee.redhat.com:443; Connection timed out

martinpitt:
@atodorov : Seems that network problem fixed itself, so unping.
@henrywang : On Jenkins the test now "almost" succeeded, the remaining failure looks like a flake. I tried to rebuild the Jenkins test as you suggested [here](https://composer-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/welder-web-PR-trigger/58/), but it doesn't work for me. Can you please retry?

henrywang:
Thanks @martinpitt. Only one case failed on cockpit test. Retried Jenkins test already.

PR #190, opened 2018-01-22
WIP: Run integration tests against lorax-composer

martinpitt:
lorax-composer will be the designated backend for RHEL/CentOS 7, so
run integration tests against that. This communicates over an Unix
socket instead of a TCP one, so this needs some more configuration:

 * Add a new welder/lorax-composer-api container that installs the
   current lorax-composer COPR on CentOS 7.
 * Add a `PORT`/`API_URI` environment variable parameter to the
   web-cockpit and end-to-end containers for passing the port.
 * Move the api container creation out of the `shared` make target into
   two new lorax-composer-api and bdcs-api targets.
 * Split the cockpit test into two new `cockpit-test` and
   `cockpit-test-lorax` targets, and have them depend on the
   corresponding API.

Locally this works in general, it configures the ports and containers directly and most tests actually  succeed:
```
Test Suites: 2 failed, 3 passed, 5 total
Tests:       6 failed, 53 passed, 59 total
Snapshots:   0 total
Time:        385.223s
```
However, there are several timeouts. Some of  them might be cured by #188, but I figure the first timeout is because the first /projects/list call to lorax-composer takes very long due to having to read the yum db first. So this will require a longer timeout. Also, there might be some legitimate failures here if lorax-composer is lacking some API (or has bugs).

I want to see how this  fares on the CI infra, and get some initial feedback.

CC @bcl

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ env:
     - COMMAND=unit-test
     - COMMAND=end-to-end-test
     - COMMAND=cockpit-test
+    - COMMAND=cockpit-test-lorax
 script:
   - make "$COMMAND"
 after_success:
--- a/Dockerfile.cockpit
+++ b/Dockerfile.cockpit
@@ -8,5 +8,5 @@ RUN rm -rf /usr/share/cockpit/kubernetes/
 COPY welder-web*.rpm /tmp/
 RUN dnf -y install /tmp/welder-web*.rpm && rm -f /tmp/welder-web*.rpm
 
-CMD ["/usr/libexec/cockpit-ws", "--no-tls"]
+CMD sed -i "/welderApiPort=/ s_=.*;_='${PORT:-4000}';_" /usr/share/cockpit/welder/js/config.js && exec /usr/libexec/cockpit-ws --no-tls
 EXPOSE 9090
--- a/Dockerfile.lorax-composer
+++ b/Dockerfile.lorax-composer
@@ -1,0 +1,14 @@
+FROM centos:7
+LABEL maintainer='"Martin Pitt" <mpitt@redhat.com>'
+
+RUN yum install -y epel-release && \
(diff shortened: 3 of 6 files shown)


martinpitt:
@atodorov : Travis now runs the new test as I added it to .travis.yml. But how to get this into Jenkins? Is this all configured statically on the Jenkins job side, or somewhere in welder-web git?

FTR, the new test is "known broken" as it just exposes some actual issues when running welder-web against lorax-composer - but if the Jenkins tests are designated to be the successor of travis tests, they need to catch this as well.

atodorov:
> But how to get this into Jenkins? 

@martinpitt Jenkins only knows how to do `make ci` and `make ci_after_success` so everything is tested sequentially. We may want to add a new Jenkins job for testing with lorax-composer though.

atodorov:
LGTM, let me know how do you want this integrated with Jenkins

martinpitt:
@atodorov: Perfect, thanks for pointing out! So I'll just add the new test to the `ci` make target. As said, this is currently supposed to fail as welder-web does not completely work with lorax-composer just yet. For now this is just the CI part of the work, but I don't think we should merge it like that yet.

martinpitt:
This just got blocked more now, as most recent version of lorax-composer is not supported to run in a docker container any more (http://weldr.io/Experimenting-with-lorax-composer/). So this needs to be re-thought completely. We can't install lorax-composer into the test VM itself, as that's Ubuntu 14.04. We could try and create/run a VM, but that would take a lot of time, need a lot of extra RAM/disk, and I'm not even sure if Travis supports nested KVM. @atodorov , would that work better in our Jenkins VMs? What OS do these run, could the tests just install the COPR on the "host" (i. e. the test VM)?

atodorov:
As agreed with Martin on IRC I cherry-picked the first 2 commits and will partially apply the 3rd one as I work on getting a Jenkins build machine ready. Closing.

PR #192, opened 2018-01-24
Update end to end test config and export related cases according to PR#185.

henrywang:
1. Remove package version in config.json.
2. Update package to httpd.
3. Add empty component checking to find issue when depsolve not working.

### DIFF
--- a/test/end-to-end/config.json
+++ b/test/end-to-end/config.json
@@ -13,7 +13,7 @@
             "name": "automation",
             "description": "UI automation recipe",
             "modules": [],
-            "packages": [{"name": "xz", "version": "5.1.2"}, {"name": "whois", "version": "5.1.1"}]
+            "packages": [{"name": "httpd", "version": ""}]
         }
     },
     "__commentComposition": "// Composition type and architechture on Create Composition Page",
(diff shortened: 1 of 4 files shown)


test/end-to-end/config.json
```diff
@@ -13,7 +13,7 @@
             "name": "automation",
             "description": "UI automation recipe",
             "modules": [],
-            "packages": [{"name": "xz", "version": "5.1.2"}, {"name": "whois", "version": "5.1.1"}]
+            "packages": [{"name": "httpd", "version": ""}]
```
martinpitt: Can you please put a rationale into the commit log? I. e. not the *what*, but the *why*? What's wrong with testing a particular version, and why not test with two packages?

atodorov: +1

henrywang: @martinpitt, @atodorov, thanks for your review. I'll add more descriptions into commit.

henrywang:
The #185 removed big metadata db file, used a light db instead.
The new db doesn't include xz and whois package, so the old setting didn't work anymore.
The new db only install httpd and its dependences, so I put httpd into settings.
But the test cases with old settings did not fail. Why?
The new db doesn't have xz and whois, so the dependences will be empty and dependences number will be 0 both from UI and API fetch result. The old test case just compared the displayed info on the UI and info fetched from API, but it ignored one scenario - 0. If both of info are 0, the case will be passed.
So I updated test case to avoid 0 situation. Only two reasons which will cause 0 dependence:
1. depsolve not working, that will be a bug. It'll not be tested by this case.
2. a recipe with selected component which doesn't have any dependence. This situation will not exist, because the recipe in e2e test adds httpd as component. The httpd does have dependences.

PR #194, opened 2018-01-26
Add some features to improve performance and failed case debug.

henrywang:
1. Screenshot for failed case and upload to S3.
2. Add error log output for failed case.
3. In some cases, remove Jest timeout and add Nightmare.js timeout to speed up test time.

Add a lifecycle rule on weldr bucket. That rule will delete screenshot after 7 days.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ env:
   - secure: Q9X11ehcl4YE0ylxDGk8NgVeO7nIpqWiI9am2GA6JVXcxeYYcLf3NHGe1JiLwFQ/xSq4KJrY0IgpivkVrPwTOojpR/tRG5QNDR/IJzN/6PV0IAWkHjPDKZexCQmy4IoPFib/cWqxo46V/JZ+gzs/TJugkl5/QpTcGlCr+d7VkuA80Bj0nlQO5GevYVgYMCrN6u6cKNOWtNQ0gppxHrpsLo36MpHqHbNbv236GPGsV4K2qpiqvvO5ogD+LCv5xoKEZK0ext4FNrrZFq6XEYdy9uUaUbSFfoG+HXESXbIt0L7krQDT/PkKI0X4nM91eYl5mS63+HVue7vLiijS5h9+YCa8HblMIiAiyrR3lPPH0Hx4FVSrQfkvtoXMBs3nRt7keu2FG370a7WoaQ71Q2jHrI9cfYzRwa/Mok73eF/oeN32hPlPDD679hWOZYqbSksSTikCLIgwJpEgIfaHgEYXFmL2JFXk0HbgIciA+C9fE0rfaUbmARtWYq8c+b52aIxAHEV9zYy3JeARrYxk7s3jQgS/LZIw4rfso+bQIvEMUDWt8MnvKaecY1E7XjOpW1kuk8gVjZbHm6oHKl2IndZA9KNpXjT2ew8d3SB69l2lswUYu4mTdywNEUt6nIaQtROiiftUKYThAbibIEIrq7GxQSOYMUPbvUYsZO5rkVssJps=
   - secure: K5mPGLWUFqX/Q0aNoUgYOONB9WKkd98FyMRX5UQQS2Q/lT4/OZWPezQ1I81ufK53r4VLXspL/EILNFXPHb8MEh5CfiIutL/39c/fAFY7XR9EzsuR/DdyIhuWRG4YhqlMU+g2F3zXqcZGgkrLQlvRlCsUp23DBBKZA76EgARMyg4TwVoNbd8vNRwuuo6qIh9d7YT2WcXv/fkF+lH36rjViC2TPFOmgOTkkdumLw/xCxOWK3IBXIPP295pq+Uo/kk6rXUvKrB4Feg2mBd+knvHVVKeu4RDqn2gOz6StP6z085mtr4EhMG4DMT1rGG9KoYnOIaH+cPBCtII+z0Snl3BXCEINR4CpyrrIuJEI5GGWNohr52KaN4i/5eS7Fn6M6ilTB0ohzvSCBEjl5EcV764yHVZJ69hKEBg4Goch17jwXL8vdFB8fPEWBqmAo7PBwXpbfTZCGgHJrpUpbAeJj77IZ7CFx2nOZSe1tsqLP7eCj7P+w9l2AHUMDoPbPstukLMkNpUpEI7ybzZCIJwWTq8KzsnjhFOPA17CF61IFKfyUD600grs2Ye7SyiVDxHoRo0IOMKhFreanwUNc9oxuVUuM/bfCpL9sm9Lxg8/bo2tGrUTb7LfKZOyi93ijHvJRCpzyf4F7otNoYFxNBHj45/xlYVO+4weKncJ3qbjAEE5DE=
   - secure: KixSETk7puLVyxEfCfvDJHl0us1X0CR9Xzeh90sz73JlnFcduv9EaJv/r9bzfjz/dnEmsuaScHJQ8mF0gayQi4ood1oB5ywWkNDfCGhgRG2BmW9FliDU4ukQRDqJq9peNb1TOD3x1LPL4dyNO4Yq8h7pR36KVTeD4hqJfnvlvdNiJ9tk/n6LaR9AFDzkyDfAqsEDqqk/fShw7CociLTsEZ4OTtX9dP/egQm0OA8KYHGxTubKI1Qn6U/4jZn85TajVfEg9HawasjJnFGURbwmCujiB2tLaMCJ+6uflkhyeW5zBKY0zBbMiSZU9lP+enkro6M9JyckGjrBDTxQTn1l0u557wdAzn7kIsyXdwfkqabjh3bicCA9xFkREyIpzAt8yNyqXZrsKJLURz2PVwmwI9ZEpE4hNfRtt3qAK98LYrtvW1JnRbaAci79TLBEfzaLNvn7+ZWXBzsoFTDrbaU70ox3gKcTtast9JtYNcWNNyGXJY6xCRYo9qPJ6vDkQnplA026zioP8q+Kuq2l5WRs4HmtoXxbmJpGya1dVgH/8giG7VU7Xf9q/7bW7e6ytM/UlJhheqJFC9aHBfIe53PNo8fL5XgjCFcAAXQ29i0X+ye6258rD0b9vswkkxQbo69dyI9fR/GPcndL3/VNfeZNTFG9BYoI5pwFnb5JOJoUfwo=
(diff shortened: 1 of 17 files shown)


henrywang:
The uploading to S3 will not work on PR due to [Travis PR security restriction](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions).
But this part of result can be found from https://travis-ci.org/henrywang/welder-web/builds/334544033. There'll not have this kind of issue, because it's not a PR on my repo, but a pushed branch.

.travis.yml
```diff
@@ -1,73 +1,82 @@
...
+  - node_modules
 before_install:
-  - |
-        git config --global user.email "travis@welder.web"
-        git config --global user.name "Travis CI"
-        git remote add welder https://github.com/weldr/welder-web.git
-        git fetch welder
-        git log -1 --pretty=format:%H welder/master
-        git rebase welder/master
+- |
```
atodorov: Can you get rid of these whitespace changes. They don't add anything wrt functionality and honestly cramming all statements to begin on the first column don't make yaml any more readable. This goes for the entire .travis.yml file b/c it's very hard to spot changes when every line has been updated with whitespace.

.travis.yml
```diff
@@ -1,73 +1,82 @@
...
+  - secure: KixSETk7puLVyxEfCfvDJHl0us1X0CR9Xzeh90sz73JlnFcduv9EaJv/r9bzfjz/dnEmsuaScHJQ8mF0gayQi4ood1oB5ywWkNDfCGhgRG2BmW9FliDU4ukQRDqJq9peNb1TOD3x1LPL4dyNO4Yq8h7pR36KVTeD4hqJfnvlvdNiJ9tk/n6LaR9AFDzkyDfAqsEDqqk/fShw7CociLTsEZ4OTtX9dP/egQm0OA8KYHGxTubKI1Qn6U/4jZn85TajVfEg9HawasjJnFGURbwmCujiB2tLaMCJ+6uflkhyeW5zBKY0zBbMiSZU9lP+enkro6M9JyckGjrBDTxQTn1l0u557wdAzn7kIsyXdwfkqabjh3bicCA9xFkREyIpzAt8yNyqXZrsKJLURz2PVwmwI9ZEpE4hNfRtt3qAK98LYrtvW1JnRbaAci79TLBEfzaLNvn7+ZWXBzsoFTDrbaU70ox3gKcTtast9JtYNcWNNyGXJY6xCRYo9qPJ6vDkQnplA026zioP8q+Kuq2l5WRs4HmtoXxbmJpGya1dVgH/8giG7VU7Xf9q/7bW7e6ytM/UlJhheqJFC9aHBfIe53PNo8fL5XgjCFcAAXQ29i0X+ye6258rD0b9vswkkxQbo69dyI9fR/GPcndL3/VNfeZNTFG9BYoI5pwFnb5JOJoUfwo=
```
atodorov: Why do we have 2 more secure strings? What variables are in there ?

test/end-to-end/test/test_createRecipe.js
```diff
@@ -22,15 +20,14 @@ describe('Create Recipe Page', () => {
...
-    helper.gotoURL(nightmare = new Nightmare(), recipesPage);
+    helper.gotoURL(nightmare = new Nightmare(pageConfig.nightmareTimeout), recipesPage);
   });
 
   describe('Input Data Validation Test', () => {
-    describe('Required Field Missing #acceptance', () => {
-      test('should show alert message by clicking Save button when create recipe without name @create-recipe-page', (done) => {
+    describe('Required Field Missing', () => {
+      const testSpec1 = test('should show alert message by clicking Save button when create recipe without name',
+      (done) => {
```
atodorov: Can you explain why remove the # and @ tagging from test case names? 

Also why do test specs need to be defined as constants and assigned to variable names?

atodorov:
Something like this will update the statuses on GitHub and you can add links to every screenshot. You may have to play around the escaping a little bit, b/c I've taken this snippet from another script which does tons of shell and json escaping.

```
REFERRER_COMMIT="$TRAVIS_COMMIT"
if [ -n "$TRAVIS_PULL_REQUEST_SHA" ]; then
    REFERRER_COMMIT="$TRAVIS_PULL_REQUEST_SHA"
fi

curl -s -X POST                                       \
       -H "User-Agent: weldr team CI"        \
       -H "Authorization: token $GITHUB_API_TOKEN"  \
       -d '{                                             \
               "state": "'$test_result'",         \
               "target_url": "screenshot URL", \
               "description": "Job '$TRAVIS_JOB_NUMBER' for '$TRAVIS_REPO_SLUG' has finished",    \
               "context": "'$TRAVIS_REPO_SLUG' '$TRAVIS_JOB_NUMBER' '$COMMAND'"     \
           }'                                            \
       https://api.github.com/repos/$TRAVIS_REPO_SLUG/statuses/$REFERRER_COMMIT
```

test/end-to-end/test/test_createRecipe.js
henrywang: Two reasons of removing @ and #:
1. The file name of screenshot comes from the full name of test case (describe+test). Some of test cases has long test name, which cause a very long filename.
2. the # and @ should not be in filename, sometimes # and @ will cause problem.

To avoid these issues, I removed # and @. But the function is still working with test file name and describe name. That means if I'd like to run edit recipe page tests only, just append "-- editRecipe.js" at the end of npm run test.

The const variable is to get full test name which will be used as screenshot file name.

.travis.yml
henrywang: These are AWS access key and AWS secret access key. They are required by Artifacts.

.travis.yml
henrywang: The yaml file format was changed by travis. In order to add AWS keys I use `travis encrypt SOMEVAR="secretvalue" --add` command. This command did all these changes. I guess this format will be the standard format for .travis.yml.

henrywang:
Thanks @atodorov for review!

> Is this really "in some cases" or in all cases? If we're doing this change I think it should be for all available tests, shouldn't it ?

Yes, it's for all test cases. Here I just gave an failed example to prove the new workflow work.

> I see some screenshots already uploaded but they aren't public and can't be viewed in a browser.

I found the weldr bucket is private. Can I add a new public bucket in S3 to store screenshoot?

> Also it would be very handy to actually make these appear as GitHub checks. I can send you a code snippet to do this.

Thanks for your snippet, I'll take some time to understand it. I don't quite understand meanings of some Travis variables.

.travis.yml
atodorov: I understand now, unfortunately you will have to change it back. 

The travis command line tool doesn't respect white space when dumping back the resulting YAML and overwrites the existing file. The easiest way is to copy the 2 new lines with secure strings, checkout the file and add them again.

.travis.yml
atodorov: 👍

test/end-to-end/test/test_createRecipe.js
atodorov: > The const variable is to get full test name which will be used as screenshot file name.

isn't there something like `this` awailable inside of the test spec ? From what I can tell the spec variable is passed as parameter to the helper.gotoError function.

atodorov:
FYI all of the Travis variables are described here.
https://docs.travis-ci.com/user/environment-variables/

Dockerfile.cockpit
```diff
@@ -1,6 +1,9 @@
 FROM fedora:latest
 MAINTAINER Xiaofeng Wang <xiaofwan@redhat.com>
 
+# to avoid the JS code doing checks we create this here
+RUN mkdir -p /tmp/failed-image
```
martinpitt: Can you please be more specific why this needs to be done explicitly? It adds an entire extra docker layer and thus makes everything slower. docker's `-v` option already creates target directories in the container, so this shouldn't be necessary? If it is, then please `&&` it to an existing `RUN` to avoid the extra layer.

test/end-to-end/README.md
```diff
@@ -37,30 +37,41 @@ $ npm install                                   # Install end-to-end dependencie
...
+```
 
 Run end-to-end test like this:
 
-`sudo docker run --rm --name welder_end_to_end --network welder_default weld/end-to-end:latest xvfb-run -a -s '-screen 0 1024x768x24' npm run test`
+```shell
+sudo docker run --rm --name welder_end_to_end --network host \
+ -v`pwd`/failed-image:/tmp/failed-image \
+ welder/web-e2e-tests:latest xvfb-run -a -s '-screen 0 1024x768x24' npm run test -- --verbose
+```
```
martinpitt: Keeping this doc in sync with the Makefile seems like a tedious and error-prone task. Why not simply refer to the Makefile targets?

test/end-to-end/pages/main.js
```diff
@@ -6,6 +6,6 @@ module.exports = class MainPage {
     this.mailUrl = pageConfig.root;
     this.title = title;
     this.iframeSelector = '.container-frame';
-    this.iframeLoadingTime = 3000;
+    this.iframeLoadingTime = 500;
```
martinpitt: This seems like making it unnecessary tight for loaded test machines?

test/end-to-end/README.md
henrywang: Yeah, good suggestion. Updated already. Thanks.

Dockerfile.cockpit
henrywang: Updated. Thanks.

test/end-to-end/pages/main.js
henrywang: I know it's a ugly hardcode time wait. I have no idea how to avoid this. Before enter into a iframe, the page has to wait the iframe loaded. I tried lots of `wait(<page element>)`, it does not work unfortunately. Do you have any suggestion on this? Thanks.

martinpitt: I wasn't complaining about the fact that it's hardcoded, just that this change seems to make the timeout too short. On loaded infrastructure 3s seems more robust to me than 0.5s.

henrywang: Thanks for your suggestion.
The first reason is I try to speed up test duration. This wait does not need to wait the whole page loaded, just make sure iframe loaded, not page inside iframe. I've tried 3, 2, 1 and 0.5s, all of them work.
Another reason to short this wait time is it has to be shorter than default wait time. I setup the default wait time to 3s, so this wait time has to be shorter than 3s.
I'll make more test on this to keep test stable with a shorter timer.

martinpitt: @henrywang : Ah, thanks for the explanation. I misunderstood this then. This sounds rather brittle - chromium and the DevTools protocol can tell you when a page or an iframe is fully loaded, so there is no reason to have to use a static timeout. However, this isn't making things worse, so fine for me for now.

henrywang: @martinpitt, thanks for your reply. We can talk DevTools protocol offline.

test/end-to-end/test/test_createRecipe.js
henrywang: `this` does not work here, because the test object has to be returned by `test()`or `it()` function and there's no object variable referencing to test object. So I have to create a variable to get test object to get full name.

.travis.yml
```diff
@@ -68,6 +71,31 @@ after_success:
                 docker push welder/web-cockpit
             fi
         fi
+after_failure:
+  - |
+        if [ "$TRAVIS_TEST_RESULT" ]; then
+            CASE_NUMBER=0
+            for SCREENSHOT in `ls ./failed-image`
```
atodorov: Aren't you missing a `;` at the end of this line, just before `do` ?

Also please move this into a separate script (called from make ci_after_failure) so we can make use of it in Jenkins if we decide to.

atodorov:
@henrywang drop the last commit that demonstrates the failures and we're ready to merge.

henrywang:
Thanks @atodorov, @martinpitt for reviewing! Reseted to last success commit.

PR #197, opened 2018-02-16
Terminology updates

jgiardino:
Includes the following updates:

- Terminology updates
  - Use "blueprint" instead of "recipe"
  - Use "image" instead of "composition"
  - Use "commit" instead of "save"
- Icon updates
  - Replace the icon that previously displayed for "composition" with a new icon
  - Replace placeholder icon for recipe with a placeholder icon for blueprint. (This icon will get updated when the final icon is added to the patternfly set)

The terminology updates are based on discussion with product management and marketing after usability tests revealed that "recipe" and "composition" did not work well.

I replaced these terms not only in the UI but throughout the project wherever they're used. Please let me know if anyone has concerns with this - @bcl @atodorov @henrywang @jkozol @martinpitt


henrywang:
Thanks @martinpitt!
@jgiardino, The failed cases on Travis are caused by #196. I filed a mail in mail list yesterday. If there's no response by the end of this week, I'll submit a PR to correct that broken issue.

martinpitt:
@henrywang : Please feel free to send a PR right now - #196 was merged with failing tests (and should thus never have been merged in the first place). So first step is to push a revert of that, and then it can be reimplemented in a working state.

henrywang:
@martinpitt Sure, PR#202 is ready to review. It fixed broken issue.
I'll re-run this PR on Travis again when #202 got merged.

henrywang:
@jgiardino and @martinpitt here's a concern about API. Do we need to update API's terminology If the terminology got finally updated? Such as:
``` shell
/api/v0/recipes/*
/api/v0/compose/* 
```

PR #198, opened 2018-02-22
Run cockpit/welder container on CentOS

martinpitt:
This provides a more conservative platform with an older Cockpit API,
to ensure welder does not use API that isn't supported on CentOS/RHEL 7
yet.


martinpitt:
end-to-end-tests and cockpit-tests are  broken in current master, and thus also in this PR, so I figure this can't yet land soon (but also, not that urgent). But I did some manual testing on the running "web" container and it works quite fine.

martinpitt:
I misunderstood this (see internal discussion on JIRA), I'll rework things to run in one container instead.

PR #200, opened 2018-02-23
WIP: Added modal to Discard Changes

jkozol:
When clicking the Discard Changes button a modal will pop up. This modal will give the user the ability to view or discard all of their pending changes. This is a work in progress because the functionality works but the UI needs to be updated.


jgiardino:
This UI flow is no longer needed. The user can undo the Discard Changes action from the ui.

PR #203, opened 2018-02-26
Simplified component state

jkozol:
As a developer, there should be minimal duplicate data in the state. For recipes, we currently have components, dependencies, and packages which all contain similar data. This will be simplified to just components and packages. We then filter our selected components and dependencies from the lists of packages and components.

### DIFF
--- a/components/ListView/BlueprintContents.js
+++ b/components/ListView/BlueprintContents.js
@@ -27,7 +27,7 @@ class BlueprintContents extends React.Component {
     return (
       <div>
         <Tabs
-          key={components.length + dependencies.length}
+          key={components.length}
           ref={c => {
             this.pfTabs = c;
           }}
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -92,8 +92,6 @@ class ComponentDetailsView extends React.Component {
   }
 
   initializeBootstrapElements() {
-    // Initialize Boostrap-select
-    $('.selectpicker').selectpicker();
     // Initialize Boostrap-tooltip
     $('[data-toggle="tooltip"]').tooltip();
   }
@@ -277,7 +275,12 @@ class ComponentDetailsView extends React.Component {
             </form>
           </div>}
         <div>
-          <Tabs key="pf-tabs" ref="pfTabs" classnames="nav nav-tabs nav-tabs-pf" tabChanged={e => this.handleTabChanged(e)}>
+          <Tabs
+            key={this.state.dependencies.length}
+            ref="pfTabs" 
+            classnames="nav nav-tabs nav-tabs-pf"
+            tabChanged={e => this.handleTabChanged(e)}
+          >
             <Tab tabTitle="Details" active={this.state.activeTab === 'Details'}>
               <h4 className="cmpsr-title">{this.state.componentData.summary}</h4>
               <p>{this.state.componentData.description}</p>
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -52,12 +52,12 @@ class ComponentDetailsView extends React.Component {
     const build = status === 'available' ? 'all' : '';
     // if the user clicks a component listed in the inputs and it's in the blueprint,
     // then use the version and release that's selected for the blueprint component
-    const selectedComponent = Object.assign({}, component);
-    if (selectedComponent.active === true && selectedComponent.inBlueprint === true) {
-      selectedComponent.version = component.version_selected;
-      selectedComponent.release = component.release_selected;
+    const activeComponent = Object.assign({}, component);
+    if (activeComponent.active === true && activeComponent.inBlueprint === true) {
+      activeComponent.version = component.version_selected;
+      activeComponent.release = component.release_selected;
     }
-    Promise.all([MetadataApi.getMetadataComponent(selectedComponent, build)])
+    Promise.all([MetadataApi.getMetadataComponent(activeComponent, build)])
       .then(data => {
         this.setState({ componentData: data[0][0] });
         this.setState({ dependencies: data[0][0].dependencies });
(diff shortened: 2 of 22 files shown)


henrywang:
LGTM.
All tests are passed in my local env.
The Travis test will be re-runed when #202 got merged.

henrywang:
@jkozol, This PR has code conflict with master after #197 got merged. Could you please fix the conflict? Thanks.

jgiardino:
@jkozol This is awesome! I'd like to spend more time reviewing, but have a few questions for you...

- When I look at the redux dev tools, I see that blueprint `packages` are included in the list of blueprint `dependencies`. Can you point out where in the code these packages get filtered out when displaying the dependencies?
- One of the updates I'd like to make in the UI is to show the circle around components that are listed in the blueprint, and alter the circle styling based on whether the component is selected or a dependency. And do this wherever these components are listed, including the list of available components (e.g. a dependency that is listed in the available components list would have `inRecipe: true` and `user_selected: false`). I'm not expecting these UI changes as part of this PR, but I just want to check that these updates support that update (i.e. conditionally add a class based on `user_selected: false`).
  - Example of circle around a selected component: 
image
  - Example of circle around a dependency:
image
- I remember you mentioned something related to this, but currently, when I click a blueprint component list item to expand and view details, I see a list of dependencies. But if I click the component name to view the full details panel, I see no dependencies. Is it possible to have dependencies listed in both places? Are there issues or concerns with doing this? When you mentioned this, I thought you were describing the reverse (i.e. no dependencies listed in the list item, but still listed in the component details panel)
  - Example of expanded list item (Dependencies count is 22)
image
  - Example of details panel (Dependencies count is 0)
image
- I get different results when I click a selected component in the list of Available Components on the left versus the list of Selected Components on the right.
  - Example of Component Details when clicking name in list on left (this is correct—this is how the panel should display for a selected component):
image
  - Example of Component Details when clicking name in list on right (this is incorrect—this is how the panel should display for a component that has not been added to the blueprint yet):
image

- When I opened a blueprint that had pending changes from a previous session, things did not work at all. I was able to create a new blueprint, and things worked fine though. I'm assuming that's more of an issue with old blueprints with bad contents rather than an issue with the workspace, but just thought I'd mention it. I'll continue reviewing this along with the save workflow and let you know if I notice anything.

henrywang:
@jgiardino, do you have any comment on this PR? Thanks.

jkozol:
@jgiardino 

> One of the updates I'd like to make in the UI is to show the circle around components that are listed in the blueprint, and alter the circle styling based on whether the component is selected or a dependency. And do this wherever these components are listed, including the list of available components (e.g. a dependency that is listed in the available components list would have inRecipe: true and user_selected: false). I'm not expecting these UI changes as part of this PR, but I just want to check that these updates support that update (i.e. conditionally add a class based on user_selected: false).

I have added these classes to the the input components when we load them through the inputs saga.

> I remember you mentioned something related to this, but currently, when I click a blueprint component list item to expand and view details, I see a list of dependencies. But if I click the component name to view the full details panel, I see no dependencies. Is it possible to have dependencies listed in both places? Are there issues or concerns with doing this? When you mentioned this, I thought you were describing the reverse (i.e. no dependencies listed in the list item, but still listed in the component details panel)

The dependencies should load now regardless of if you select the component from the left or right panel.

> I get different results when I click a selected component in the list of Available Components on the left versus the list of Selected Components on the right.

I believe this is fixed. I updated the blueprints to make sure they had the inBlueprint and/or user_selected classes.

> When I opened a blueprint that had pending changes from a previous session, things did not work at all. I was able to create a new blueprint, and things worked fine though. I'm assuming that's more of an issue with old blueprints with bad contents rather than an issue with the workspace, but just thought I'd mention it. I'll continue reviewing this along with the save workflow and let you know if I notice anything.

Are you still experiencing this issue?

jgiardino:
> I have added these classes to the the input components when we load them through the inputs saga.

Awesome!

> The dependencies should load now regardless of if you select the component from the left or right panel.

This works as expected, except the selected component is also included in the list of dependencies. Can you remove that one from the list? For example, in the image below, `basesystem` is the selected component but it also displays in the list of it's dependencies.

I also have some questions for you about the list of dependencies for an individual component as shown in the example below:

- Are you able to know when a dependency of one component is also a selected component? For example, in the image below, `bash` is a dependency of `basesystem` but it's also a selected component. If this is not there already and not easy to add for this PR, we can capture it as a future enhancement.
- I eventually need to update the list of dependencies shown below so that the user click a dependency and open the Component Details view. I can't find where the list of dependencies shown below is in the state. I'm assuming clicking a dependency in this list would work the same as clicking a dependency in the Dependencies tab, but just want to confirm that's the case and is possible to add later.

image

> I believe this is fixed. I updated the blueprints to make sure they had the inBlueprint and/or user_selected classes.

Yes, this is awesome. 

This is more of a rhetorical question, and not something we need to address now, but I'm curious what your thoughts are on the structure of `inputs.selectedInput`.  `inputs.selectedInput.status` is kind of a duplicate of `inBlueprint` and `user_selected`, but this information is only available on the first component. Ideally, if I click on a dependency of a component, it would be nice to be able to indicate if that dependency is also `user_selected`.

> Are you still experiencing this issue?

No

Finally, I should have mentioned this before, but it would be nice to standardize on property names. It looks like the preference is camelCase. If it's easy enough to do a find and replace, could you update `ui_type` and `user_selected` to be camelCase.

jkozol:
@jgiardino 

> This works as expected, except the selected component is also included in the list of dependencies. Can you remove that one from the list? For example, in the image below, basesystem is the selected component but it also displays in the list of it's dependencies.

This should be fixed now.

> Are you able to know when a dependency of one component is also a selected component? For example, in the image below, bash is a dependency of basesystem but it's also a selected component. If this is not there already and not easy to add for this PR, we can capture it as a future enhancement.

This should be possible but may be better captured in a future enhancement.

> I eventually need to update the list of dependencies shown below so that the user click a dependency and open the Component Details view. I can't find where the list of dependencies shown below is in the state.

These dependencies are actually fetched from the api on an as needed basis per the component being viewed.

> could you update ui_type and user_selected to be camelCase.

This should be fixed now.

henrywang:
@jgiardino, do you have any comment on this PR? Thanks!

jgiardino:
Thanks for the updates! This looks good to me!!

PR #206, opened 2018-03-02
Add 5 cases on view blueprint page.

henrywang:
1. should show correct blueprint description title
2. should show correct blueprint name under Details tab
3. should show correct blueprint description under Details tab
4. should update description
5. should not update description if click cancel button

### DIFF
--- a/test/end-to-end/package.json
+++ b/test/end-to-end/package.json
@@ -6,6 +6,7 @@
     "test": "jest --runInBand --forceExit"
   },
   "dependencies": {
+    "faker": "^4.1.0",
     "jest": "^20.0.4",
     "nightmare": "^2.10.0",
     "nightmare-iframe-manager": "0.0.4",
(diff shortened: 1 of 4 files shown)


test/end-to-end/package.json
```diff
@@ -6,6 +6,7 @@
     "test": "jest --runInBand --forceExit"
   },
   "dependencies": {
+    "faker": "^4.1.0",
```
atodorov: Why do we need this?

atodorov:
@martinpitt can you review?

test/end-to-end/package.json
henrywang: We need to generate a random blueprint description sentence to update for every test run. That'll avoid some cache issue which will impact string match with cached value but not new value.
faker is a good tool to generate all kinds of random things, like word, sentence, etc.

atodorov: In this case please update the commit log to describe why the commit adds a new dependency.

henrywang: OK, Thanks.

PR #207, opened 2018-03-02
Include modal confirmation for Delete Blueprint

jgiardino:
This PR includes the following changes:

* Changes the text "Archive" to "Delete" for the action that deletes a blueprint
* Includes a Delete Blueprint modal that displays when the user clicks Delete for a blueprint in the list of blueprints. Previously, this action immediately made the api call to delete the recipe. Now this action is associated with the red button on the modal.


henrywang:
@martinpitt, do you have any comment?

jgiardino:
Thanks, @henrywang!

PR #209, opened 2018-03-05
Replace MAINTAINER with LABEL instruction in Dockerfile.

henrywang:
MAINTAINER is deprecated now. As doc suggested, use LABEL instead.
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


henrywang:
Sure. Thanks @bcl.

PR #215, opened 2018-03-16
Add a make target to test with lorax-composer

atodorov:


atodorov:
@martinpitt @henrywang for the pendingCommit test I wait for the components to be shown under Selected Components and that seems to timeout in Travis. For lorax-composer depsolve takes more than 30 seconds and I've bumped the timeout there to 90 seconds. Should I do the same for the bdcs-api backend ?

Makefile
```diff
@@ -85,6 +83,11 @@ end-to-end-test: shared
 	sudo docker build -f Dockerfile.with-coverage -t welder/web-with-coverage:latest .
 	sudo docker run -d --name web -p 3000:3000 --restart=always --network welder welder/web-with-coverage:latest
 
+	until curl http://localhost:4000/api/status | grep '"supported":true'; do \
```
bcl: Might be better to use curl --connection-timeout here so that it can't get completely stuck.

Makefile
```diff
@@ -115,6 +118,11 @@ cockpit-test: shared build-rpm
 # RPM file is inside docker image already
 	rm -f welder-web*.rpm welder-web*.tar.gz
 
+	until curl http://localhost:4000/api/status | grep '"supported":true'; do \
```
bcl: Same comment here.

Makefile
```diff
@@ -130,4 +138,42 @@ ci_after_success:
...
+	sudo docker build -f Dockerfile.nodejs -t welder/web-nodejs:latest .
+	sudo docker build -f ./test/end-to-end/Dockerfile -t welder/web-e2e-tests:latest ./test/end-to-end/
+
+# execute lorax-composer in the background to serve as the API backend
+# making the socket accessible to the cockpit-ws group
+	sudo mkdir /recipes
+	sudo lorax-composer --group cockpit-ws /recipes &
+
+# wait for the backend to become ready
+	until sudo curl --unix-socket /run/weldr/api.socket http://localhost:4000/api/status | grep 'db_supported": true'; do \
```
bcl: And here.

atodorov: This part here is executed on the RHEL host running lorax-composer and with curl-7.29.0-46.el7.x86_64 this option is not available. I will see of it is available for the above 2 places which are executed inside newer environments (Ubuntu/Travis).

atodorov: Turns out --connection-timeout isn't supported in none of our environments.

martinpitt: How about

    for retry in `seq 60`; do curl [...] && exit 0; sleep 1; done; echo  "TIMED OUT" >&2; false

?

atodorov: Not against this but I think 60 seconds is not enough for lorax-composer. This will also depend on the local system that we happen to get for testing and the size/count of available repositories.

We can definitely fix this in a follow up commit though.

PR #217, opened 2018-03-23
Add case to test different button status according to blueprint changes.

henrywang:
1. Commit button should be in disabled stauts
2. Discard Changes button should be in disabled stauts
3. Commit button should be in enabled stauts
4. Discard Changes button should be in enabled stauts
5. Pending Changes link should not exist
6. Pending Changes link should be displayed


henrywang:
I'll merge this PR because it passed Travis tests.

PR #218, opened 2018-03-27
Update timeout options

atodorov:
Updates on how the test suite does timeouts because with lorax-composer these have to be extended significantly.

### DIFF
--- a/test/end-to-end/utils/helper.js
+++ b/test/end-to-end/utils/helper.js
@@ -16,7 +16,7 @@ module.exports = {
     }
   },
   gotoError: (error, nightmare, testSpec) => {
-    console.error(`Failed on case ${testSpec.result.fullName} - ${error}`);
+    console.error(`Failed on case ${testSpec.result.fullName} - ${error.toString()} - ${JSON.stringify(error)}`);
     const dtStr = new Date().toISOString()
                             .replace(/T/, '-')
                             .replace(/:/g, '-')
--- a/test/end-to-end/utils/helper.js
+++ b/test/end-to-end/utils/helper.js
@@ -8,7 +8,9 @@ module.exports = {
     if (pageConfig.root.includes('9090')) {
       nightmare
         .goto(page.url)
-        .wait(page.iframeLoadingTime)
+        // note: waiting for the iframe to load should be less than waitTimeout
+        // with the default values this is around 3 seconds
+        .wait(0.6 * pageConfig.nightmareOptions.waitTimeout)
         .enterIFrame(page.iframeSelector);
     } else {
       nightmare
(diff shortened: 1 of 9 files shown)


test/end-to-end/utils/helper.js
```diff
@@ -8,15 +8,15 @@ module.exports = {
...
+        .wait(page.iframeSelector)
         .enterIFrame(page.iframeSelector);
     } else {
       nightmare
         .goto(page.url);
     }
   },
   gotoError: (error, nightmare, testSpec) => {
-    console.error(`Failed on case ${testSpec.result.fullName} - ${error}`);
+    console.error(`Failed on case ${testSpec.result.fullName} - ${error.toString()} - ${JSON.stringify(error)}`);
```
henrywang: Why "${JSON.stringify(error)"?
From the result log, this did not work here.
Failed on case View Blueprint Page Single Word Blueprint Name Scenario Menu Nav Bar Check should show a blueprint name - Error: .wait() timed out after 5000msec - {}

atodorov: When testing with lorax-composer Cockpit is served via https and initially I was getting an error because of untrusted SSL certificate. Without JSON.stringify the only thing I was shown was [object Object]. Below is an example with both error logging methods enabled:

```
  console.error utils/helper.js:19
    Failed on case View Blueprint Page Single Word Blueprint Name Scenario Menu Nav Bar Check should show a blueprint name - [object Object] - {"message":"navigation error","code":-501,"details":"ERR_INSECURE_RESPONSE","url":"https://localhost:9090/welder#/blueprint/automation"}
```

In other cases, like you point out the error object is logged as string and JSON.stringify returns an empty dictionary. I think this is due to how these error objects are created but this isn't something under our control. That's why I've added both logging methods hoping to be able to get the most error information as possible.

henrywang: Make sense to me. Thanks @atodorov.

henrywang:
@atodorov, iframeLoadingTimeout cannot be set to the same number as wait timeout. If the hard code wait a timeout time, it'll timeout. iframeLoadingTimeout should be less than wait timeout. Setup a new global iframeLoadingTime in config.js to 3000.

atodorov:
> iframeLoadingTimeout cannot be set to the same number as wait timeout. If the hard code wait a timeout time, it'll timeout. iframeLoadingTimeout should be less than wait timeout. Setup a new global iframeLoadingTime in config.js to 3000.

@henrywang how about setting iframeLoadingTime to `0.7 * waitTimeout` ?  with the default values this is a bit more than 3 seconds.

I really would like to avoid defining numerical constants for settings that sort of depend on each other.

PR #220, opened 2018-04-02
Switch to using the blueprints API instead of recipes.

bcl:
This requires an API server that has also made the change.

Merging this needs to be coordinated with the related PRs from lorax-composer, bdcs-api, and bdcs-cli.

### DIFF
--- a/core/apiCalls.js
+++ b/core/apiCalls.js
@@ -56,15 +56,15 @@ export function fetchBlueprintInputsApi(filter, selectedInputPage, pageSize) {
 
 export function fetchBlueprintNamesApi() {
   const blueprintNames = utils.apiFetch(constants.get_blueprints_list)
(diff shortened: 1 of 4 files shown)


jgiardino:
When I review the updates, I don't see anything that was missed. But I also am not able to test this locally with the new api. I haven't been able to get the new api set up on my machine. 

Should we see if @jkozol can test this with his local build?

jkozol:
This looks good to me. I am working on running the tests through lorax-composer. @bcl Do you have instructions on how to build and start the bdcs-api so I can run the UI tests against it?

atodorov:
@jkozol see #215, it's currently blocked by the switch to the Haskell backend and #221 but let's keep in touch once these two are done.

atodorov:
cherry-picked and merged as part of #225

PR #222, opened 2018-04-08
Add a case to verify icon of avaliable component.

henrywang:
1. should show correct icon according to component status


henrywang:
PR passed both Travis and Jenkins test. Merged.

PR #225, opened 2018-04-11
Use haskell api backend

atodorov:
Expect stuff to fail.

### DIFF
--- a/test/end-to-end/config.json
+++ b/test/end-to-end/config.json
@@ -3,8 +3,8 @@
     "api": {
         "uri": "http://localhost:4000/",
         "test": "api/v0/test",
-        "newBlueprint": "api/v0/recipes/new",
-        "deleteBlueprint": "api/v0/recipes/delete/",
+        "newBlueprint": "api/v0/blueprints/new",
+        "deleteBlueprint": "api/v0/blueprints/delete/",
         "moduleInfo": "api/v0/modules/info/",
         "moduleListTotalPackages": "api/v0/modules/list?limit=0&offset=0"
     },
--- a/test/end-to-end/config.json
+++ b/test/end-to-end/config.json
@@ -36,9 +36,10 @@
     },
     "__comment3__": "// Nightmare.js timeout settings",
     "nightmareOptions": {
-      "waitTimeout": 5000,
-      "gotoTimeout": 1000,
-      "loadTimeout": 1000,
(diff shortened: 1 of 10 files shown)


test/end-to-end/config.json
```diff
@@ -36,9 +36,10 @@
...
     "nightmareOptions": {
-      "waitTimeout": 5000,
-      "gotoTimeout": 1000,
-      "loadTimeout": 1000,
-      "executionTimeout": 1000
+      "show": "true",
+      "waitTimeout": 90000,
+      "gotoTimeout": 30000,
+      "loadTimeout": 60000,
+      "executionTimeout": 60000
```
martinpitt: OOI, is it only slow the first time when it reads the yum db, or always? In the former case, a better workaround might be to prime the cache before starting the test, with e. g. a curl query to /projects/list ?

test/end-to-end/test/test_pendingCommit.js
```diff
@@ -95,7 +95,7 @@ describe('Changes Pending Commit Page', () => {
     const testSpec3 = test('should show correct pending change content',
       (done) => {
         // Highlight the expected result
-        const expectedAction = 'Removed';
+        const expectedAction = 'Added';
```
martinpitt: Is that because on the previous bdcs-rs backend we had a bunch of predefined recipes, but the haskell one doesn't?

atodorov:
> but what about the cockpit test on travis? That just seems to be stuck, and the tests don't even start?

I just pushed a commit for that which limits the maximum time we're waiting for the iframe to load to 5 seconds.

test/end-to-end/test/test_pendingCommit.js
atodorov: Nope, the Haskell backend also uses the same predefined recipes. They are bind-mounted to the backend container. 

From what I can tell this test has been just broken. Maybe the call to `updateComponent` has something to do with revealing this issue. I don't really know.

test/end-to-end/config.json
atodorov: I'm not sure I understand the question here but the code is not entirely correct either. 

These are the timeouts which worked for me before the 100% CPU load issue was fixed on the backend. I tried playing around with lower values today but couldn't find such that will consistently pass on my local system. 

These timeouts don't have anything to do with testing on lorax-composer though.

PR #228, opened 2018-05-02
Internationalize welder-web using react-intl

dashea:
There's a lot missing. These commits can hopefully get things started, by providing the framework for marking translatable messages and retrieving the translations for those messages from Zanata.

The basic idea with react-intl is that anything translatable should be wrapped in a <FormattedMessage> component, and I'll write up a blog post with more of the details on how to go about that. I went ahead and did that for everything I could, though I probably missed some here and there. Additionally, I did not touch two of the components files because I think they might need some extra work to make them translatable: components/Modal/PendingChanges.js appears to split sentences up across columns, which doesn't really work. components/Pagination/Pagination.js also has a bunch of markup interleaved with the text; this one might be usable as-is, but I got pretty confused about what is being displayed and how, so someone else can handle that one.

Besides those components and any other strings that I missed, the main thing missing is actually selecting a language. I don't know how to do that, and I don't know how cockpit handles that, so that's something that needs to be solved. Related to that, react-intl requires that locale data be loaded for any languages being used (like the import at https://github.com/weldr/welder-web/compare/master...dashea:i18n-react-intl?expand=1#diff-7a9076d6d94e62c13d641aa71f19ae8eR7 plus the addLocaleData call a few lines down), so we'll need some way of defining or generating that list of locales to load.

There is another npm module, 'react-intl-redux', that combines the react-intl <IntlProvider> and the redux <Provider>, so that might also be something we want to use?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+build/
 node_modules
 public/index.html
 public/dist
--- a/components/CardView/CardView.js
+++ b/components/CardView/CardView.js
@@ -1,6 +1,7 @@
 /* global $ */
 
 import React from 'react';
+import {FormattedMessage} from 'react-intl';
 import PropTypes from 'prop-types';
 
 class CardView extends React.Component {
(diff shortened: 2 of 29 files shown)


dashea:
Something I forgot to mention: I set this up so that every instance of a string gets its own "context" when pushed to Zanata, which allows them to be translated separately. That might not be necessary. For instance, if five components have the string "Edit" in there somewhere, translators will see five different "Edit" strings to localize, and could localize them independently.

In anaconda we used contexts to handle strings with embedded keyboard shortcuts, so that, say, "_Edit" could be translated to Spanish as "_Editar" on one page, but translated instead to "E_ditar" on another so that it doesn't interfere with the shortcut for "_Eliminar". That might not actually be useful here? It can be turned off by taking the `-c id` options out of the `rip` commands.

dashea:
https://github.com/weldr/weldr.github.io/blob/master/_drafts/Transling-welder-web-strings.md is a draft of the blog post for how to mark translations.

dashea:
Closing this one for now, still working how to work better with cockpit

PR #229, opened 2018-05-02
Convert e2e tests to webdriver.io

atodorov:
This is a rather large PR which does several things at once:

1) Converts all e2e tests to webdriver.io and gets rid of Nightmare.js. The biggest problem with Nightmare is debugging async calls and it doesn't make it very easy to figure out what's wrong. With WDIO we get a built-in test runner (other nice things can be done with it in the future) but more importantly easy to read stack traces with actual line numbers when stuff failed.

WDIO also claims to be synchronous or rather everything is still async but it tries to play smart and present itself as a sync process to the user.

It is also a bit easier to select and interact with components on the page directly from node scope, instead of switching to browser scope.

2) Removes all backend API calls. create/delete Blueprint are replaced with UI automation. This will make it easier to enable testing for lorax-composer and not bother about unix domain sockets.

3) Removed test explicit scenarios which are tested elsewhere or covered by the create/delete automation. Also squashed some related assertions into 1 scenario to reduce execution time a bit.
Also for the various buttons that open pop-up dialogs I've left only 1 full test scenario and the rest just validate that the same popup can be triggered by different buttons.

4) Removed unnecessary bits and adjusted the container infrastructure to be able to run the new test suite.

Notes:

* Appears to work fine locally and in CI (for the stand alone web mode)
* Fails with Cockpit, but some cases do pass. This leads me to think this is related to timeouts or improperly waiting for things on the screen. Still debugging.

I would appreciate a quick review so we can merge by the end of the week,


test/end-to-end/specs/test_createBlueprint.js
```diff
@@ -0,0 +1,110 @@
...
+          .waitForVisible(createBlueprintPage.labelAlertInfo);
+
+        const actualAlertText = $(createBlueprintPage.labelAlertInfo).getText();
+        const actualHelpText = $(createBlueprintPage.spanHelpBlockMsg).getText();
+
+        assert.equal(actualAlertText, createBlueprintPage.varAlertInfo);
+        assert.equal(actualHelpText, createBlueprintPage.varHelpBlockMsg);
+      });
+
+      it.skip('show alert message when creating blueprint without name by clicking Enter', () => {
```
henrywang: Should remove skip here if there's no special reason.

henrywang:
@atodorov , can we enable firefox headless for the test?

test/end-to-end/config.json
```diff
@@ -1,13 +1,4 @@
 {
-    "root": "http://localhost:3000/",
```
henrywang: wdio.conf.js already has the same settings as this one, this file should be removed.

test/end-to-end/specs/test_createBlueprint.js
atodorov: there is, see the comment down below in the test. From what I can tell this is not supported on Firefox.

atodorov:
To address some of the comments while still trying to figure out why stuff fails under Cockpit:

>    Can we enable firefox headless for the test? 

Is there any difference? I will give it a try once I get everything to work reliably but IMO if there's no difference then I prefer to run in full graphical mode because it makes things easier for debugging. There's no need to reconfigure locally before debugging.

> We do not need vnc and just with xvfb-run.

For some reason I was not able to export a DISPLAY for selenium server to use when running with xvfb-run hence the VNC workaround. Again if there isn't much difference between VNC and Xvfb then I'd prefer going with VNC (and graphical Firefox) because you can connect a client and observe what is happening in real time. This isn't essential to the migration though so let's think about it after this PR gets merged.


> 2. The test, test_blueprints.js - should open Edit Blueprint page when clicking Edit Blueprint button, always fail on my local env. 

Any traceback ? Why does it fail? 


I am experiencing some failures with Cockpit but none when running in stand alone mode.

PR #233, opened 2018-05-16
Make the welder-web package more koji and COPR friendly

dashea:
This makes the bundled npm stuff reproducible, allows the spec file to be used without command-line releasever macro stuff, fixes the COPR make_srpm thing to work with the exciting new COPR of today, and as a bonus makes the source archive a little bit less weird.

I built one in copr over at https://copr.fedorainfracloud.org/coprs/dshea/welder-web-test/build/755116/ to make sure it works.

The git-based version stuff doesn't work with copr's make_srpm mode, since it doesn't pull down the tags, but I don't think there's a whole lot we can do about that.


henrywang:
@dashea, could you please add jq in Dockerfile.buildrpm. All of the rpm build job will be run on this docker container.

dashea:
Fiddled with the dependencies in package.json to get fsevents out of the shrinkwrap file, which has always been the problem we've had with it, and fixed a problem with the new Makefile commands. I was able to successfully run `make cockpit-test` locally. This iteration of the travis build failed because COPR is down.

henrywang:
@dashea, I re-run test on Travis successfully. jq works inside buildrpm container.

atodorov:
> @henrywang ok, added. welder-web-lorax-composer-PR-runtest still fails with missing jq (and rpmbuild?). It doesn't look like that test is using a container, so I assume something needs to be added to the jenkins VM.

the slave config lives in git (internally) and the add-on bits are mostly Ansible playbooks. I will happily distribute the URL to others. I was even thinking to move the playbooks inside the project code-base and use them from there.

PR #234, opened 2018-05-21
Less docker files for welder-web

henrywang:
1. Removed Dockerfile.with-coverage and merge it into Dockerfile
2. Delete Dockerfile.nodejs and put node.js download into Dockerfile for welder-web and end to end test.
3. Update Dockerfile for welder-web and end to end test not depend on welder/fedora:latest, but fedora:latest.
4. Update .travis.yml and Makefile to follow above changes.


.travis.yml
```diff
@@ -25,11 +25,8 @@ before_install:
             sudo docker pull welder/web-cockpit:latest
         fi
         if [ "$COMMAND" == "end-to-end-test" ]; then
-            sudo docker pull welder/web-nodejs:latest
             sudo docker pull welder/web:latest
             sudo docker pull welder/web-e2e-tests:latest
-        else
-            npm install
```
atodorov: IMO this should not be deleted. From what I can tell Travis does a `npm install` before it executed the test targets (e.g. eslint) simply because the build is labeled as Nodejs one. 

This is an artifact of the build/test infrastructure which we should not rely upon. 

Either revert or explain why this line needs to be deleted.

Dockerfile.buildrpm
```diff
@@ -1,10 +1,39 @@
...
-# work around https://bugzilla.redhat.com/show_bug.cgi?id=1532293 by upgrading cmake first
-RUN dnf upgrade -y cmake && dnf install -y rpm-build jq && dnf clean all
+
+RUN dnf install -y make cmake rpm-build gnupg tar xz curl jq && dnf clean all
+
+RUN echo 'PATH=/usr/local/bin/:$PATH' >> /etc/bashrc
+
+# Based on official node docker image
+# gpg keys listed at https://github.com/nodejs/node
+RUN set -ex \
```
atodorov: Don't like the repetition of a huge block of code between 3 docker files.

I think the way to go here would be to :+1: 

1) build the web image as first thing in the shared target
2) base Dockerfile.buildrpm on welder/web image and change Makefile target to depend on the shared target
3) Same goes for the test image, better not duplicate the Nodejs configuration part.

.travis.yml
henrywang: npm install is the default command in Travis install stage. So we do not put a duplicated one here.
You can find "npm install" from Travis log of this PR.

Dockerfile.buildrpm
henrywang: Or copy nodejs from Travis to container. But that'll not work in downstream Jenkins test.

atodorov: Errr nope, don't copy anything from an Ubuntu container to Fedora. That's asking for trouble.

.travis.yml
atodorov: Yes, I know this and I don't want to rely on it implicitly. Every command should be clearly visible in our own source code, not hidden into log files or 3rd party documentation.

henrywang:
@atodorov, looks "/usr/bin/rpmbuild" not work.
`/bin/sh: /usr/bin/rpmbuild: No such file or directory`

Makefile
```diff
@@ -53,19 +54,39 @@ stylelint:
 unit-test:
 	npm run test:cov
 
+download-nodejs:
```
atodorov: Just `dnf install nodejs` inside all the containers that need it.

Makefile
```diff
@@ -33,7 +34,7 @@ srpm: dist-gzip welder-web.spec
 rpm: dist-gzip welder-web.spec
 	mkdir -p "`pwd`/output"
 	mkdir -p "`pwd`/rpmbuild"
-	rpmbuild -bb \
+	/usr/bin/rpmbuild -bb \
```
atodorov: Why not then `RPMBUILD=$(shell which rpmbuild)` ?

Makefile
henrywang: That'll install node.js 8.x.x, not 6.9.x.

atodorov: adding the IRC log here b/c you dropped off just before that:

```
(12,50,56) atodorov: as part of the PR to minimize the number of docker files I propose to do `dnf install nodejs` which will pull 8.11 (current LTS) from Fedora 28
(12,51,10) atodorov: prior to that 6.9 has been specified in the files
(12,51,50) atodorov: I did the change and everything seems to build and pass tests. Is there any reason to continue using a soon to be unsupported version diretly from upstream instead of the version coming with Fedora 
(13,13,26) pitti: atodorov: what is "version directly from upstream" here? using the fedora version seems fine of course
(13,14,43) atodorov: pitti: all the docker files do curl https://nodejs.org/dist/***-tar.gz
(13,15,02) pitti: atodorov: oh, uh; I'd prefer the distro ones too indeed
(13,15,24) pitti: atodorov: maybe in the past that was done because fedora's packaged node was too old?
(13,15,27) pitti: but 8.11 certainly seems good
(13,15,44) atodorov: ok, any particular reason to specify 6.9? or was that historical ? 
(13,16,05) pitti: I don't have any knowledge about that; but if 8 works, that seems like a better choice if it's LTS indeed
(13,17,01) atodorov: yes it is, see https://github.com/nodejs/Release#release-schedule
```

henrywang: Reasonable to me! Thanks @atodorov.

henrywang:
@atodorov, still report "which: no rpmbuild in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin)
"

test/end-to-end/Dockerfile
```diff
@@ -1,10 +1,12 @@
...
+FROM fedora:latest
 LABEL maintainer="Xiaofeng Wang" \
       email="xiaofwan@redhat.com" \
       baseimage="Fedora:latest" \
       description="A welder-web end to end test runner container running on Fedora"
 
-RUN dnf --setopt=deltarpm=0 --verbose install -y tigervnc-server which xclip firefox java
+RUN dnf --setopt=deltarpm=0 --verbose install -y gnupg tar xz curl tigervnc-server which xclip firefox java nodejs && dnf clean all
+
+RUN echo 'PATH=/usr/local/bin/:$PATH' >> /etc/bashrc
```
atodorov: This is not needed anymore.

Makefile
```diff
@@ -1,6 +1,7 @@
 # Extract the version from package.json
 VERSION=$(shell $(CURDIR)/rpmversion.sh | cut -d - -f 1)
 RELEASE=$(shell $(CURDIR)/rpmversion.sh | cut -d - -f 2)
+RPMBUILD=$(shell which rpmbuild)
```
atodorov: It appears that this shouldn't be necessary. According to Brian something else is going on with the builders and on RHEL /bin is a hard-link to /usr/bin which I've verified on my system. So let's revert back to using `/usr/bin/rpmbuild` everywhere and I'll look at it separately.

PR #237, opened 2018-05-23
Add a cockpit-composer-* target

bcl:
This is going to be the name of the package for RHEL7, same content,
just a different name, and the Requires on lorax-composer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,11 +24,22 @@ welder-web.spec: welder-web.spec.in
 	    -e 's|@RELEASE@|$(RELEASE)|' \
 	    < welder-web.spec.in > welder-web.spec
 
+cockpit-composer.spec: cockpit-composer.spec.in
+	sed -e 's|@VERSION@|$(VERSION)|' \
+	    -e 's|@RELEASE@|$(RELEASE)|' \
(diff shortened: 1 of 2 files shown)


bcl:
Note that I cannot get the current version (based on master, commit 7e389e5abd18a0d55106530ebe6ca264ad9b6abd) to actually work. The javascript console spews a big pile of errors that I don't even know where to begin debugging. And yes, I did change config.js to point to the socket.

Makefile
```diff
@@ -44,6 +55,20 @@ rpm: dist-gzip welder-web.spec
 	find `pwd`/output -name '*.rpm' -printf '%f\n' -exec mv {} . \;
 	rm -r "`pwd`/rpmbuild"
 
+cockpit-composer-rpm: dist-gzip cockpit-composer.spec
+	mkdir -p "`pwd`/output"
+	mkdir -p "`pwd`/rpmbuild"
+	rpmbuild -bb \
```
atodorov: Can you specify `/usr/bin/rpmbuild` here? There's an issue with the Jenkins builders where sometimes they can't find programs in the standard path. Not sure why is that.

PR #239, opened 2018-05-28
Test target for cockpit composer

atodorov:
Depends on #237.
Adds two targets which are used by Jenkins to schedule RPM builds and do some sanity tests on the resulting RPMs.

ATM we don't do any sanity testing, just build the artifacts.


atodorov:
Finally everything passed and infrastructure isn't acting up. I am merging this.

PR #240, opened 2018-05-31
Internationalize welder-web using react-intl

dashea:
This one actually works. Translations are loaded and used both within cockpit and as a standalone app.

I did not attempt to mark translatable strings in two of the components files:

  * components/Modal/PendingChanges.js appears to split sentences up across columns, which doesn't really work.
  * components/Pagination/Pagination.js has a lot of markup interleaved with text, which might work, but I found it confusing so I'll leave that for someone else

For everything else, I attempted to mark all translatable strings as translatable. I probably missed a few. The Zanata project is at https://fedora.zanata.org/project/view/welder-web. I went ahead and provided translations for the handful of Spanish words I either knew or could copy out of anaconda's translations, so if you set your browser to Spanish you can see parts of this working right now.

I wrote a guide on how to translate strings at https://github.com/weldr/weldr.github.io/blob/master/_drafts/Translating-welder-web-strings.md. Feel free to provide feedback on that, as well. I will move the guide out of the drafts and on to the main site once this PR can be merged.


dashea:
Pushed fixes for the eslint errors

dashea:
Fixed the issues with end-to-end-test: there was a legitimate error with how I was building a couple of translated attributes, there was a problem with one of the test selectors being too general and erroneously picking up one of the new translation spans, and the rest was missing packages required by the new javascript dependencies.

dashea:
Rebased to master

jkozol:
This looks good to me. I can do the i18n for pendingChanges and pagination next week. It should be a good exercise to familiarize myself with the 'react-intl' library. Once done with that I'll hopefully be able to provide feedback on your guide.

README.md
```diff
@@ -164,5 +164,31 @@ This project can be packaged as either a noarch rpm or an srpm.
 $ make rpm                # Or, `make srpm`
 ```
 
+### i18n
+
+For a general guide on how to write translatable strings, see [weldr.io](http://weldr.io/Translating-welder-web-strings/)
```
jgiardino: Have you posted this article yet? I'm not able to find it on weldr.io

dashea: Not yet, it's in _drafts. It's viewable here, with slightly worse formatting: https://github.com/weldr/weldr.github.io/blob/master/_drafts/Translating-welder-web-strings.md

README.md
```diff
@@ -164,5 +164,31 @@ This project can be packaged as either a noarch rpm or an srpm.
...
+will add `id` attributes to all of the messages, and [babel-plugin-react-intl](https://github.com/yahoo/babel-plugin-react-intl)
+will extract all of the messages to JSON files, written to `./build/messages`.
+
+**Step 3**. The developer runs `npm run translations:push`. This uses [react-intl-gettext](https://github.com/rubencosta/react-intl-gettext)
+to collect the JSON files into a gettext-style POT file and uploads the POT file to [Zanata](https://fedora.zanata.org/).
+
+**Step 4**. Translators provide translations on Zanata.
+
+**Step 5**. The developer runs `npm run build`. This downloads the translations from Zanata as gettext-style .po files,
+converts the .po files back to JSON, and bundles the JSON translations with the rest of the project into main.js.
```
jgiardino: I'm able to confirm what's happening up until step 2, but is there any way to verify that steps 3 and 5 are actually working as expected? When I go to https://fedora.zanata.org/, is there a way to see what files were uploaded and what translations were provided? The signup process seems to assume that I'm signing up to be a translator, but really I just want to see what's there and what's not, so that I can compare that to what I'm getting when I build.

Do we have any translated strings available yet?

For step 3, does that POT file exist anywhere in the project?

For step 5, am I supposed to be able to see a file somewhere that has the translated text strings?

README.md
jgiardino: This article is great!

jgiardino:
This is an impressive PR! Thanks for writing up the article on how to create translatable strings. That helps with understanding the rest of this.

I had a couple of comments inline. At this point, I'm not really sure how to test that everything is working as expected. I can review the individual updates to see if they match the examples you documented, but I'm guessing the better option is to have translated strings for everything to make sure they're getting pulled in. But I'm not sure if we do have all strings translated in a specific language yet, or if there's a way to simulate that?

I can take a look at updating the strings used in the pagination component.

README.md
dashea: For the `translations:push` target, to upload the .pot file, you will need a Zanata account. And I'll probably need to add you to the project, I can't really remember how that works. You should be able to run `translations:pull` without an account, though, and that will also download the .pot file that's currently on Zanata. Everything is written to ./build/po/

The .pot file is not in the git repo. It's generated from the strings extracted from the javascript files. You can generate the .pot file without doing a push by running `npm run translations:json2pot`. The .pot file will be written to the same place used by `translations:pull`, ./build/po/welder-web.pot.

We have a handful of translated strings. I added some Spanish translations for the simple stuff, and you can see those in ./build/po/es.po, and the .po file is converted to ./build/poJSON/po.es.js. The .js version is a little confusing since the original English strings are not included, but you can see in there, for instance, that "components.Layout.Header.2222696742", which corresponds to a "Help" string, maps to "Ayuda".

dashea:
@jgiardino for testing, I just added `npm run translations:test`, which creates some auto-generated Pig Latin and saves it as "Volapük". You can then do `node run` like normal, select Volapük as your browser's language, and anything marked as translated will be "translated".

https://dshea.fedorapeople.org/test-translations.png

Also fixed two bugs I found along the way.

README.md
jgiardino: Thanks for the clarification! I'll create an account for Zanata.

I did switch to Spanish, but never saw any translated strings. But maybe I wasn't looking in the right place.

jgiardino:
Glad you were able to catch a few issues. I'm able to see the translations that you included. I think I was using "node run build" instead of "npm run build" before. 

The only way I'm able to connect the UI to the API is by copying the build files over to my VM that's running cockpit with composer. When I do that, I am able to see the Spanish translations, but not the Volapük translations. Maybe that's because the Cockpit Display Language menu doesn't include Volapük? (@martinpitt Would you know why this is happening?) I'm not sure that's an issue, but maybe just something worth noting. 

I am curious how to make changes to the UI when text strings become available and we want to keep those translations. For example I update some components so that the text strings should be the same as before, but the ids are now different because of the files I moved them to.  Is there anyway in Zanata to say that the text string that once had id 'x' should now have id 'y'?

henrywang:
@dashea , could you please resolve conflict and rebase master to make end to end test work. Thanks.

dashea:
@jgiardino For keeping the translations when strings move, it's possible to just ignore the ids when sending the translations to Zanata, it'll just take some reworking of the translation scripts. That way, any time a string "x" appears in welder-web, it will use the translation "y", regardless of location. Would that be preferable?

I don't know if there would ever a string in welder-web that you would want translated differently, depending on where it appears. In anaconda we had that sometimes since some strings contained embedded keyboard shortcut information, but I don't think that would ever be an issue here.

jgiardino:
@dashea I think what you're proposing would work. I'm not sure what you mean exactly by "x" when you say 'any time a string "x" appears in welder-web'. Would "x" be an id for the string, or the string itself? 

My only goal is that we don't identify strings based on the file they're in, because it limits the amount of refactoring we're able to do.

martinpitt:
gettext identifies strings by their value, so if the same C string gets used in different places, they will use the same translations. This can be changed by using different [contexts](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html), but they are rarely used. In other words, what @dashea proposes is a well-tried and good default, and saves a lot of redundant translations.

dashea:
Reworked the JSON to gettext part to remove the msgctxts, and I went ahead and removed all msgctxt info from the Japanese translations on Zanata.

henrywang:
@dashea, could you please update some lines of configurations on .travis.yml to make lorax-test work? Thanks!
1. Add the following block of setting
```
env:
  - CXX=g++-4.8
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
    packages:
    - g++-4.8
```
2. Insert the following line of setting before line 38 - `node run build --with-coverage`
`npm rebuild`

.travis.yml
```diff
@@ -35,14 +37,16 @@ install:
         fi
 
         if [ "$COMMAND" == "lorax-test" ]; then
+            npm rebuild
             node run build --with-coverage
```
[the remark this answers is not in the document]
dashea: I would rather leave the build as it is. `node run build` will create an empty translations.json file instead of pulling from Zanata. This way, Zanata failures and bad translation data cannot cause the tests to fail.

There was an issue where translations.json was not created because I didn't create the directory first. This has been fixed.

PR #245, opened 2018-06-22
Added start compose functionality

jkozol:
This PR includes the removal of redux-persist, some fixes to how we fetch the blueprint contents when on the the blueprint page, and most of the core functionality for creating images. Currently, we don't load all of the images for the blueprints from the api but only the ones created in the current session. Also, when creating an image before we load the blueprint contents, there is a delay before that image appears in the image list. Finally, the undo/redo functionality removes images from the state. These issues will be addressed in a future PR since they are going to rely on how we load the images at the start of a session.


jgiardino:
Adding the [documentation for UI workflow](https://docs.google.com/document/d/1O5AoTYWAoS0oyGHPofew5n3f29O90XgKp092n5eN6xY/edit?usp=sharing) related to layout and messaging here.

jgiardino:
The icon class names look good. I had a couple of comments inline regarding text strings for the different states.

I also noticed that when I had more than two items in the list, that the sort order would change every minute or so. I expect the images to be sorted in the order they were created, with the newest listed first.

andreasn:
Probably outside the scope of this PR, but I don't get any kind of estimate on how the build is going. The notification just sits there and spins and spins and spins. It seems I can's get that from composer-cli either. Is this something I need to file a ticket against lorax for?

andreasn:
It seems the Cancel action doesn't work.

andreasn:
@jgiardino Is this spinner supposed to spin?

screenshot from 2018-06-29 15-14-09

jgiardino:
Yes. I was thinking I would handle that plus some other minor css updates in a separate PR. I went ahead and noted this in #264.

andreasn:
@jgiardino sounds good!

jkozol:
@andreasn Yes, the cancel button is non-functional right now. I can add the functionality into this PR if you'd like or I can push it in the next one. @jgiardino Where is it documented how we want to handle cancelled composes in the UI? Should I just remove the image or leave an option to rerun it?

andreasn:
@jkozol I think that ideally the UI needs to be in a state where you can put a user in front of a build that came out of the master branch at any point. Having UI elements that doesn't do anything breaks that idea.
I would either hook the button up to cancel the action in this PR, or remove the button from this PR and create a separate PR with both the button and the functionality as a followup.

jkozol:
@andreasn Understood. I removed it from this PR and will add it, with its functionality, in another.

@jgiardino In regards to the list order changing, I would prefer to make an issue and fix this when I implement fetching prior-session images from the api because I am going to refactor how we store images.

bcl:
To clarify what's available from the API -- while the build is running it really isn't possible to get a decent estimate of completeness or remaining time. You can query the queue status, the individual build's status (using the UUID returned when starting it), or you can grab the end of the anaconda.log while anaconda is running.

jgiardino:
> ...or you can grab the end of the anaconda.log while anaconda is running.

Terry had suggested during a previous meeting that we could show something, maybe it was this, in a console component.  I'm open to UI ideas for this, but also am not sure if this is a must-have for the first implementation.

andreasn:
I don't think it's as urgent. So no worries. Mostly wanted it recorded.

jgiardino:
I'm not sure if this [console](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Columns=90&knob-Rows=35&selectedKind=react-console%20%28%40patternfly%29%2FSerial%20Console&selectedStory=SerialConsole&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) would be the exact component we would use, but this is what I was thinking of when Terry made the suggestion. 

Also, maybe I'm mixing up what's available from the anaconda logs vs what would be displayed in the cli. I think Terry was expecting the UI could show what displays in the cli during the build process, and this would only display on demand from the user. This is something that would definitely need some UI exploration for sure.

andreasn:
This doesn't apply against master any more, so needs to be rebased.

andreasn:
But the UI looks good to me now!

jkozol:
I rebased so is this good to merge now?

andreasn:
Works for me now!

PR #246, opened 2018-06-23
core: Use escalated privileges to access Lorax API if possible

stefwalter:
If the logged in Cockpit user can escalate privileges (ie: sudo)
then use this to access the Lorax API.

Without this, users who are non-root and not in the weldr group
will fail access to composer. This failure can be seen in the
Javascript Console and results in:

    cockpitFetch: ... failed: Not permitted to perform this action.

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -12,9 +12,11 @@ welderApiScheme = welderApiScheme || 'http';
 function setupCockpitHttp() {
   const useHttps = welderApiScheme === 'https';
   const port = welderApiPort || (useHttps ? 443 : 80);
+  const superUser = welderApiPort ? "try" : undefined;
   cockpitHttp = cockpit.http(port, {
     address: welderApiHost,
     tls: useHttps ? {} : undefined,
+    superuser: superUser
   });
 }
 


atodorov:

atodorov:

atodorov:

atodorov:

stefwalter:
This is complementary to: https://github.com/cockpit-project/cockpit/pull/9476

The other pull request allows admins to add a user to the weldr group, for users that do not or should not have admin/sudo/privileged access.

core/utils.js
```diff
@@ -12,9 +12,11 @@ welderApiScheme = welderApiScheme || 'http';
 function setupCockpitHttp() {
   const useHttps = welderApiScheme === 'https';
   const port = welderApiPort || (useHttps ? 443 : 80);
+  const superUser = welderApiPort ? "try" : undefined;
```
martinpitt: Why is this conceptionally conditional on setting an explicit port, vs. using an implicit default?

stefwalter: @martinpitt Should I always try for superuser?

martinpitt: My gut feeling is that *if* it is legit to talk to welder as superuser, it should always be. Right now it should be; there once was some discussion about per-user recipes and such, which would be broken if "everybody" (by and large) talked to the API as root.

A better option would maybe be to check if the user already has read/write access to the Unix socket (through the `weldr` group), and only use `try` if not.

bcl: The conclusion that I came to is that if we want to implement per-user recipes that we do that using branches and that it be enforced by the UI not by weldr. Currently nothing is using the branch support (it exists, but is likely not fully tested) so all blueprints are in the master branch.
So, as things stand now, there is no difference between accessing it as a member of the weldr group or as root.

henrywang:
@stefwalter , could you please rebase master to make end to end test work. Thanks.

larskarlitski:
Filed #295 with Stef's patch updated to always require root. Please close this one.

Issue #247, opened 2018-06-25
Update Tabs components


jgiardino:
The current implementation of Tabs is based on the source code for an old patternfly-webcomponents implementation. But a newer Tabs component in patternfly-react is available and should be used instead.

As part of this issue:
- I _think_ we can remove anything related to [webcomponents from webpack.config.js](https://github.com/weldr/welder-web/blob/master/webpack.config.js#L109). 
- The entire [Tabs](https://github.com/weldr/welder-web/tree/master/components/Tabs) folder should be removed
- The [patternfly-react Tabs component](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Tab%201%20Tittle=Tab%201&knob-Tab%202%20Tittle=Tab%202&knob-Tab%203%20Tittle=Tab%203&selectedKind=patternfly-react%2FWidgets%2FTabs&selectedStory=Basic%20Tabs%20-%20PF%20style&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) can replace all instances of the Tabs component, which occurs in the following pages/files:
  - **Blueprint page**, pages/blueprint/index.js
![image](https://user-images.githubusercontent.com/21063328/42388283-fd777c86-8112-11e8-950d-dc0e0f3776dc.png)
  - **Blueprint contents**, components/ListView/BlueprintContents.js
    - This UI displays in two places:
      - On Blueprint page, click Selected Components tab. This tab has a second set of tabs as shown below:
        ![image](https://user-images.githubusercontent.com/21063328/42388316-2543e358-8113-11e8-94bb-e9f8944a597d.png)
      - On Edit Blueprint page, in the Blueprint Components panel:
        ![image](https://user-images.githubusercontent.com/21063328/42389401-9b0c284a-8116-11e8-94e1-79d966967d60.png)
  - **Component details view**, components/ListView/ComponentDetailsView.js 
    - These tabs use the [patternfly style](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Tab%201%20Tittle=Tab%201&knob-Tab%202%20Tittle=Tab%202&knob-Tab%203%20Tittle=Tab%203&selectedKind=patternfly-react%2FWidgets%2FTabs&selectedStory=Basic%20Tabs%20-%20PF%20style&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)
[...]

PR #248, opened 2018-06-25
implements filtering and refactors toolbars

jgiardino:
The following updates are included:

1. Adds ability to filter blueprints by name, and filter selected components and dependencies by name, version and release.
    - Filters are applied by typing in the filter input field and hitting Enter.
    - Applied filters display below the toolbar.
    - Sort order can be modified on filtered contents
    - Filters can be cleared by: 
      - Clicking the X for a single filter value
      - Clicking "Clear All Filters" link below the toolbar
      - Clicking "Clear All Filters" link in empty state message that displays if no content items match the filter criteria
    - When filtering blueprint contents, the same filters are applied to both Selected Components and Dependencies tabs
      - There are a couple of issues with these tabs. But these tabs should also be updated to use the patternfly-react component. A separate issue is created: #247
2. Refactors toolbar components so that blueprints and components share common components.
3. Includes an empty state for blueprints and selected components.
4. Includes a loading state for selected components and available components.
    - Ability to filter available components should be refactored to provide a similar empty state message when the filter returns no results. This is captured in a separate issue: #249


jgiardino:
@henrywang There were some tests that failed, but I wasn't sure how to fix them. I'm happy to try if you have suggestions that will point me in the right direction.

core/reducers/filter.js
```diff
@@ -0,0 +1,70 @@
...
+  },
+
+  [BLUEPRINTS_FILTER_REMOVE_VALUE]: (state, action) => {
+    return updateFilters(state, filtersWithRemovedFilter(state.components.filterValues, action.payload.filter), 'blueprints')
+  },
+
+  [BLUEPRINTS_FILTER_CLEAR_VALUES]: (state) => {
+    return updateFilters(state, [], 'blueprints')
+  },
+};
```
jgiardino: @jkozol The feedback I had received from one of the UXD js devs was to use this pattern instead of a `switch` statement. But when he shared it out with the rest of the team, one of the other devs shared [this example from another product](https://github.com/ManageIQ/miq_v2v_ui_plugin/blob/master/app/javascript/react/screens/App/Overview/screens/MappingWizard/components/MappingWizardClustersStep/MappingWizardClustersStepReducer.js) which is using it. 

I'm happy to change this back to `switch` which seems to be the standard that everyone's using. But I wanted to give you a chance to review and let me know what you think.

jkozol:
This looks good to me. The only issue I'm experiencing is when I add a Name, Version, and Release filter I start getting duplicate filters.

core/reducers/filter.js
jkozol: Thanks for sharing this. Did he explain why he prefers this pattern over switch statements? I don't have a particular preference and just used switches because that's what is mainly used in the redux documentation.

jgiardino: Some of the comments regarding `switch`:

> switch/case is just a weird exception to the normal JavaScript pattern of "code blocks are in curly brackets"
because it's sometimes useful to match multiple patterns and run both branches of code
so braces become less useful there.
But instead people use it as a replacement for `if() { } else if() { } else { }`
because zero braces is better than 6 right?
it's also more confusing. So-- functions. no if-else-if-else, but clear lines between what runs and what doesn't
I am surprised it is not more popular-- people like the syntax that involves the least typing.
I believe `switch` should probably only be used when functions don't work as well.. if you only want exactly one case to run, then it should just be a function in a lookup table

jgiardino:
@jkozol Thanks for catching that! That issue should be fixed with my last commit.

jgiardino:
@andreasn @larskarlitski - Let me know if you have any feedback regarding these updates.

andreasn:
@jgiardino tested it out, and works great here! Ready to merge as far as I'm concerned

jgiardino:
Thanks for the review, @andreasn!

Issue #251, opened 2018-06-27
Dependencies list looks like links, but are not clickable


andreasn:
### Description of problem

If I look at the list of dependencies for a particular blueprint component, the dependencies look click-able, but the links doesn't lead anywhere.

### Component (web, API, etc)

Web.

### How often reproducible

Always

### Steps to Reproduce

1. Create a blueprint
2. Add a component from the "Available Components" list.
3. Under "Blueprint Components", expand that component
4. Try to click the links under "Dependencies"

### Actual results

Links doesn't work.

### Expected results

For them to lead somewhere.

andreasn:
As a quickfix, I suggest removing the a-tag element until we have the code hooked up to lead you to the right destination.

PR #252, opened 2018-06-28
Remove non-functional UI elements/components

jgiardino:
This removes all remaining non-functional UI elements and unnecessary UI components.


jgiardino:
Aside from this and other PRs, the only remaining things I can think of are:
- Navigation - I think there's some logic that hides the navigation when in the context of cockpit. But at the moment, this tool is only available in the context of cockpit. Should we just remove navigation altogether and add it back later when it becomes necessary?
- Ability to comment on a commit - See example below. This UI *is* functional, but it also implies that the comments are viewable somewhere in the UI, which currently is not available.

image

jgiardino:
closes #251

andreasn:
>  Should we just remove navigation altogether and add it back later when it becomes necessary?

Sounds like a good plan to me.

andreasn:
> Ability to comment on a commit - See example below. This UI is functional, but it also implies that the comments are viewable somewhere in the UI, which currently is not available.

Can you see them from the CLI somehow?

jgiardino:
> Can you see them from the CLI somehow?

@bcl Can you comment on that? I thought that `composer-cli blueprints changes http-server` would show comments along with the changes, but I'm not seeing them.

bcl:
> @bcl Can you comment on that? I thought that composer-cli blueprints changes http-server would show comments along with the changes, but I'm not seeing them.

It isn't included in the POST, right? Right now the content is converted directly to a blueprint and the subject of the commit is auto-generated using the version number and name. We haven't added any way to include a commit message along with the blueprint yet. That would require a schema change, instead of posting the blueprint JSON directly you'd wrap it {"comment":"wonderful fixes", "blueprint":{"name":...

andreasn:
This is what I get from the CLI with the `changes` command:

```
[root@m1 ~]# composer-cli blueprints changes newprint
newprint
    2018-06-28T17:29:01Z  f0295738fbc248472e05c3b4421b12ad77f0b304
    Recipe newprint, version 0.0.3 saved.

    2018-06-28T17:28:20Z  07c3e07a12d58d2712847402cfa9ea4a052606eb
    Recipe newprint, version 0.0.2 saved.

    2018-06-28T17:27:50Z  5d897a8c5592521c695dd914c7da7dd70131cadb
    Recipe newprint, version 0.0.1 saved.
```

bcl:
> This is what I get from the CLI with the changes command:

That's the correct output. To implement including comments in the commit we would need API and UI changes.

larskarlitski:
If there's currently no way to get at the commit message, I agree that we should remove the possibility to enter one in the UI.

jgiardino:
I'm not sure what the best approach is for this one. There is more to the Comments field than just an input field that isn't hooked up. If we plan to add the ability to include commit messages, then I'm wondering if we should just hide this element in the UI, and keep the remaining code there so that we can easily hook it up later.

This is the approach I took in my last commit, which results in the UI shown below:
image

But if we think it's best to pull everything related to the commenting feature out, then I'd prefer to capture that in a separate issue and PR. 

Let me know what you think.

bcl:
@jgiardino I think hiding it is the right approach, I don't see any reason to remove code that's eventually going to be useful.

andreasn:
Looks good to me ui-wise.

andreasn:
@larskarlitski can you review the code on this one?

henrywang:
@jgiardino, could you please resolve conflict and rebase master to make end to end test work. Thanks.

jgiardino:
@henrywang I made those changes, but I'm still seeing this test fails.

When I look at the file, this looks incorrect to me, since the Edit Blueprint is only a link:
```
    // Button - Edit Blueprint
    this.btnEditBlueprint = '.list-view-pf-actions a button';
```
Should it be this instead:
```
    // Button - Edit Blueprint
    this.btnEditBlueprint = '.list-view-pf-actions a button';
```
Also, we could add `data-` attributes to make it easier to identify elements for tests, for example:
```
    // Button - Edit Blueprint
    this.btnEditBlueprint = '[data-test="edit-blueprint"]';
```

larskarlitski:
> @larskarlitski, do you have any comment on this PR? Thanks.

This looks good to me. Thanks!

henrywang:
@jgiardino, there's a conflict due to PR merge order. Sorry about that.
Could you please resolve this conflict? Thanks.

jgiardino:
@henrywang Before I rebased, I though all tests were passing. I've tried restarting the build, but that doesn't seem to change the outcome. Do you have any suggestions?

henrywang:
@jgiardino, the failed case is not caused by your code change. Sometimes, Available Components and Selected Components keep on loading status when entering edit blueprint page by entering edit blueprint page directly by URL, like http://localhost:9090/welder#/edit/test.I filed a new issue #286 to track this issue and let it move on.

PR #273, opened 2018-07-03
Domain socket support in UI testing.

henrywang:
1. Remove Welder-web stand-alone UI test.
2. Add domain socket support for Cockpit integrated welder-web.
3. Add domain socket support for end to end test.
4. Fix a page element.
5. Increase wait time to 60 seconds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ shared: metadata.db
 	sudo mkdir -p failed-image
 	sudo docker network inspect welder >/dev/null 2>&1 || sudo docker network create welder
 	sudo docker ps --quiet --all --filter 'ancestor=welder/bdcs-api-img' | sudo xargs --no-run-if-empty docker rm -f
-	sudo docker run -d --name api --restart=always -p 4000:4000 -v bdcs-recipes-volume:/recipes -v `pwd`:/mddb --network welder --security-opt label=disable welder/bdcs-api-img:latest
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -98,7 +98,7 @@ shared: metadata.db
 	sudo mkdir -p failed-image
 	sudo docker network inspect welder >/dev/null 2>&1 || sudo docker network create welder
 	sudo docker ps --quiet --all --filter 'ancestor=welder/bdcs-api-img' | sudo xargs --no-run-if-empty docker rm -f
-	sudo docker run -d --name api --restart=always -p 4000:4000 -v bdcs-recipes-volume:/recipes -v `pwd`:/mddb --network welder --security-opt label=disable welder/bdcs-api-img:latest
+	sudo docker run -d --name api --restart=always -v bdcs-recipes-volume:/recipes -v `pwd`:/mddb -v bdcs-socket:/run/weldr --security-opt label=disable welder/bdcs-api-img:latest
```
clumens: I don't see where bdcs-socket comes from.  Regardless, I wouldn't even use the name "bdcs" in new stuff.  We're not using it right now.

henrywang:
I have another PR which packages lorax-composer and web in one container, then run end to end test against it.
My plan is to run end to end test on two scenarios:
1. welder-web + cockpit + bdcs-api.
2. welder-web + cockpit + lorax-composer.
This PR is for the first scenario. Sorry for making you confuse.

clumens:
Thanks - my initial comment was made when I was only considering the lorax-composer case, and I'm unable to delete it so I'm just going to let it stay.  It doesn't apply to the case of testing bdcs.

henrywang:
The cockpit-test passed. Unit test issue will be fixed in a separate PR.

PR #279, opened 2018-07-05
Unit test fixes for Filter updates

jgiardino:
- updates mockState in unit tests to match state updates for Filters
- fixes issue flagged by tests related to propTypes in ToolbarLayout
- fixes issue related to inconsistent quotation marks in core/store.js

### DIFF
--- a/core/store.js
+++ b/core/store.js
@@ -30,7 +30,11 @@ const initialState = {
         value: '',
     },
   },
-  blueprints : [],
+  blueprints: {
+    errorState: null,
+    fetchingBlueprints: true,
+    blueprintList: [],
+  },
   modals: {
     createImage: {
       name: '',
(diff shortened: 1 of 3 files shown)


jgiardino:
@henrywang Thanks for your feedback about the unit tests! It was very helpful and exactly what I needed. Let me know if you see any issues with these updates.

core/store.js
```diff
@@ -30,7 +30,11 @@ const initialState = {
         value: '',
     },
   },
-  blueprints : [],
+  blueprints: {
```
henrywang: There init state should be an empty array.
From blueprints reducer file core\reducers\blueprints.js, it initialized state to an empty array. they're should be the same or remove blueprints initialization  from store.js.
```javascript
const blueprints = (state = [], action) => {
  switch (action.type) {
    case CREATING_BLUEPRINT_SUCCEEDED:
      return [
        ...state.filter(blueprint => blueprint.present.id !== action.payload.blueprint.id), {
          past: [],
          present: Object.assign({}, action.payload.blueprint,  {
            localPendingChanges: [],
            workspacePendingChanges: {addedChanges: [], deletedChanges: []},
            images: [],
          }),
          future: [],
        }
      ];
```

jgiardino:
Thanks for catching that, @henrywang! I pushed an update. Let me know if there are any other changes needed.

PR #281, opened 2018-07-06
Fix an unit test issue inside create.image.spec.js.

henrywang:
[PR#279](https://github.com/weldr/welder-web/pull/279) fixed the other 3 unit test issues.


larskarlitski:
Ah sorry, I've been working on the same thing in PR #283. It goes a bit further than this one: It make the handlers on `CreateImage` optional, so that unrelated failures like this don't happen anymore. Also, it verifies that `handleStartCompose` is actually called. 

Please have a look.

henrywang:
Right! I'll close this PR. Thanks @larskarlitski.

PR #282, opened 2018-07-06
Fix two issues.

henrywang:
1. Remove docker folder because we do not have stand alone welder-web.
2. Update Makefile with correct test name.


henrywang:
Thanks @clumens.
The Dockerfile, which copies nginx.conf into docker container, has been removed already.

PR #284, opened 2018-07-06
components: Use consistent syntax for handlers

larskarlitski:
The coverage tool complained about this non-standard (?) syntax and failed.

### DIFF
--- a/components/Actions/Actions.js
+++ b/components/Actions/Actions.js
@@ -5,8 +5,11 @@ import utils from '../../core/utils';
 
 class Actions extends React.Component {
 
-  state = { actions: [] };
+  constructor() {
+    super();
 
+    this.state = { actions: [] };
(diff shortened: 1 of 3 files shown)


jgiardino:
These updates look good, but... The actions and wizard components aren't being used currently and were removed as part of #252. The navigation is conditionally displayed depending on whether it's included in cockpit or not, but until we know that it would exist outside of cockpit, I created issue #285 to go ahead and remove that too.  ¯\_(ツ)_/¯

larskarlitski:
> These updates look good, but... The actions and wizard components aren't being used currently and were removed as part of #252.

Ah, sorry about that. I only went through the warnings without paying too much attention if the code is being used. Unfortunately, this is on master already and your branch on #252 has conflicts. Very sorry!

PR #294, opened 2018-07-11
Propagate api errors

### LINKED ISSUE #293: Endless spinner
### Description of problem
Since yesterday, when I open welder-web in Cockpit, I only get an endless spinner.

### Component (web, API, etc)
web

0.1.1

### How often reproducible
Always

### Steps to Reproduce

1. Make sure to run the latest build
2. Open the web UI

### Actual results
It's just an endless spinner

### Expected results
That the Blueprints list would show up, or the empty state

### Additional info
screenshot from 2018-07-11 13-36-32

Console output:
screenshot from 2018-07-11 13-37-38

larskarlitski:
Propagate API errors, so that the UI doesn't show an infinite spinner when an error occurs. This was most visible when logging in with a user that's not root or in the `weldr` group.

See individual commits for details.

Fixes #293

### DIFF
--- a/core/apiCalls.js
+++ b/core/apiCalls.js
@@ -47,7 +47,7 @@ export function fetchBlueprintInputsApi(filter, selectedInputPage, pageSize) {
         }).catch(e => console.log(`Error getting blueprint metadata: ${e}`));
       })
       .catch(e => {
-        console.log(`Failed to get inputs during blueprint edit: ${e}`);
+        console.log('Failed to get inputs during blueprint edit', e);
         reject();
       });
   });
@@ -76,7 +76,7 @@ export function fetchBlueprintInfoApi(blueprintName) {
 export function fetchModalCreateImageTypesApi() {
   const imageTypes = utils.apiFetch(constants.get_image_types)
     .then(data => data.types)
-    .catch(e => console.log(`Error getting component types: ${e}`));
+    .catch(e => console.log('Error getting component types', e));
   return imageTypes;
 }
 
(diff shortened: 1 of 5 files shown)


jgiardino:
@larskarlitski Thanks for making these updates. I'm not sure what the expected outcome is when I look at the UI. 

For example, when I add a new user, log in as that user and navigate to the Blueprints page, I still see the spinner. I also see these errors in the console:
image

Does this match what you're expecting? Or were you expecting an error message to replace the spinner? The error message displays based on the app state, which is currently handled with redux (e.g. an action that triggers a reducer to update the state).

 For example, in core/sagas/blueprints.js, the action `blueprintsFailure(error)` is called when the api fails to return the list of blueprints. This action calls the reducer which updates the state with the error passed by the action.

larskarlitski:
> Does this match what you're expecting?

No, that seems like the old behavior? In fact, I have removed exactly this error message in this patch ([line 35 in `core/utlils.js`](https://github.com/weldr/welder-web/pull/294/files#diff-f2a7728780063ff6d2e72c6bd379684cL35).

> Or were you expecting an error message to replace the spinner? 

Yeah, it should show a full screen error message "An error occurred" with the message itself as a sub heading. (That's not in this patch, I just made sure it is also shown in cases when the weldr api returns an error or is not reachable.)

> The error message displays based on the app state, which is currently handled with redux.

Right, I didn't change that at all. The only change here is that the promise created in `apiFetch()` is rejected properly on errors.

jgiardino:
Thanks for clarifying! I definitely had some issue pulling this branch earlier. I checked again and now see what you're describing as the expected behavior. 

Looks good to me!

PR #295, opened 2018-07-11
core: Use escalated privileges to access Lorax API

larskarlitski:
Update of @stefwalter's patch in #246. Always require sudo, as discussed on that pull request.


henrywang:
@larskarlitski, most of cases are failed due to this changes.

henrywang:
@larskarlitski, I'm so sorry to accidentally close this PR. Re-open it now.

andreasn:
Works for me!

larskarlitski:
Change to use `superuser: "try"` instead of `require`.

PR #298, opened 2018-07-12
Adjustments to how we start tests with lorax-composer

atodorov:
See https://github.com/weldr/lorax/pull/387/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,8 +150,8 @@ ci_after_success:
 	[ -f ./node_modules/codecov/bin/codecov ] || npm install codecov
 	[ -f ./node_modules/nyc/bin/nyc.js ] || npm install nyc@11.1.0
 
-	cat ./coverage/lcov.info | ./node_modules/codecov/bin/codecov
-	./node_modules/nyc/bin/nyc.js report --reporter=lcov && ./node_modules/codecov/bin/codecov
+	[ -f ./coverage/lcov.info ] && cat ./coverage/lcov.info | ./node_modules/codecov/bin/codecov
+	[ -d .nyc_output/ ] && ./node_modules/nyc/bin/nyc.js report --reporter=lcov && ./node_modules/codecov/bin/codecov
 
 # NOTE: this is executed on a RHEL 7 or CentOS 7 host which is
 # VM or bare metal. All required repositories must be configured, e.g.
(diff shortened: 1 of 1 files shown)


bcl:
This should not be needed. The lorax-composer.service has this in it to make sure the socket is always right:
    ExecStartPre=/usr/bin/systemd-tmpfiles --create /usr/lib/tmpfiles.d/lorax-composer.conf

atodorov:
@bcl what if we start lorax-composer manually like we do in the tests ? 

I am seeing the following in the Jenkins console.log:

```
sudo lorax-composer --group cockpit-ws /recipes &
until sudo curl --unix-socket /run/weldr/api.socket http://localhost:4000/api/status | grep '"db_supported": true'; do \
    sleep 1; \
    echo "Waiting for backend API to become ready before testing ..."; \
done;
curl: (7) Failed to connect to /run/weldr/api.socket: No such file or directory
  File "/sbin/lorax-composer", line 277, in <module>
    make_yum_dirs(server.config["COMPOSER_CFG"])
  File "/usr/lib/python2.7/site-packages/pylorax/api/config.py", line 83, in make_yum_dirs
    os.makedirs(p_dir)
  File "/usr/lib64/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/var/lib/lorax/composer/repos.d'
Error in atexit._run_exitfuncs:
  File "/usr/lib64/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/usr/lib64/python2.7/multiprocessing/util.py", line 315, in _exit_function
    p._popen.terminate()
  File "/usr/lib64/python2.7/multiprocessing/forking.py", line 171, in terminate
    os.kill(self.pid, signal.SIGTERM)
OSError: [Errno 1] Operation not permitted
Error in sys.exitfunc:
  File "/usr/lib64/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/usr/lib64/python2.7/multiprocessing/util.py", line 315, in _exit_function
    p._popen.terminate()
  File "/usr/lib64/python2.7/multiprocessing/forking.py", line 171, in terminate
    os.kill(self.pid, signal.SIGTERM)
OSError: [Errno 1] Operation not permitted
```

atodorov:
Note: as a side note, I don't think we even have systemd on the RHEL7 docker image that is used for testing so the above approach may not work.

bcl:
Ah! Sorry, missed that you are starting it manually. Nevermind, looks fine to me if that works for you.

Issue #269, opened 2018-07-02
Upon reload, created images are not loaded


jkozol:
### Description of problem

When a user creates an image it will display in the image list. If the user refreshes the page the image is not loaded and the image list will be empty.

### Component (web, API, etc)

Web

### Version or commit hash (if applicable)

0.1.0

### How often reproducible

Always

### Steps to Reproduce

1. Create an image
2. Refresh page

### Actual results

The list of images is empty

### Expected results

The image should be displayed in the list of images.

### Additional info

PR #299, opened 2018-07-12
Fix created image not being added to state

jkozol:
Created images were not being added to the state.


jkozol:
Sorry, I should have clarified more. This does not address #269. I noticed that no images were being added to the image list and error was coming up in the console because the created image wasn't being added to the state. This fix merely allows created images to be added to the state in the session they were created in.

Issue #235, opened 2018-05-22
Commit, Discard Changes, Create Image buttons covered title bar


henrywang:
![screenshot-2018-5-22 blueprint](https://user-images.githubusercontent.com/672230/40345306-3b317c7c-5dcb-11e8-8055-93c6af1c6fe8.png)

jgiardino:
@henrywang Is this screen capture from the context of Cockpit? Since this isn't a stand-alone application, maybe we should remove the navigation. The current version of patternfly doesn't support the layout I'm using on the Edit Blueprint page, but I'm hoping that the next version of patternfly handles this type of layout better. But for now, since it only is visible in Cockpit, then I think it's fine to remove the navigation.

jgiardino:
@henrywang Can you check this issue again now that #301 is merged?

henrywang:
Thanks @jgiardino. I cannot reproduce this issue with new version due to unix socket api. But from the first page (blueprints page) of stand alone application, the navigation got removed. So It can be closed from my perspective.

PR #301, opened 2018-07-12
Remove unused Layout components

### LINKED ISSUE #285: Navigation component isn't needed
Currently, this tool is only available within cockpit, which has it's own navigation. For now, navigation should be removed and added back later when needed.

This was noted in PR #252 
This would also address issues #235

jgiardino:
- Removes unused components in components/Layout/
- Removes Header from components/Layout/Layout.js and the logic for displaying the Header based on whether this is in the context of Cockpit or not. Header and this logic can be added back if this ever becomes a stand-alone app.
- Updates styles for positioning toast notifications to be visually consistent with PatternFly. PF styles are based on the presence of a header. Updated styles are based on this app loading in an iframe with the header outside the iframe.

This closes #285 and #235


jgiardino:
Thanks for catching that, @henrywang!

henrywang:
@jkozol and @larskarlitski, I'll merge this PR if you do not have any comment on this PR. Thanks!

PR #303, opened 2018-07-17
cockpituous-release: Use upstream release-source

martinpitt:
cockpituous' release-source script recently learned how to build release
tarballs for plain Makefile projects (like welder-web) that use
`make dist-gzip`. So drop our own function for it.


martinpitt:
Tested like this: In a completely clean checkout, I ran

    RELEASE_SOURCE=$PWD/_release/source ~/upstream/cockpituous/release/release-source

and got a valid tarball:

    $ ls -l _release/source/
    -rw-rw-r-- 1 martin martin 6663290 17. Jul 08:12 welder-web-0.1.1.tar.gz

Issue #290, opened 2018-07-10
Refresh clears contents from Images tab


jgiardino:
### Description of problem
After images have been added to the list view on the Blueprint page > Images tab, hitting refresh for the page will result in these contents not displaying.

### Component (web, API, etc)
web

### Version or commit hash (if applicable)
0.1.0

### How often reproducible
Always ?

### Steps to Reproduce

1. Navigate to the Blueprint page, Images tab
2. If no images exist for the blueprint, create one. (Images can only be created if contents are added to the blueprint)
3. After one or more list items are present on the Images tab, hit refresh

### Actual results
The empty state displays for the tab.

### Expected results
The images that were there should still be listed

Issue #271, opened 2018-07-02
Undo/redo removes images


jkozol:
### Description of problem

If changes are made to a blueprint, an image is created, and the blueprint changes are undone/redone the image will appear and disappear from the images list.

### Component (web, API, etc)

web

### Version or commit hash (if applicable)

0.1.0

### How often reproducible

Always

### Steps to Reproduce

1. Go to the edit blueprint page
2. Make changes to the blueprint but don't commit them
3. Create an image (it will display in the images list)
4. Undo the blueprint changes

### Actual results

The image disappears from the images list.

### Expected results

The image remains in the images list since images are independent from workspace changes.

### Additional info

Issue #270, opened 2018-07-02
Image created from blueprints page not displaying for 60s


jkozol:
### Description of problem

When creating an image before the blueprint contents have been loaded, the image will not display in the list of images until the next api poll (60s).

### Component (web, API, etc)

Web

### Version or commit hash (if applicable)

0.1.0

### How often reproducible

Always

### Steps to Reproduce

1. On session start, create an image from the blueprints page.
2. Navigate to the blueprint page for the blueprint the image was created for.

### Actual results

The image appears in the images list 60s after it was created.

### Expected results

The image appears in the images list immediately after it was created.

Issue #272, opened 2018-07-02
Image list order changes


jkozol:
### Description of problem

The order images are listed in changes when the API is polled to check the image status every 60s.

### Component (web, API, etc)

web

### Version or commit hash (if applicable)

0.1.0

### How often reproducible

Always

### Steps to Reproduce

1. Create an image
2. Wait 30s
3. Create a second image

### Actual results

Image 1 will display at the top of the list for 60s then image 2 will display at the top for 30s and they will continue to swap positions every 30s as one of them gets their status updated from the API.
 
### Expected results

Images sort by the order they were created in.

bcl:
FYI on the API side of things the order should be stable until a new build is started. Currently the ordering is going to be the determined by the UUID. The other option is to sort it by the timestamp, which is the time of the last update to the build's status. Or I could change timestamp to be the time the build's metadata was written to disk (ie. the start time).

PR #304, opened 2018-07-17
Add loading images from prior sessions and sort by start time.

jkozol:
Images from prior sessions are now loaded at the start of a session. They will display in the blueprint page's list view sorted by start time. This PR should address issues #290 #269 #270 #271


jkozol:
I have found an issue with polling unfinished composes. I will push a fix tomorrow. This fix should also address #272

core/apiCalls.js
```diff
@@ -176,3 +176,46 @@ export function fetchImageStatusApi(uuid) {
...
+}
+
+export function fetchComposeFailedApi() {
+  const p = new Promise((resolve, reject) => {
+    utils.apiFetch(constants.get_compose_failed)
+    .then(data => {
+      resolve(data.failed);
+    })
+    .catch(e => {
+      console.log(`Error fetching diff: ${e}`);
```
jgiardino: Should this message be more specific for each case?

core/apiCalls.js
```diff
@@ -176,3 +176,46 @@ export function fetchImageStatusApi(uuid) {
   });
   return p;
 }
+
+export function fetchComposeQueueApi() {
+  const p = new Promise((resolve, reject) => {
+    utils.apiFetch(constants.get_compose_queue)
+    .then(data => {
+      const queue = data.new.concat(data.run);
+      resolve(queue);
```
jgiardino: Does this get both composes that are pending plus the one that's in progress?

core/actions/composes.js
```diff
@@ -0,0 +1,34 @@
...
+});
+
+export const FETCHING_COMPOSE_STATUS_SUCCEEDED = 'FETCHING_COMPOSE_STATUS_SUCCEEDED';
+export const fetchingComposeStatusSucceeded = (blueprintName, compose) => ({
+  type: FETCHING_COMPOSE_STATUS_SUCCEEDED,
+  payload: {
+    blueprintName,
+    compose,
+  },
+});
```
jgiardino: Should we also add actions for handling errors for these?

jgiardino:
These updates look good! I can review again when you get the fix in that you mentioned, but I wanted to get your thoughts on something that I noticed. 

In the redux inspector, I'm noticing that we don't display contents on the Blueprints page until we've fetched blueprint contents. But as soon as we have the blueprint names, we're fetching both blueprint contents and composes? And it looks like we're getting information about composes before blueprint contents. Since we don't display contents on this page until blueprint contents are fetched, this results in taking a long time to load contents on the page. 

Is it possible to control the order that the state gets created, so that we wait until blueprint contents are loaded before getting the composes? I'm not sure this is a higher priority compared to other issues, but want to at least capture thoughts around how this could be solved later.

Can you describe how the `composes` array in the state gets it's data in the context of the Blueprints page? Maybe this is where I have a gap in my understanding of how redux-saga gets pulled in. 

BTW, reviewing this highlighted an issue with how I was handling the loading state for the Blueprints page, so I added issue #305.

core/actions/composes.js
jkozol: I can add a error action alongside the fix I'm making.

core/apiCalls.js
jkozol: Yes.

jkozol:
@jgiardino 
In regards to the blueprint contents and composes loading, we should be able to check if the blueprint contents are loaded before calling the selector. If they aren't we can return an empty list and once they are the selector will get called to load the composes. I'll need to look more into this later, however, to make sure.

The way we load compose data is when a user opens the UI all of the composes will be immediately fetched from the API like we do with the basic blueprint info. When you go to a blueprint page a selector (getBlueprintComposes) is called which takes the compose state and selects the composes that correspond to the blueprint being viewed. These are then sorted by start time and the list is passed to the props of the blueprint page.

core/apiCalls.js
jkozol: Probably. Any opinions on how should we handle the error?

jgiardino: I'm assuming "Error fetching diff" was added for /api/v0/blueprints/diff/.

I'd just have them identify the api call that's being made, like "Error fetching finished composes", etc...

Also, updates were made in #294 regarding the syntax for these. This syntax:
```console.log(`Error fetching diff: ${e}`);```
was updated to this:
`console.log('Error fetching diff', e);`

**Comment by henrywang** [comment]
@jkozol, could you please fix the eslint issue? Thanks.
```
/home/travis/build/weldr/welder-web/core/apiCalls.js
  17:12  error  'err' is defined but never used            no-unused-vars
  27:12  error  'err' is defined but never used            no-unused-vars
  32:9   error  'page' is assigned a value but never used  no-unused-vars
  47:18  error  'e' is defined but never used              no-unused-vars
```

jkozol:
@henrywang Oops. Should be fixed now.

henrywang:
@jkozol, I got the "An Error Occurred" and "Internal error" when I load "http://localhost:9090/welder" first time. And web console reports "errorloadBlueprintsSaga". But it doesn't have same issue when I refresh this page. So it caused some end to end test cases failed. Could you please take a look? Thanks!
screenshot_2018-07-19 image builder - cockpit

core/apiCalls.js
larskarlitski: I've always wondered why you're wrapping those calles in another `Promise`. `utils.apiFetch` already returns a promise, doesn't it?

jkozol:
@henrywang I am looking into the test failures. In regards to your errorloadBlueprintsSaga on first load, what environment are you experiencing this in? I am unable to replicate it.

core/apiCalls.js
jkozol: You are correct. I will remove those extra promises from the compose API calls in this PR and will refactor the rest of the API calls in a future fix.

henrywang:
@jkozol, here's my review workflow:
1. checkout PR branch.
2. node run build
3. sudo docker run -d --name web --restart=always -v bdcs-socket:/run/weldr -v `pwd`/public:/usr/share/cockpit/welder --network host welder/web-lorax:latest
4. open http://localhost:9090/welder

jgiardino:
These updates look good!

The only issue that I notice is there was a failed compose already in my vm. When I look at it in the list of images, the Date Created is showing today, but this is one that I had started last week. Maybe timestamps don't really reflect the start time? Maybe this was really an issue with my vm?  If there's a way to check using the cli what the timestamp is, I would, but I'm not sure there is a command for that. Let me know if you have any thoughts on this one. 

As for handling when we get composes (i.e. immediately on page load vs on user action to view composes), I don't think we need to make any changes in this PR. This would be a good topic for discussion next week, and is probably a larger topic that potentially requires other changes than just how we get composes.

bcl:
Timestamps are currently the last time the STATUS file of a compose was updated. This could be changed, and/or we could also add a start time by looking at the timestamp from one of the other metadata files like final-kickstart.ks

jkozol:
@henrywang I added explicit actions to trigger loading blueprints and composes and modified the sagas a little bit. Now, only the import content sanity test fails. Hopefully I can fix that issue during the hackathon.

PR #309, opened 2018-07-19
Build srpm together with rpm

atodorov:
we don't really consume the src.rpm package for the moment but
at least build it so we can see if the build breaks.


henrywang:
@atodorov, the srpm will be copied and installed in docker by welder-web*.rpm. So bdcs test reported an error. Could you please take a look? Thanks.
```
Error: Will not install a source rpm package (welder-web-0.0.1-1.fc28.src).
```

henrywang:
Failed case is not caused by code changes. Merged.

PR #310, opened 2018-07-20
Download images

larskarlitski:
Since images can become quite large, we don't want to load them into memory with `cockpit.http()`. Instead, use cockpit's special `cockpit/channel/` url to create a url which is backed by a cockpit http-stream2 channel.

Blocked on
- [x] #304 (this branch includes @jkozol's commits, but I'll rebase once his PR lands)
- [x] figure out why file size is not sent in the headers

### DIFF
--- a/components/ListView/ListItemImages.js
+++ b/components/ListView/ListItemImages.js
@@ -62,11 +62,11 @@ class ListItemImages extends React.PureComponent {
                 <FormattedMessage defaultMessage="Failed" />
               </div>
             }
-            {listItem.queue_status === 'READY' &&
+            {listItem.queue_status === 'FINISHED' &&
               <div className="list-pf-actions">
-                <button className="btn btn-default" type="button">
+                <a className="btn btn-default" role="button" download href={this.props.downloadUrl}>
                   <FormattedMessage defaultMessage="Download" />
-                </button>
+                </a>
               </div>
             }
           </div>
--- a/core/sagas/composes.js
+++ b/core/sagas/composes.js
@@ -45,9 +45,9 @@ function* pollComposeStatus(compose) {
 
 function* fetchComposes() {
   try {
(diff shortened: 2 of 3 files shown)


clumens:
Whoops, excuse the message about me editing.  I was just checking whether those really were checkboxes, and then I changed it back.

larskarlitski:
Update: fix eslint errors

> Whoops, excuse the message about me editing. I was just checking whether those really were checkboxes, and then I changed it back.

No worries, there's no notification about those :)

bcl:
@larskarlitski Looks like Flask is including the length ok:

< HTTP/1.1 200 OK
< Content-Disposition: attachment; filename=c233d71c-6c26-4a4b-a5b6-9744e64ae281-disk.qcow2
< Content-Length: 1520435200
< Content-Type: application/octet-stream
< Last-Modified: Sat, 14 Jul 2018 00:55:45 GMT
< Cache-Control: public, max-age=43200
< Expires: Sat, 21 Jul 2018 06:20:59 GMT
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: HEAD, OPTIONS, GET
< Access-Control-Max-Age: 21600
< Date: Fri, 20 Jul 2018 18:20:59 GMT

So it's likely cockpit stripping that out.

larskarlitski:
> So it's likely cockpit stripping that out.

Yeah, I already suspected that. Thanks for finding out!

jgiardino:
I was able to successfully download a file.

larskarlitski:
Rebased.

jgiardino:
Tested again after rebase, and still works for me.

larskarlitski:
The tests seem to be flakes. I've extensively tested this manually and I'm seeing the same (mostly) tests fail on other PRs.

PR #311, opened 2018-07-24
Add Cockpit login support.

henrywang:
The end to end test will support cockpit login without authentication and cockpit login with username and password authentication.

### DIFF
--- a/test/end-to-end/utils/helper.js
+++ b/test/end-to-end/utils/helper.js
@@ -1,4 +1,5 @@
 const config = require('../wdio.conf.js');
+const CockpitLoginPage = require('../pages/cockpitLogin');
 
 module.exports = {
   // Switch to welder-web iframe if welder-web is integrated with Cockpit.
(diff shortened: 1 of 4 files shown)


test/end-to-end/utils/helper.js
```diff
@@ -7,8 +8,25 @@ module.exports = {
     if (config.config.baseUrl.includes('9090')) {
       browser
         .url(page.url)
-        .waitForExist(page.iframeSelector);
+        // wait for final page loaded not intermedia page
+        .pause(5000);
```
larskarlitski: Is there some way to wait for an element on the page, instead of having a long timeout here?

Timeouts are always tricky to get right. They can either make the tests run too long or turn into false negatives on resource-constrained test systems.

henrywang: @larskarlitski. I hate hard code in UI automation test as well. In this case, I have to wait here to make page in a final status.
The code has to cover two scenarios: cockpit auth required and no auth required.
Both scenarios go through login page. The workflow in no auth scenario will go to cockpit login page, then go to blueprints page without authentication.
I cannot find a way to distinguish two scenarios based on page element because they all go through the same page (cockpit auth page and blueprints page). So I have to wait for a while to let page go to final status. That means the page goes to cockpit login in cockpit auth required scenario and the page goes to blueprints page in no cockpit auth required scenario.

henrywang:
Close this PR and send a new PR instead.

PR #312, opened 2018-07-25
Add modal for listing sources

jgiardino:
This adds a kebab menu on the home page with an action "View Sources". When clicked, a modal displays that lists sources that are defined. For now, this information is read-only, but later we can add functionality for adding and removing custom sources.

### DIFF
--- a/components/Modal/ManageSources.js
+++ b/components/Modal/ManageSources.js
@@ -9,10 +9,13 @@ import EmptyState from '../../components/EmptyState/EmptyState';
 
 const messages = defineMessages({
   errorStateTitle: {
-    defaultMessage: "Translatable message"
+    defaultMessage: "An Error Occurred"
   },
   errorStateMessage: {
-    defaultMessage: ""
+    defaultMessage: "An error occurred while trying to get sources."
+  },
+  closeButtonLabel: {
+    defaultMessage: "Close"
   }
 });
 
@@ -47,6 +50,7 @@ class ManageSources extends React.Component {
                 type="button"
                 className="close"
                 data-dismiss="modal"
+                aria-label={formatMessage(messages.closeButtonLabel)}
               >
                 <span className="pficon pficon-close"></span>
               </button>
@@ -91,7 +95,7 @@ class ManageSources extends React.Component {
                 type="button"
                 className="btn btn-default"
                 data-dismiss="modal"
-              >Close</button>
+              ><FormattedMessage defaultMessage="Close" /></button>
             </div>
           </div>
         </div>
--- a/components/Modal/ManageSources.js
+++ b/components/Modal/ManageSources.js
@@ -3,9 +3,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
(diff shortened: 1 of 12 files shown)


components/Modal/ManageSources.js
```diff
@@ -0,0 +1,75 @@
...
+                <span className="pficon pficon-close"></span>
+              </button>
+              <h4 className="modal-title" id="myModalLabel">Sources</h4>
+            </div>
+            <div className="modal-body">
+              {this.props.sources.length === 0 &&
+                <EmptyState
+                  title="An Error Occurred"
+                  message="An error occurred while trying to get sources."
+                />
```
jgiardino: @larskarlitski @jkozol Can you review this and provide feedback? I'm not sure how to produce an error for testing in the UI, and am not sure what to base the display of the error message on.

andreasn:
Would it makes sense to have name, instead of repo address sorted first?
That's usually how it's sorted in the repo-files (for example in /etc/yum.repos.d/fedora.repo on fedora), and how it's [referred to from subscription-manager in RHEL](https://access.redhat.com/solutions/912213)

andreasn:
screenshot from 2018-07-26 15-11-38

Having **System Sources** crammed in within the columns looks a bit weird to me, maybe it could work as just a label above, and then repeat the column headers for the table for custom sources.

jgiardino:
@andreasn I added the class for the bootstrap modal that displays it at a larger size, but it only kicks in at a certain breakpoint. I'm not a fan of how it displays at the smaller breakpoints, but I also think we can come back and modify it more later. Some other ideas are to go back to a single table structure, but add css so that we still have the same visual styles that are there now (this will have the columns always align in both sections, and let them resize in width depending on the widest contents, which may help with the wrapping url). Or we add a custom style to have the larger modal width applied at some of the smaller breakpoints (but we'd still have the misalignment at the smallest size).

henrywang:
@jgiardino, could you please fix conflict and push again? Thanks.

```
Conflicting files
pages/blueprints/index.js
```

andreasn:
> I'm not a fan of how it displays at the smaller breakpoints, but I also think we can come back and modify it more later.

Do you mean the scrolling of the table that kicks in at below 835? Yeah, that doesn't look great, but yeah, we can do it as a followup fix.

components/Toolbar/BlueprintsToolbar.js
```diff
@@ -34,6 +34,26 @@ const BlueprintsToolbar = props => (
...
+            data-toggle="dropdown"
+            aria-haspopup="true"
+            aria-expanded="false"
+          >
+            <span className="fa fa-ellipsis-v" />
+          </button>
+          <ul className="dropdown-menu dropdown-menu-right" aria-labelledby="dropdownKebab">
+            <li><a href="#" onClick={(e) => props.handleShowModalManageSources(e)}>
+              <FormattedMessage
+                defaultMessage="View Source"
```
andreasn: Singular here, even though it's plural elsewhere. Should be View Sources to match with the dialog.

henrywang:
Tests passed. Merged.

PR #313, opened 2018-07-26
Fix PR#309 imported issue.

henrywang:
The rpm package should be welder-web*.noarch.rpm, not welder-web*.x86_64.rpm


henrywang:
This's a quick fix and passed all tests. It need to be merged to make the other PR test work.

PR #318, opened 2018-07-31
Automatic pushes to Zanata

dashea:
Upload the translatable strings to Zanata on pushes to master.

This PR is fun since nothing extra is supposed to happen, travis-wise, on the PR itself. Sure hope it works when it's merged!


dashea:
Huh, well that's disappointing. I have no clue why lorax-test would fail here, I didn't touch anything related to that.

PR #319, opened 2018-07-31
Don't call compose API on the blueprints page

larskarlitski:
We don't need composes there.


jgiardino:
Looks good!

larskarlitski:
Test failures are unrelated.

PR #325, opened 2018-08-02
Fixed bug where startComposeApi would not return start compose response

sfondell:
When creating an image the console would log a startComposeError and our start composer saga wasn't getting a response, I fixed this by adding a missing return statement


larskarlitski:
Test failures are unrelated.

PR #326, opened 2018-08-03
Fix blueprint packages getting added to history

jkozol:
Removed the updateBlueprint function since it provided no functionality not being handled by redux and was adding/removing components for a blueprint's entire history.

### DIFF
--- a/data/BlueprintApi.js
+++ b/data/BlueprintApi.js
@@ -105,7 +105,7 @@ class BlueprintApi {
     return components;
   }
 
-  // sets the ui type of a list of components 
+  // sets the ui type of a list of components
   setType(components, type) {
     components.map(component => {
       component.ui_type = type;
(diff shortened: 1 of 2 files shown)


jgiardino:
The updates relative to changes stored in `localPendingChanges` seems to behave as expected. But I'm noticing some issues with `workspacePendingChanges`. 

Try these steps, and you'll see a couple of weird behaviors:

1. Create a new blueprint
2. Add a few packages
3. Log out, then log back in.
    - The packages are still listed in the UI as expected, but they are also listed in the state for the `past` version of the blueprint, which is not expected. 
4. Hit Discard Changes
    - The packages are still listed in the UI, but shouldn't be there.
    - Additionally, the package description now displays with the name. This is the only time I'm seeing a description display in the list, but I think it should always be there. But if this is unrelated to this PR, we can create a separate issue for it.

jkozol:
I am seeing this behavior in master as well so I don't think it is related to this PR. I am going to look into that issue now if you don't have any other issues with this PR.

jgiardino:
Sure, that sounds good to me.

PR #328, opened 2018-08-03
Update Create Image notifications

jgiardino:
- Removes previous notifications. These weren't accurately reflecting the image creation process.
- Adds notification text for each status type
- Only displays notification for status 'imageWaiting'. The other notifications will be added in a later PR.

### DIFF
--- a/data/NotificationsApi.js
+++ b/data/NotificationsApi.js
@@ -32,32 +32,39 @@ class NotificationsApi {
   notificationMessage(blueprint, message) {
     let notification = {};
     switch (message) {
-      case 'creating': {
+      case 'imageWaiting': {
         notification = {
-          type: 'process',
+          type: 'info',
           message: <span>
             <FormattedMessage
-              defaultMessage="{blueprint} Creating image."
+              defaultMessage="{blueprint} Image creation has been added to the queue."
               values={{
                 blueprint: <strong>{blueprint}:</strong>
               }}
             />
           </span>,
-          action: <a href="#"><FormattedMessage defaultMessage="Cancel" /></a>,
           dismiss: true,
+          fade: true,
         };
-        const index = this.notifications.length;
-        setTimeout(() => {
-          this.closeNotification(index);
-          this.displayNotification(blueprint, 'created');
-        }, 2500);
-        // setTimeout is only temporary, and included to simulate what will happen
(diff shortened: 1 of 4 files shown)


data/NotificationsApi.js
```diff
@@ -32,32 +32,39 @@ class NotificationsApi {
...
+        };
+        break;
+      }
+      case 'imageStarted': {
+        notification = {
+          type: 'success',
           message: <span>
             <FormattedMessage
-              defaultMessage="{blueprint} Creating image."
+              defaultMessage="{blueprint} Image creation has started."
```
henrywang: One failed case is due to this change. Just update the 22nd line of `pages/toastNotif.js`.
The ```this.varStatusCreating = `${this.varBlueprintName} Creating image.`;``` should be updated to ```this.varStatusCreating = `${this.varBlueprintName} Image creation has started.`;```

henrywang:
@jgiardino, the notification workflow changed, but just got notification with  "Image creation has been added to the queue." message. No more notification found in my local environment. I think there should have notification when image creation failed or success.

It has to update two files to make test work again.
1. spec/test_viewBlueprint.js
Please remove lines from 155-159, because there's no second status in notification workflow.
```
            browser
              .waitForVisible(toastNotifPage.iconComplete);

            const textComplete = $(toastNotifPage.labelStatus).getText();
            assert.equal(textComplete, toastNotifPage.varStatusComplete);
```
2. pages/toastNotif.js
`Line 13`: update to ```this.iconCreating = '#cmpsr-toast-0 .pficon-info';```
`Line 22`: update to ```this.varStatusCreating = `${this.varBlueprintName} Image creation has been added to the queue.`;```

jgiardino:
Thanks for the suggestions, @henrywang! I made these updates and the tests passed locally.

jgiardino:
I forgot to comment on this:

> I think there should have notification when image creation failed or success.

Yes, I agree that we should display notifications for each status change in the image build process. For this PR, I just wanted to remove the notifications that weren't accurately reflecting what's happening with the build process. Also, I can easily display a notification for when a build is added to the queue, but the other status changes will require more work and will be handled in a separate PR.

henrywang:
@jgiardino, thanks for your explanation, it makes sense to me.

PR #329, opened 2018-08-04
Adjust Images list layout to improve alignment

jgiardino:
- Modifies css to accommodate extra `<span>` added for translated strings so that date and type values will wrap on wider screens.
- Adds `div.list-pf-actions` for list items that don't have actions since this div is required for flex layout and alignment across all the list items.

**Screencaps of expected layout
window width < 992px**
image
**window width > 992px**
image


jkozol:
This looks good to me.

Issue #280, opened 2018-07-05
Image list display states and available actions


jgiardino:
The Blueprint page > Images tab needs refinements related to displaying proper states for the UI and enabling/disabling actions based on contents in the tab. 

- While fetching images for a blueprint, a loading message should display
- After fetch is complete, if an error occurred, then an error message should display (similar to the one added for Blueprints in #268)
- ~~If an empty state message displays due to no images, the Create Image actions should only be enabled when the blueprint has contents.~~ This flow has been updated, and addressed in #330. See comments below.

jgiardino:
The original issue description doesn't reflect the work that's needed.

As part of #304, images are fetched immediately. This would eliminate the need for displaying a loading state. I ultimately think the workflow related to fetching data could be refactored, but I don't consider that critical.

As part of a chat with @bcl, the user _could_ compose an image if the blueprint is empty. Some image types, like live-iso, would result in a minimal bootable image. But some image types might be totally empty. So instead of prevent the user from creating an image, we should let them, but display a warning in the dialog that the blueprint is empty.

jgiardino:
Closing this issue. Any remaining work is captured as part of #323.

PR #330, opened 2018-08-07
Display warning messages to the user in Create Image modal

jgiardino:
These updates will display a warning message in the modal dialog to the user for the following cases:

- The blueprint is empty
- The blueprint has unsaved changes
  - This case required a small change to the data returned by fetchBlueprintInfoApi to also include the `changed` value. On the blueprints page, this is the only way we know if there are unsaved changes in the workspace, without fetching the blueprint contents.

image

_Deleted the example of the second case. An updated example is included [below](https://github.com/weldr/welder-web/pull/330#issuecomment-416083737)_

### DIFF
--- a/components/ListView/BlueprintListView.js
+++ b/components/ListView/BlueprintListView.js
@@ -20,19 +20,13 @@ class BlueprintListView extends React.PureComponent {
               <Link to={`/edit/${blueprint.name}`} className="btn btn-default">
                 <FormattedMessage defaultMessage="Edit Blueprint" />
               </Link>
-              {(blueprint.modules.length === 0 && blueprint.packages.length === 0) &&
-                <button type="button" className="btn btn-default" disabled="disabled">
-                  <FormattedMessage defaultMessage="Create Image" />
-                </button>
-              ||
-                <button
-                  type="button"
-                  className="btn btn-default"
-                  onClick={(e) => this.props.handleShowModalCreateImage(e, blueprint)}
-                >
-                  <FormattedMessage defaultMessage="Create Image" />
-                </button>
-              }
+              <button
+                type="button"
+                className="btn btn-default"
+                onClick={(e) => this.props.handleShowModalCreateImage(e, blueprint)}
+              >
+                <FormattedMessage defaultMessage="Create Image" />
+              </button>
               <div className="dropdown pull-right dropdown-kebab-pf">
                 <button
                   className="btn btn-link dropdown-toggle"
--- a/components/Modal/CreateBlueprint.js
+++ b/components/Modal/CreateBlueprint.js
@@ -221,7 +221,7 @@ class CreateBlueprint extends React.Component {
                   type="button"
                   className="btn btn-primary"
                   onClick={(e) => this.showInlineError(e)}
-                >Create</button>
+                ><FormattedMessage defaultMessage="Create" /></button>
                 ||
                 <button
                   type="button"
--- a/components/Modal/CreateImage.js
+++ b/components/Modal/CreateImage.js
@@ -4,12 +4,17 @@ import React from 'react';
 import {FormattedMessage, defineMessages, injectIntl, intlShape} from 'react-intl';
 import PropTypes from 'prop-types';
 import NotificationsApi from '../../data/NotificationsApi';
-import { Alert } from 'patternfly-react';
 import BlueprintApi from '../../data/BlueprintApi';
+import { Button, OverlayTrigger, Popover, Icon, Alert } from 'patternfly-react';
 import { connect } from 'react-redux';
 import { setBlueprint } from '../../core/actions/blueprints';
+import { fetchingQueue, clearQueue } from '../../core/actions/composes';
 
 const messages = defineMessages({
+  infotip: {
+    defaultMessage: "This process can take a while. " +
+      "Images are built in the order they are started."
+  },
   warningUnsaved: {
     defaultMessage: "This blueprint has changes that are not committed. " +
       "These changes will be committed before the image is created."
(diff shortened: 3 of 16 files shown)


jgiardino:
@henrywang There are some tests that are failing related to translated strings, for example:

```
CreateImage › props test › setNofifications, passed by props, 
should be called by clicking Create button

Invariant Violation: [React Intl] Could not find required `intl` 
object. <IntlProvider> needs to exist in the component ancestry.
```

Do you have any recommendations for how to fix this? I've reviewed the [blog post about translating strings](http://weldr.io/Translating-welder-web-strings/), but can't figure out why it's failing.

henrywang:
@jgiardino, according to [stackoverflow solution](https://stackoverflow.com/questions/37021217/injecting-react-intl-object-into-mounted-enzyme-components-for-testing), I did some update to make unit test work again.
1. added a helper.js file into test/unit folder. Here's content of helper.js
```javascript
/**
 * Components using the react-intl module require access to the intl context.
 * This is not available when mounting single components in Enzyme.
 * These helper functions aim to address that and wrap a valid,
 * English-locale intl context around them.
 */

import React from 'react';
import { IntlProvider, intlShape } from 'react-intl';
import { mount, shallow } from 'enzyme';

// Create the IntlProvider to retrieve context for wrapping around.
const intlProvider = new IntlProvider({ locale: 'en' }, {});
const { intl } = intlProvider.getChildContext();

/**
 * When using React-Intl `injectIntl` on components, props.intl is required.
 */
function nodeWithIntlProp(node) {
  return React.cloneElement(node, { intl });
}

/**
 * Export these methods.
 */
export function shallowWithIntl(node) {
  return shallow(nodeWithIntlProp(node), { context: { intl } });
}

export function mountWithIntl(node) {
  return mount(nodeWithIntlProp(node), {
    context: { intl },
    childContextTypes: { intl: intlShape }
  });
}
```
2. Some changes on file `create.image.spec.js`
`Line 2` should be updated to ```import { shallowWithIntl } from './helper';```
`Line 35` should be updated to ```   return shallowWithIntl(```
`Line 37` should be updated to ```    ).first().shallow();```

jgiardino:
Thanks for the suggestions, @henrywang! I made those updates and the tests pass.

henrywang:
@jgiardino, could you please update pages/viewBlueprint.js to fix end to end failed test? Thanks.
`Line 43` updated to ```this.btnCreateImage = '#cmpsr-btn-crt-image';```

jgiardino:
The latest commits include significant changes to this dialog. 

There are also failing tests which I'm not sure how to fix currently, and will continue working on those.

- In the case where there are unsaved changes: 
  - the Create button in the dialog changes to Commit and Create
  - A warning message displays stating that the unsaved changes will be committed.
  - When the user hits the Commit and Create, the same notifications for saving the blueprint display before the build process starts.
- Information about queue status has also been added to the dialog, as shown in the example below. Queue status shown is for the entire app, and not specific to any blueprint.

image

And here is an example of the updated button label and message for when there are unsaved changes:
image

jgiardino:
@henrywang I get the following error when I run the tests locally. I'll try to investigate how to fix this, but please let me know if you have any suggestions.

Invariant Violation: Could not find "store" in either the context or props of "Connect(InjectIntl(CreateImage))". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(InjectIntl(CreateImage))".

jgiardino:
@henrywang I took a further look at the issues flagged in unit/create.image.spec.js, and it isn't obvious to me how to update the test to handle the later changes I made in the Create Image modal. Please let me know if you have any suggestions on how to fix. Feel free to push commits to this PR, if that's easier for you. Thank you!!

henrywang:
@jgiardino, I mocked store with the following code:
```javascript
import configureStore from 'redux-mock-store'
const mockStore = configureStore(middlewares)
const initialState = {}
const store = mockStore(initialState)
```
Then include apply this store to shallow method:
```javascript
<CreateImage {...Object.assign({ imageTypes: typeList }, props)} store={store}/>
```
The "Invariant Violation" was gone, but there's a new error in this case:
```
Invariant Violation: [React Intl] Could not find required ``intl`` object. <IntlProvider> needs to exist in the component ancestry.
```
Cannot resolve this new error so far.

jgiardino:
Thanks for looking at this, @henrywang!

For the new error, the other unit test index.norecipes.spec.js uses `<IntlProvider locale='en'>`. But there are quite a few differences between these two tests, so I'm not sure how to integrate that into the create.image.spec.js tests.

andreasn:
Uncommited changes works great, thanks for fixing that!

Previously the Create Image button was disabled if the blueprint was empty.
What's the motivation for allowing to create an empty image (with a warning)?

jgiardino:
Thanks for reviewing, @andreasn!

> What's the motivation for allowing to create an empty image (with a warning)?

This is based on feedback from @bcl. I pasted below some notes I included in issue #280 after chatting with him about this:

> As part of a chat with @bcl, the user could compose an image if the blueprint is empty. Some image types, like live-iso, would result in a minimal bootable image. But some image types might be totally empty. So instead of prevent the user from creating an image, we should let them, but display a warning in the dialog that the blueprint is empty.

larskarlitski:
Fixed up @jgiardino last commit to pass eslint and merged it into the commit that introduced the breaking changes.

Thanks @jgiardino for this!

andreasn:
> This is based on feedback from @bcl. I pasted below some notes I included in issue #280 after chatting with him about this:

Ah, ok. Thanks for the explanation!

PR #331, opened 2018-08-08
Change when loading states stops showing on Blueprints page

### LINKED ISSUE #305: Blueprints list loading state
### Description of problem
Currently, the app state gets updated with `fetchingBlueprints: false` when the action FETCHING_BLUEPRINT_NAMES_SUCCEEDED is called. However, the contents needed to populate the page aren't actually added at this time. Now that the ability to pull the list of images is included in PR #304, it's much more obvious that this is the case.  

Instead, the state should be updated when the image contents are successfully fetched (e.g. FETCHING_BLUEPRINTS_SUCCEEDED).

jgiardino:
Modifies actions, reducers and saga for the blueprints page so that the loading state remains on the screen until `FETCHING_BLUEPRINTS_SUCCEEDED` is called.

Closes #305

### DIFF
--- a/core/actions/blueprints.js
+++ b/core/actions/blueprints.js
@@ -20,11 +20,6 @@ export const fetchingBlueprints = () => ({
   type: FETCHING_BLUEPRINTS,
 });
 
-export const FETCHING_BLUEPRINT_NAMES_SUCCEEDED = 'FETCHING_BLUEPRINT_NAMES_SUCCEEDED';
-export const fetchingBlueprintNamesSucceeded = () => ({
-  type: FETCHING_BLUEPRINT_NAMES_SUCCEEDED,
-});
-
 export const FETCHING_BLUEPRINTS_SUCCEEDED = 'FETCHING_BLUEPRINTS_SUCCEEDED';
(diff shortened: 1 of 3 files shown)


henrywang:
@jgiardino, the blueprints page keeps on loading status, so it caused failed cases. Could you please take a look at this issue if you have time? Thanks!

jgiardino:
I think the issue that we're seeing in the tests has been a recurring issue, and is not specific to this PR. I'm not sure how to move forward with the issues we're having with tests, since my local set up is not similar to the way the tests are set up, and I'm not able to reproduce the issues that are encountered in the tests. 

The updates in this PR don't affect how blueprints are fetched. They just change how long the loading state is displayed. Before this change, the loading state disappeared before the blueprints were fetched. With this change, the loading state is removed when we have blueprints to display on the page.

KKoukiou:
@jgiardino Is thi sstill relevant? If no, let's cose it to clean up the open PR list.

jgiardino:
This is out of date, and other changes related to this have been merged since this PR was opened

PR #334, opened 2018-08-10
Add the .spec files to .PHONY

bcl:
So that they will always get updated with the correct version-release
even if they already exist.


henrywang:
@bcl, the test broken is caused by Travis itself. I just re-run all tests, It recovered.
The failed tests are not caused by your code change. So feel free to merge this PR.
BTW: I'm working on another PR to make some of tests more stable.

PR #335, opened 2018-08-14
WIP: Integrate with Cockpit CI workflow.

henrywang:
- [x] Update end to end test case to support cockpit login.
- [x] Optimize end to end test case to make test faster and stable.
- [x] Run end to end test on Openstack VM.
- [ ] Move end to end test CI platform to Jenkins.
- [x] Keep eslint, css style, unit test, build welder-web rpm and build cockpit-composer rpm in Travis.
- [ ] Generate code coverage result and upload to codecov.
- [ ] Generate test report and upload to fedorapeople.org.


martinpitt:
@henrywang : As per our recent discussion, here are some pointers to how Cockpit tests use selenium:

 * The selenium backends run in a separate VM. We have a "selenium" image (`bots/image-download selenium`) which has the selenium drivers for Firefox and Chromium, a "windows-10" image for Edge and its Selenium driver, and a "windows-8" image for IE (but don't bother about that, it's slated for removal).

 * The selenium test frontend is running in the test VM, which has cockpit installed (currently our fedora-27 and fedora-28 images both have that). We use Fedora's `python2-selenium` package for that, plus `python2-avocado` (but I *think* this is just because our tests happen to be written in Avocado style, it's not a requirement). If you use something else, like a JavaScript test driver for selenium, you can add this to your VM preparation script for now. Once it works, I'm happy to add it to the VM image, to save the installation time at test running time.

 * These two VMs are started and controlled by the top-level [run-tests](https://github.com/cockpit-project/cockpit/blob/master/test/avocado/run-tests) script. In particular, [run_avocado()](https://github.com/cockpit-project/cockpit/blob/master/test/avocado/run-tests#L163) is the interesting bit here, which launches the two VMs, waits for selenium to start, and then runs the actual test script. 

   This is the bit which is rather dependent on our [Python test VM API](https://github.com/cockpit-project/cockpit/blob/master/bots/machine/testvm.py). You can [start VMs from JavaScript or other languages](https://cockpit-project.org/blog/cockpit-custom-test-framework.html), but the CLI does not currently allow passing options for networking, so this might require some additions to testvm.py. I. e. if you just launch two VMs independently, they won't be able to talk to each other, this is where the `networking=network.host()` bits come in. I'm happy to look at this part, i. e. being able to launch two VMs with testvm.py's CLI that can talk to each other.

 * The actual [test scripts](https://github.com/cockpit-project/cockpit/blob/master/test/avocado/selenium-base.py) are fairly straightforward. They just get copied to the test VM and launched there (from run-tests) once everything is set up. I. e. these should just be the existing tests.

henrywang:
Close this PR because another PR will replace this one.

PR #341, opened 2018-08-20
Make strings translatable in Pending Changes dialog

jgiardino:
This PR updates the Pending Changes dialog so that the strings can be translated.


henrywang:
It works for me. Failed test is not related with this PR.

PR #342, opened 2018-08-21
Makefile: don't run po-pull on dist

larskarlitski:
Updating translations and releasing a tarball are independent actions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ install: all
 	cp -r public/* /usr/share/cockpit/welder
 
 dist-gzip: NODE_ENV=production
-dist-gzip: po-pull all
+dist-gzip: all
 	mkdir -p welder-web-$(VERSION)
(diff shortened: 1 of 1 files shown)


martinpitt:
@bcl: This should take care of the cockpituous release failure.

larskarlitski:
Test failure is unrelated to this Makefile change. Merging.

bcl:
How do we include the latest translations if po-pull isn't going to be run on release? Is that something that cockpituous will do as a separate step?

martinpitt:
In cockpit we have a bot that regularly (every week or so) creates PRs with the result of `make po-pull`. This sometimes causes test regressions because of invalid translations, so it really should be validated by CI (of course you need a test that actually checks a non-English language or two, but at *least* make sure that the build system can process the po files) and also (coarsely) by a human. Doing it automatically at release time is a lingering trap.

PR #343, opened 2018-08-21
Clean up package.json

larskarlitski:
GitHub has been showing a helpful "We found potential security vulnerabilities in your dependencies." alert for a while now, because dependencies were out of date. Most importantly, we were still pulling in jquery 2.

Clean this up a bit by dropping some unused dependencies and updating many of the others to recent versions.

I've been using `npm audit` to figure out which dependencies have know vulnerabilities. `audit` only exists in npm 6, which writes a different versions into `package-lock.json` than earlier npms. It's been a pain to revert those changes all the time as I was switching npm versions. I've included a commit that removes it. @dashea, you added package-lock.json in the first place. Are you ok with removing it for now?

@jgiardino, did I drop anything we're actually using? I'm especially unsure about the patternfly and web component packages.


martinpitt:
FTR, I strongly believe package-lock.json should never be committed to git. package.json should have appropriate dependency constraints (usually using something like `^1.2.3` to specify the minimal required version and then use the maximum one that is compatible to that). Otherwise you keep falling behind npm security and bug fix updates.

OTOH package-lock.json makes sense to get shipped in released tarballs (`make dist`), so that node_modules/ can be reconstructed later for patching.

martinpitt:
This conceptually LGTM, but it seems to break the tests. So there's some refinement necessary?

dashea:
I do not think package-lock.json should be removed. The recommendation from npm is pretty much the opposite of what @martinpitt suggests: it *should* be committed to source control, so that development and continuous integration is using a stable, reproducible base, and it *should not* be published, so that downstream consumers can set their own exact versions. Fixing security bugs shouldn't be a surprise, if package-lock.json needs newer versions, update them and commit it back to the repo.

martinpitt:
>  so that downstream consumers can set their own exact versions

Sorry, but this is absurd. Changing to new npm modules can introduce any amount of breakage, so such a downstream change would have to be vetted again with the full test suite run, code and build system changes to adjust, etc. This is clearly in the upstream development domain, not in the downstream packaging area.

If you don't publish package-lock.json, then you have to publish and ship the entire node_modules/ in your release tarball. This is also quite impractical, as then you suddenly ship tons of stuff which isn't relevant at all for runtime, but which you have to get through license review and all that.

Well, we can agree to disagree of course, but this npm recommendation is just completely at odds with best upstream development and downstream maintenance practices.

larskarlitski:
@dashea If we keep it for development, how could we solve the issue that different versions of npm produce different `package-lock.json` contents?

dashea:
@larskarlitski if you generate the file at different times it's going to be different, because it'll try to pin the versions to whatever is the newest at that point in time. if different versions of npm are doing that differently, that's a problem with npm, and that would actually be something package-lock.json would help with since every developer would be developing against the same set of libraries regardless of what npm would otherwise want to do.

larskarlitski:
That's my point: `npm install` changes the contents of the file all the time, which is annoying when developing. As you were saying, updates shouldn't come as a surprise, but have separate commits.

The way it is now, when I `git clone` the repository and run `npm install` and immediately have changes in `package-lock.json`. Should I always revert those?

dashea:
Depends on the change, I guess.

martinpitt:
> Fixing security bugs shouldn't be a surprise, if package-lock.json needs newer versions, update them and commit it back to the repo.

Fair enough. This is actually not a bad approach for `dependencies`, as long as you actually *do* it. But you don't need the unreliable and illegible package-lock.json for that, the tool to express that is package.json. I. e. if you don't want automatic microrelease updates, just pin down the version and update them manually.

 For `devDependencies` I have a different opinion, and it has proven in practice that using `^1.2.3` style constraints works quite well. Sure, occasionally people break stuff in newer versions without marking them as incompatible, but then you can either adjust and follow the herd, or make the version constraint in package.json harder if you explicitly want to opt out of new development tools.

So in summary, I strongly think that package.json is the place to control that behaviour, not package-lock.json.

larskarlitski:
> Depends on the change, I guess.

I'm sorry, I don't understand. For example, [this is the patch](https://gist.github.com/larskarlitski/080d5d102a4251af03da974ef9f04b3d) I get on current master when running with npm 6.2.0.

I like the concept of package-lock.json (for dependencies, as Martin was saying), but the way npm is handling it right now is broken. Yes, they recommend committing it, but the web is full of people having problems with the current implementation.

The only other solution I see is requiring people and CI to use a specific version of npm.

larskarlitski:
Thanks Martin. I've dropped the commit removing web components (@jgiardino's comment was about those) and did it properly in #346. That PR also (attempts) to fix the test failures.

The changes in here are now removing package-lock.json and updating dependencies. None of those should affect the tests and manual testing shows everything to work.

martinpitt:
Now even the unit tests fail:

   react-dom is an implicit dependency in order to support react@0.13-14. Please add the appropriate version to your devDependencies. See https://github.com/airbnb/enzyme#installation

These did work before, so I don't dare merging it like that.

larskarlitski:
> Now even the unit tests fail:

Sorry about that. Enzyme, which provides unit testing helpers for React apps, probably wasn't compatible with the newest version of React that the `~` change brought in. I've updated both enzyme and react to the newest minor versions and added an `Adapter`, which those versions need now.

We would have had to do this at some point anyway.

PR #344, opened 2018-08-22
Makes strings translatable in pagination for available components

jgiardino:
Updates make strings translatable but also available to screen readers.

### DIFF
--- a/components/Pagination/Pagination.js
+++ b/components/Pagination/Pagination.js
@@ -167,7 +167,7 @@ class Pagination extends React.Component {
           />:
         </span>
         <FormattedMessage
-          defaultMessage="{startNumber}–{endNumber} of {totalNumber}"
+          defaultMessage="{startNumber} - {endNumber} of {totalNumber}"
           values={{
             startNumber: startItems,
             endNumber: endItems,
--- a/components/Pagination/Pagination.js
+++ b/components/Pagination/Pagination.js
@@ -10,7 +10,11 @@ const messages = defineMessages({
     defaultMessage: "Show Next Page"
   },
   currentPage: {
-    defaultMessage: "Current Page"
+    defaultMessage: "Current Page",
+    description: "Label of form input for setting current page number"
+  },
+  pagination: {
+    defaultMessage: "Pagination"
   }
 });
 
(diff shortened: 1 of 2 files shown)


components/Pagination/Pagination.js
```diff
@@ -123,25 +141,36 @@ class Pagination extends React.Component {
...
-            {(currentPage + 1) * pageSize - pageSize + 1}
-            <span> - </span>
-            {currentPage === totalPages && totalItems || (currentPage + 1) * pageSize}
-          </span> of {totalItems}
+        <FormattedMessage
+          defaultMessage="{items} {start} {to} {end} of {total} {items}"
+          values={{
+            items: <span className="sr-only"><FormattedMessage defaultMessage="Items" /></span>,
+            start: startItems,
+            to: <span className="pagination-cmpsr-to"><FormattedMessage defaultMessage="to" /></span>,
```
martinpitt: This is the first time I see `<FormattedMessage>` and reactl-intl (what a pain, compared to the usual gettext!), so excuse me thinking aloud: Why does "to" get translated separately, but "of" doesn't? I thought the outer `"{items} {start} {to} {end} of {total} {items}"` is the string that would actually end up being translated, so both the words themselves and their order can change.

I figure the difference is that the "to" gets wrapped into this extra `pagination-cmpsr-to` span, and you want to avoid putting that part into the translatable string. This makes sense, and the actual "to" word can still be translated (although it might be a little confusing to do that without seeing the context).

The same situation arises with normal gettext, and there's usually no better solution than "avoid formatting within one sentence".

So bottom line, this looks okay to me.

components/Pagination/Pagination.js
```diff
@@ -1,6 +1,19 @@
...
+  previousPage: {
+    defaultMessage: "Show Previous Page"
+  },
+  nextPage: {
+    defaultMessage: "Show Next Page"
+  },
+  currentPage: {
+    defaultMessage: "Current Page"
+  }
+});
```
martinpitt: Has it been considered to just use icons for that? Like arrows?

components/Pagination/Pagination.js
jgiardino: I had pushed an updated version of this yesterday. Originally with this implementation, I was hoping to align with the PF design by using a dash between start and end values, like "1-50", but provide a friendlier version to screen reader users by using "to", as in "one to fifty". But there's no clean way to provide this in a way that sends the string to translators as "{startNumber} to {endNumber} of {totalNumber}". 

In my commit yesterday, I simplified the strings so that translators get "{startNumber} - {endNumber} of {totalNumber}", and screen readers get something like "(Current Items:) one dash fifty of one hundred". I still don't like the dash (-)—I'm not sure how well a dash works for either translation or screen reader users—but I also would like to be consistent with PF designs.

I'm still on the fence for this one, so I could easily be convinced one way or the other :-)

components/Pagination/Pagination.js
jgiardino: This is the label for the input field, which is only visible by assistive technology like screen readers.

martinpitt: I see, thanks! So, fine from my POV to land, if you are confident that CI didn't break any further.

larskarlitski:
> So, fine from my POV to land, if you are confident that CI didn't break any further.

Please hold off landing, I'm almost done with fixing tests.

larskarlitski:
> @larskarlitski : Should we still hold off, or land this now?

I've gotten the tests to a better state, but there are still some flakes (in particular in `CreateNewBluePrint`, which sadly is called from many others).

This PR looks good to me and the tests suite passes locally some of the time, just like for master. It's unlikely that this is introducing new failures.

PR #345, opened 2018-08-23
[WIP] Refactor API Calls

jkozol:
I refactored the createBlueprint API call so it gets called from a saga. I refactored the fetchBlueprintContents API call so it doesn't call as many other functions and is more readable.


jgiardino:
Closing this one. These updates were pulled into #500

PR #346, opened 2018-08-28
Use upstream patternfly-react's Tab component

larskarlitski:
This lets us drop `components/Tabs` and all webcomponents dependencies.

Tab transitions are animated by default. Should we turn those off, as we didn't have animation before?

Note: this doesn't include patternfly-react's css yet, because doing so requires reworking how we include css. Seems to be working fine with only patternfly.css for now, so I think this can be handled in a follow-up.


components/ListView/ComponentDetailsView.js
```diff
@@ -339,14 +328,12 @@ class ComponentDetailsView extends React.Component {
...
+              <Tab eventKey="components" title="Components">
                 <p><FormattedMessage defaultMessage="Components" /></p>
               </Tab>}
             <Tab
-              tabTitle={formatMessage(messages.dependencies,
-                {count: `<span class="badge">${this.state.dependencies.length}</span>`}
-              )}
-              active={this.state.activeTab === 'Dependencies'}
+              eventKey="dependencies"
+              title={<LabelWithBadge title={formatMessage(messages.dependencies)} bagde={this.state.dependencies.length} />}
```
jgiardino: Oh, maybe this typo is causing the dependencies count to not display: `bagde`?

jgiardino:
There are a couple more references to webcomponents, so just double-checking if these are still needed:

- in webpack.config.js: 
  ```
    // for console log error related to webcomponents
    // https://github.com/webcomponents/webcomponentsjs/issues/794
    new webpack.IgnorePlugin(/vertx/),
  ```
- in index.ejs:
`<script src="dist/webcomponents-lite.js"></script>`

components/ListView/ComponentDetailsView.js
larskarlitski: Oops, indeed. Thank you!

larskarlitski:
Rebased and fixed the type. Thanks for having a look, @jgiardino!

> There are a couple more references to webcomponents, so just double-checking if these are still needed:

Thanks. I've removed those as well.

martinpitt:
Looks plausible to me, thanks for the nice cleanup! Retrying tests.

jgiardino:
looks good from my end!

jgiardino:
Also wanted to note that this addresses all the issues noted in #247. 
And that I created #349 to capture the remaining updates related to the pf-styled tabs.

martinpitt:
Restarted the tests a third time, and they apparently keep failing. Before they were green, and e. g. Jenn's PR #347 is also green, so this raises a flag.

larskarlitski:
> Restarted the tests a third time, and they apparently keep failing. Before they were green, and e. g. Jenn's PR #347 is also green, so this raises a flag.

I see your point and I don't like having to argue against it, but these tests are super flaky. I see a completely different set of tests failing when I run them locally.

I don't see were they were green before. Master is failing, as was my PR that updated versions, as are most currently open PRs.

Sorry :/

larskarlitski:
Case in point: #347 failed after @jgiardino pushed minor fixes.

martinpitt:
@larskarlitski OK, I see. If they succeed for you locally, or more generally, if you are confident that this doesn't regress, I'm happy to merge.

PR #351, opened 2018-09-03
Keep translations upstream

larskarlitski:
Modify the build so that `.po` files live upstream in `po/`. Also pull the current translations.

We probably want to make `npm run translations:po2json` part of the build at some point, but I hesitate making the build even slower right now. Moving to a Makefile which would only generate `build/translations.json` when `.po` files have changed would make this feasible.


run.js
```diff
@@ -54,7 +54,7 @@ tasks.set('locale-data', () => {
     var translationsData = JSON.parse(fs.readFileSync(translationsPath));
 
     // Create a module that imports and loads each language's locale data
-    var languages = Object.keys(translationsData);
+    var languages = Object.keys(translationsData).map(lang => lang.split('_')[0]);
```
martinpitt: This does work for for most languages, except for en, es, zh, and pt, where the region code does make a difference. pt_BR is quite different from pt_PT, and zh_TW/zh_CN are *completely* different (traditional vs. simplified). Could it be that it perhaps just needs to be named like "zh-cn" or "zh-CN"?

larskarlitski: Right, but react-intl doesn't ship those in `locale-data/`, which is where the code below is trying to pull them from. I wonder why it needs to do this explicitly at all...

martinpitt: OK, I now better understand what this does -- this is just a glorified list of `import` statements for the groups of languages/locales that composer wants to support. So *this* looks ok.

The whole thing still does not work at all, though - if you have po/pt_BR.po, po/zh_CN.po etc, and set these languages in the browser, they are being completely ignored, and Composer appears in English. It only works for "pt" and "zh", but these are *not* proper languages.

And of course a thing like "en" or "es" is *never* a locale, so formatting dates, times, currencies and the like will not work with just these.

larskarlitski:
Pushed an update with @martinpitt's suggestion (thanks!). Also fixed README.md and added a warning when running `translations:pull` without zanata-js installed.

PR #353, opened 2018-09-05
translations: Use ~/.config/zanata.ini

larskarlitski:
This is how upstream zanata documents using the official client. zanata-js
reads this file as well. Zanata even offers UI with the contents of zanata.ini
filled with the server and API key to copy and paste.

We can get rid of our zanata-js wrapper now.


dashea:
This will break the po-push target run from Travis. The values in ~/.config/zanata.ini should already be being used if no values are provided from the environment.

larskarlitski:
> This will break the po-push target run from Travis. The values in ~/.config/zanata.ini should already be being used if no values are provided from the environment.

Ah, indeed. Thanks for pointing this out.

Sorry for the noise.

PR #355, opened 2018-09-10
Bug fix and case update:

henrywang:
1. Fix failed cases in test_viewBlueprint and test_pendingCommit. The CSS changes is the root cause of case failed.
2. Update some array loop from forEach to some because forEach loop all of elements in array but cannot break.
3. Update case to go to editBlurprint page by clicking Edit Blueprint button instead of using url.
4. Add package-lock.json to make test happy.


henrywang:
The lorax-composer updated from x86_64 28.15-1.fc28 to x86_64 28.18-1.fc28. The welder-web cannot work with lorax-composer x86_64 28.18-1.fc28. That caused lorax-test failed.
All the tests are passed on bdcs-test scenario.

larskarlitski:
Regarding your last commit, please change that to something like this:

```diff
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,12 +1,7 @@
 #!/bin/bash
 
-# Make sure the folder exists
-mkdir -p /run/weldr
-# Set permissions
-chmod u=rwX,g=rwX,o=--- /run/weldr
-
 # Run lorax-composer
-lorax-composer --group root /blueprints &
+lorax-composer /blueprints &
 
 # Execute the CMD
 exec "$@"
```

lorax-composer creates `/run/weldr` by itself, there's no need to create it anymore. Also, not passing `--group root` brings us closer to what's used in production and gets rid of the error on startup.

henrywang:
@larskarlitski, If we let lorax-composer create /run/weldr folder, this folder will have root:root ownership. That'll not work inside docker container. So I have to add the following settings back.
```
# Make sure the folder exists
mkdir -p /run/weldr
# Set permissions and owership
chmod u=rwX,g=rX,o=--- /run/weldr
chown root:weldr /run/weldr
```

entrypoint.sh
```diff
@@ -2,11 +2,12 @@
 
 # Make sure the folder exists
 mkdir -p /run/weldr
-# Set permissions
-chmod u=rwX,g=rwX,o=--- /run/weldr
+# Set permissions and owership
+chmod u=rwX,g=rX,o=--- /run/weldr
+chown root:weldr /run/weldr
```
larskarlitski: As discussed on IRC, I'd like to know why we need this in case we need to file a bug against lorax-composer.

In my local testing, lorax-composer always creates this directory with the correct permissions.

test/end-to-end/specs/test_pendingCommit.js
```diff
@@ -126,8 +132,18 @@ describe('Changes Pending Commit Page', () => {
...
+        const rowSelector = `${blueprintsPage.itemsBlueprint}[data-blueprint="${blueprintName}"]`;
+        helper.goto(blueprintsPage);
+
+        browser
+          .waitForVisible(rowSelector);
+
+        browser
+          .click(`${rowSelector} a[href*="edit"]`)
+          .waitForExist(editBlueprintPage.componentListItemRootElement);
+          // .waitForVisible(editBlueprintPage.btnCommit);
```
larskarlitski: Please remove this comment.

henrywang:
@larskarlitski, Thanks for your review. Code updated.

larskarlitski:
Thanks @henrywang for fixing the tests!

I've addressed the other two issues while merging this manually. I've also added an explanatory comment to the change in `entrypoint.sh`

Issue #327, opened 2018-08-03
Most recently committed blueprint version not getting fetched properly


jkozol:
When there are pending changes in the workspace the blueprint does not properly load its most recently committed version. This is an issue with how we fetch blueprint contents

To replicate this:
1. Create a new blueprint
2. Add a few packages
3. Log out, then log back in. The packages are still listed in the UI as expected, but they are also listed in the state for the past version of the blueprint, which is not expected.

PR #356, opened 2018-09-13
pass blueprint data that's expected on save

### LINKED ISSUE #337: welder-web strips unknown blueprint fields
### Description of problem
There are several blueprint sections that the UI does not yet support (eg. groups, customizations.*), if a blueprint is created with these from the cmdline or shipped with lorax-composer and it is then edited using the UI it will lose all of the fields that the UI doesn't know about. It should preserve them instead of removing them.

http://weldr.io/lorax/lorax-composer/lorax-composer.html

### Component (web, API, etc)
lorax-composer-19.7.21-2
cockpit-composer-0.1.2-1

jgiardino:
This closes #337

The issue with saving changes made on the Edit Blueprint page was mostly addressed in #330. But these changes now only pass in data needed to post blueprint updates, and not _everything_ like before this PR. 

However, when editing the description on the blueprint page, the groups and customizations weren't being preserved, and the updated version number wasn't getting fetched and updated in the store. Edit Description action will now pass blueprint groups and customizations when posting the updated description and also reloads the updated version number.

Since the user could make changes to the blueprint and then edit the description before saving those changes, I decided to handle this use case by doing the following:

1.  When setting the updated description, the oldest blueprint object we have in the store is posted
2. All blueprint objects are then updated with the new description and new version number
3. The present blueprint object is posted to the workspace, since the workspace was cleared when saving the updated description

### DIFF
--- a/components/Modal/CreateImage.js
+++ b/components/Modal/CreateImage.js
@@ -75,10 +75,9 @@ class CreateImage extends React.Component {
         // and call create image     
         Promise.all([BlueprintApi.reloadBlueprintDetails(this.props.blueprint)])
           .then(data => {
-            const blueprintToSet = this.props.blueprint;
-            blueprintToSet.name = data[0].name;
-            blueprintToSet.description = data[0].description;
-            blueprintToSet.version = data[0].version;
+            const blueprintToSet = Object.assign({}, this.props.blueprint, {
+              version: data[0].version,
+            });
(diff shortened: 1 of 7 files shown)


jgiardino:
Note, due to issue #327, the workflow where the user edits the description when there are changes saved in the workspace will actually result in committing the workspace changes, instead of the version of the blueprint prior to the workspace changes.

PR #359, opened 2018-09-24
Update Discard Changes to delete workspace

### LINKED ISSUE #327: Most recently committed blueprint version not getting fetched properly

jgiardino:
This closes #327.

These updates will delete a workspace if one exists when the user selects an action to discard the changes in a workspace, using either the Discard Changes button or the Undo button.

After the workspace is deleted, an additional api call is made to get the blueprint so that we can load the correct blueprint contents in the UI.

A couple of things to note:
- I'd like for the user to be able to hit Redo to get changes back that were discarded. I think this could be a simple change, but will open a separate PR for that.
- While working on this, I noticed that there are other places where we don't handle components that are stored in the blueprint as 'modules' very well, like removing components that are modules, or reloading the blueprint when changes are made. I will open a separate PR for those updates.

### DIFF
--- a/core/sagas/blueprints.js
+++ b/core/sagas/blueprints.js
@@ -1,10 +1,11 @@
 import { call, put, takeEvery, select } from 'redux-saga/effects';
 import {
-  fetchBlueprintInfoApi, fetchBlueprintNamesApi, fetchBlueprintContentsApi, fetchWorkspaceBlueprintContentsApi,
+  fetchBlueprintInfoApi, fetchBlueprintNamesApi, fetchBlueprintContentsApi,
   deleteBlueprintApi, setBlueprintDescriptionApi,
   createBlueprintApi,
   depsolveComponentsApi,
   commitToWorkspaceApi, fetchDiffWorkspaceApi,
+  deleteWorkspaceApi,
 } from '../apiCalls';
 import {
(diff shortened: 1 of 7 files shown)


core/sagas/blueprints.js
```diff
@@ -182,6 +175,31 @@ function* commitToWorkspace(action) {
   }
 }
 
+function* deleteWorkspace(action) {
```
jgiardino: There's a lot of duplication between this function and fetchBlueprintContents, and I couldn't figure out a simple way to handle that without making major changes. Any recommendations are welcome.

larskarlitski: Yeah, it's duplicated, but hard to entangle right now. I think it's fine to keep it as is for now.

jgiardino:
Thanks for the review and for the suggestion, @larskarlitski! I updated the commit message.

martinpitt:
Sorry for the gazillion cockpit-related PRs, I botched testing the enabling of Cockpit CI for welder. The next force-push will clean them up.

larskarlitski:
Thanks Jenn!

PR #360, opened 2018-09-26
Add support for Modules during Add/Remove

jgiardino:
Previously, the UI only handled packages during add/remove
and ignored modules, which resulted in odd behavior, like
modules disappearing from the UI during add, or not being
able to remove modules from a blueprint. These updates
add support for blueprints that have components listed as
modules in the blueprint contents.


martinpitt:
Sorry for the gazillion cockpit-related PRs, I botched testing the enabling of Cockpit CI for welder. The next force-push will clean them up.

jgiardino:
@larskarlitski If you have time, would you review this one?

PR #361, opened 2018-09-26
README: We're running tests on Cockpit's CI now

larskarlitski:
Small fix to `README.md` to check that the Cockpit's bots are running.


martinpitt:
@henrywang: So this ran the tests, and there even is a 
[result html artifact](https://fedorapeople.org/groups/cockpit/logs/pull-361-20180926-112727-b8fd998d-cockpit-rhel-7-6/wdio-report.html). All tests apparently succeeded, but do you know why it still exited with 2?

PR #362, opened 2018-09-26
Find the root cause of exiting with none 0 code and fix it.

henrywang:
It's caused by return value. The run_webdriver_tests() has to return ture or false in this case. Adding return value in run_webdriver_tests() fixed this issue.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -41,10 +41,16 @@ def run_webdriver_tests(machine, env=[]):
         Return success of the tests
         (True: all passed, False: at least one failed)
     """
-    cmd_parts = ["cd /root/end-to-end;"] + env + ["npm run test", ">&2"]
+    return_state = True
+    cmd_parts = ["cd /root/end-to-end;"] + env + ["npm run test"]
     print(" ".join(cmd_parts))
 
-    machine.execute(" ".join(cmd_parts), timeout=3600)
+    try:
+        machine.execute(" ".join(cmd_parts), timeout=3600)
+    except Exception as e:
+        return_state = False
+        print(e)
+    return return_state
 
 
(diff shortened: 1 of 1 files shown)


henrywang:
@martinpitt, the issue should be fixed in this PR. But the cockpit/rhel-7-6 test did not start. Could you please take a look at it? Thanks.

test/check-application
```diff
@@ -41,10 +41,15 @@ def run_webdriver_tests(machine, env=[]):
...
-    cmd_parts = ["cd /root/end-to-end;"] + env + ["npm run test", ">&2"]
+    return_state = True
+    cmd_parts = ["cd /root/end-to-end;"] + env + ["npm run test"]
     print(" ".join(cmd_parts))
 
-    machine.execute(" ".join(cmd_parts), timeout=3600)
+    try:
+        machine.execute(" ".join(cmd_parts), timeout=3600)
+    except Exception:
+        return_state = False
```
martinpitt: Please no blanket Exception catching. This will easily hide stupid things like `AttributeError` or `NameError`. You should always catch specific exceptions only, like `testvm.Timeout`. But in fact I think this should not catch an exception at all -- if the `.execute()` fails, then the program will just crash with an unhandled exception with all the details, and non-zero. So IMHO the only thing to do here is to append a `return True`.

martinpitt: As per IRC conversation: It seems we don't want to abort here, to make `--sit` work. So how about something like
```py
try:
    machine.execute(" ".join(cmd_parts), timeout=3600)
except (subprocess.CalledProcessError, RuntimeError) as e:
    sys.stderr.write("ERROR: %s\n", e)
    return_state = False

PR #365, opened 2018-09-30
Add selenium debug support.

henrywang:
We can use VNC to watch the end to end test process in real browser (firefox and chrome).

@martinpitt, could you please forward port 5901 and 5902 in selenium VM? Thanks. 

 - [x] enable -debug containers on Cockpit selenium VM: https://github.com/cockpit-project/cockpit/pull/10191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,10 @@ vm: $(VM_IMAGE)
 check: $(VM_IMAGE)
 	PYTHONPATH=`pwd`/bots/machine test/check-application -v -b $(BROWSER) -C 2 -M 2048 $(TEST_OS)
 
+# debug end to end test
+debug-check:
+	DEBUG_TEST=true $(MAKE) check
+
 # checkout Cockpit's bots/ directory for standard test VM images and API to launch them
 # must be from cockpit's master, as only that has current and existing images; but testvm.py API is stable
 bots:
--- a/test/check-application
+++ b/test/check-application
@@ -61,17 +61,18 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
                                   networking=network.host(),
                                   memory_mb=memory, cpus=cpus)
     selenium = windows = None
+    networking_selenium = network.host(forward={"5901": 5901, "5902": 5902})
 
     # settings on selenium VM
     if 'edge' in browser:
         windows = testvm.VirtMachine(image='windows-10', verbose=verbose,
-                                     networking=network.host(),
+                                     networking=networking_selenium,
                                      memory_mb=4096, cpus=4)
         # wdio.conf.js needs key word MicrosoftEdge for edge browser
         browser = 'MicrosoftEdge'
     else:
         selenium = testvm.VirtMachine(image="selenium", verbose=verbose,
-                                      networking=network.host())
+                                      networking=networking_selenium)
     try:
         composer.start()
         if selenium:
@@ -83,7 +84,7 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
             selenium.upload(["selenium_start.sh"], "/root/")
 
             # start selenium on the server
-            selenium.execute(command="/root/selenium_start.sh")
+            selenium.execute("/root/selenium_start.sh")
         elif windows:
             selenium = windows
             selenium.pull(selenium.image_file)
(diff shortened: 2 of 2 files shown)


test/selenium_start.sh
```diff
@@ -28,7 +28,15 @@ systemctl start docker
 
 docker run -d -p 4444:4444 --name selenium-hub selenium/hub:3
 wait_curl /grid/console "Grid Console"
-docker run -d --link selenium-hub:hub --shm-size=512M selenium/node-chrome:3
-wait_curl /grid/console "browserName: chrome"
-docker run -d --link selenium-hub:hub --shm-size=512M selenium/node-firefox:3
-wait_curl /grid/console "browserName: firefox"
\ No newline at end of file
+
+if [ -z ${DEBUG_TEST+x} ]; then
```
martinpitt: Note that this is a rather unusual syntax. `[ -z "$DEBUG_TEST" ]` seems easier to read and it's a much more common quoting style.

test/selenium_start.sh
```diff
@@ -28,7 +28,15 @@ systemctl start docker
...
+    docker run -d --link selenium-hub:hub --shm-size=512M selenium/node-chrome:3
+    wait_curl /grid/console "browserName: chrome"
+    docker run -d --link selenium-hub:hub --shm-size=512M selenium/node-firefox:3
+    wait_curl /grid/console "browserName: firefox"
+else
+    docker run -d -P -p 5901:5900 --link selenium-hub:hub -e VNC_NO_PASSWORD=1 --shm-size=512M selenium/node-chrome-debug:3
+    wait_curl /grid/console "browserName: chrome"
+    docker run -d -P -p 5902:5900 --link selenium-hub:hub -e VNC_NO_PASSWORD=1 --shm-size=512M selenium/node-firefox-debug:3
+    wait_curl /grid/console "browserName: firefox"
+fi
```
martinpitt: Is there any downside with always running the -debug variants? This of course requires changing the containers on Cockpit's selenium image, but that's fine -- we would need to put them on the image for this option anyway. So instead of having two, I'd much rather continue to only have one.

test/check-application
```diff
@@ -96,6 +100,19 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
...
         selenium_address = "10.111.112.10" if selenium else ""
 
+        if "DEBUG_TEST" in os.environ:
+            try:
+                subprocess.call(["vncviewer", selenium_address+":5901", "&"])
+                subprocess.call(["vncviewer", selenium_address+":5902", "&"])
+            except OSError as e:
+                if e.errno == os.errno.ENOENT:
+                    print("Please install vncviewer by \
+                          'dnf install tigervnc'.")
```
martinpitt: Please avoid the line break here (and below), otherwise you have a lot of whitespace in the resulting string. The lines are short enough anyway.

Makefile
```diff
@@ -135,6 +135,10 @@ vm: $(VM_IMAGE)
 check: $(VM_IMAGE)
 	PYTHONPATH=`pwd`/bots/machine test/check-application -v -b $(BROWSER) -C 2 -M 2048 $(TEST_OS)
 
+# debug end to end test
+debug-test: $(VM_IMAGE)
+	DEBUG_TEST=true PYTHONPATH=`pwd`/bots/machine test/check-application -v -b $(BROWSER) -C 2 -M 2048 $(TEST_OS)
```
martinpitt: how about calling `DEBUG_TEST=true $(MAKE) check` to avoid duplicating the actual invocation? And maybe rename this to `debug-check` for consistency with `check`?

martinpitt:
> could you please forward port 5901 and 5902 in selenium VM? Thanks.

This is not a property of the images, but you set this up in the `testvm.VirtNetwork(0)` call in welder's check-application. Something like `testvm.VirtNetwork(0, forward={ "5901": 5900, "5902": 6900})`. The key is the port in the VM, the value is the port number on the host. The actual port number is that base number plus some number to avoid conflicts (depending on how many tests run in parallel).

martinpitt:
I now landed the selenium image changes to have the -debug variants. So this needs a rebase, and the above changes.

henrywang:
@martinpitt, updated based on your comments and some new thoughts. Could you please review it? Thanks.

test/check-application
```diff
@@ -61,17 +61,26 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
...
     selenium = windows = None
+    networking_selenium = network.host(forward={"5901": 5901, "5902": 5902})
+    # parse libvirt forward setting string:
+    # "hostfwd=tcp:127.0.0.2:5903-:5901,hostfwd=tcp:127.0.0.2:5904-:5902"
+    # the parsing result will be a dict like:
+    # {'22': '2202', '9090': '9092', '5901': '5903', '5902': '5904'}
+    fwd_port = {}
+    for rule in networking_selenium["forwards"].split(","):
+        port_info = rule.split(":")
+        fwd_port[port_info[-1]] = port_info[-2][:-1]
```
martinpitt: Please use `networking_selenium["forward"]`, then you don't need to parse the QEMU option string.

test/check-application
```diff
@@ -103,6 +112,22 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
         # Now actually run the tests
         selenium_address = "10.111.112.10" if selenium else ""
 
+        if "DEBUG_TEST" in os.environ:
+            try:
+                vnc_cmd = "vncviewer 127.0.0.2:{} &"
```
martinpitt: And likewise, don't hardcode the IP.  `networking_selenium["forward"]["5901"], is the full `host:address` value.

test/check-application
```diff
@@ -103,6 +112,22 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
...
+                if browser == "chrome":
+                    os.system(vnc_cmd.format(fwd_port["5901"]))
+                if browser == "firefox":
+                    os.system(vnc_cmd.format(fwd_port["5902"]))
```
martinpitt: Can you please use

    subprocess.call(["vncviewer", networking_selenium["forward"]["5901"]])

here? This avoids an intermediate shell, thus is safer. You can also use `check_call` if you want to get an exception on non-zero exit codes.

henrywang:
@martinpitt, thanks for your feedback. Updated.

PR #373, opened 2018-10-05
Creates user-friendly labels to display for image types

### LINKED ISSUE #371: User-friendly labels for image types
In the Create Image dialog shown below, user-friendly labels for image types should display.

image

Currently, the image types are defined in composer-cli as:

- ext4-filesystem
- live-iso
- partitioned-disk
- qcow2
- tar

But we will also be adding options for:

- AWS
- Windows virtual machine optimized for Azure
- [VMWare vSphere image](https://en.wikipedia.org/wiki/VMware_vSphere)

jgiardino:
Closes #371 

Adds user-friendly labels to display in the Create Image dialog, as shown below. 

image

Please note - Labels are provided for new images types added in https://github.com/weldr/lorax/pull/475, but are not shown in the example above since I am currently only running the latest version of lorax-composer that is available to install.

### DIFF
--- a/core/apiCalls.js
+++ b/core/apiCalls.js
@@ -65,8 +65,22 @@ export function fetchBlueprintInfoApi(blueprintName) {
 }
 
 export function fetchModalCreateImageTypesApi() {
+  const imageTypeLabels = {
+    'ami': 'Amazon Machine Image Disk (.ami)',
+    'ext4-filesystem': 'Ext4 File System Image (.img)',
+    'live-iso': 'Live Bootable ISO (.iso)',
+    'partitioned-disk': 'Raw Partitioned Disk Image (.img)',
+    'qcow2': 'QEMU QCOW2 Image (.qcow2)',
+    'tar': 'TAR Archive (.tar)',
+    'vhd': 'Azure Disk Image (.vhd)',
+    'vmdk': 'VMware Virtual Machine Disk (.vmdk)'
+  };
   const imageTypes = utils.apiFetch(constants.get_image_types)
-    .then(data => data.types)
+    .then(data => data.types.map(type => {
+      return Object.assign({}, type,
+        {label: imageTypeLabels[type.name] || type.name}
+      );
+    }))
     .catch(e => console.log('Error getting component types', e));
   return imageTypes;
 }
--- a/components/Modal/CreateImage.js
+++ b/components/Modal/CreateImage.js
@@ -138,7 +138,7 @@ class CreateImage extends React.Component {
                   <div className="col-sm-9">
                     <select className="form-control" value={this.state.imageType} onChange={this.handleChange}>
                       {this.props.imageTypes !== undefined && this.props.imageTypes.map((type, i) =>
-                        <option key={i} value={type.name} disabled={!type.enabled}>{type.name}</option>
+                        <option key={i} value={type.name} disabled={!type.enabled}>{type.label}</option>
                       )}
                     </select>
                   </div>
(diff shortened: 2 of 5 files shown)


stefwalter:
I think this needs a test change in: https://github.com/weldr/welder-web/blob/master/test/end-to-end/pages/createImage.js

@jgiardino Is that something you're up for?

jgiardino:
@stefwalter Good call! I'll take a look and let you know if I'm stuck on this.

jgiardino:
@henrywang Could you review the updates I made for the tests?

henrywang:
@jgiardino, I'll give you a test case code diff when I'm back on Monday. I'm still on public holiday. Sorry for late response.

core/apiCalls.js
```diff
@@ -66,11 +66,47 @@ export function fetchBlueprintInfoApi(blueprintName) {
...
     .catch(e => console.log('Error getting component types', e));
   return imageTypes;
 }
 
+export function mapImageTypeLabels(name) {
+  let label = '';
+  switch (name) {
+    case 'ami':
+      label = 'Amazon Machine Image Disk (.ami)';
+      break;
```
larskarlitski: I prefer using a dictionary, because that's more conventional in JavaScript than `switch`. Also, it avoids the extra function. Something like this at the top of the file:

```js
const imageTypeLabels = {
  "ami": "Amazon Machine Image Disk (.ami)",
  // ...
}
```

core/apiCalls.js
```diff
@@ -66,11 +66,47 @@ export function fetchBlueprintInfoApi(blueprintName) {
 
 export function fetchModalCreateImageTypesApi() {
   const imageTypes = utils.apiFetch(constants.get_image_types)
-    .then(data => data.types)
+    .then(data => data.types.map(type => {
+      type.label = mapImageTypeLabels(type.name);
+      return type;
```
larskarlitski: Changing the input argument to of `map` often leads to confusions. It's not really a problem here, because `data` isn't used after calling `map`, but I prefer creating a new object:

```js
data.types.map(type => {
  return {
    name: type.name,
    label: imageTypeLabels[name] || name // see my suggestion below
  };
})
```

core/apiCalls.js
stefwalter: We still need to pass through the image type directly, in the case where no matching label is found.  So this may not avoid an extra function.

jgiardino:
Thanks for the comments, @larskarlitski and @stefwalter!

I went with @larskarlitski's suggestion on using the dictionary method instead of a switch, and it seemed to handle the missing label case fine when I removed one of the labels (e.g. using `label: imageTypeLabels[type.name] || type.name`).

I used @larskarlitski's suggestion to create a new object, but since there's more being fetched than just name (i.e. whether it's enabled or not), I used `Object.assign`.

I saw your comment on https://github.com/weldr/lorax/pull/496, @larskarlitski. I can go ahead and add that type to this PR, but I'm not really sure what the label would be. This is my best guess: **OpenStack QCOW2 Image (.qcow2)**

henrywang:
@jgiardino, all tests passed now. Thanks @larskarlitski to re-start test.

larskarlitski:
Thanks for the changes, Jenn!

> I saw your comment on weldr/lorax#496, @larskarlitski. I can go ahead and add that type to this PR, but I'm not really sure what the label would be. This is my best guess: OpenStack QCOW2 Image (.qcow2)

We can follow-up once that lands in lorax-composer.

Issue #324, opened 2018-08-01
Refactor end-to-end tests


jgiardino:
Currently the selectors used to identify the UI elements for end-to-end tests are flimsy, and easily broken when UI updates are made.

Instead, any element that is targeted in an end-to-end test should have an html `id` attribute assigned which can be used to identify the element in the test.

jgiardino:
Currently, these tests cannot be tested with a local build.

jgiardino:
@xiaofeng, can you provide comments on the suggestions in this issue. You mentioned during standup that the latest version of firefox doesn't support using the `id` attribute as a selector. Is that in the context of tests? or just in general (tests and css)?

jgiardino:
@xiaofeng, are these instructions still valid for running end-to-end tests: https://github.com/weldr/welder-web/blob/master/test/end-to-end/README.md

henrywang:
Just run ```make lorax-test```.
The [PR357](https://github.com/weldr/welder-web/pull/357) changes a lot of things. But for now, ```make lorax-test``` works for you.

henrywang:
Here's a detailed description about locating element by id - https://github.com/webdriverio/webdriverio/issues/2392
That means ```#id``` will not be supported by geckodriver. But ```body #id``` still works.

No better way to locate element then by id. The best id attribute should be unique and meaningful.

The unique means every id in html should have unique name. That'll be easy to locate element when write UI automation test and easy to work with automation code.

The meaningful means the id name should represent which content this id identifies. Reading react and automation code becomes easy if id is meaningful. e.g. In the edit blueprint page, each available or selected component should have id attribute with a meaningful name, like ```[data-component="bash"]```.

larskarlitski:
Not sure if that's what you're suggesting @henrywang, but I've used `data` attributes instead of ids before and that worked quite nicely. For example in https://github.com/weldr/welder-web/commit/0be7c1c89a71a1e69c86c775920c5cbc72497e64

henrywang:
@larskarlitski, I found your change. This change saves at least 6 lines of code and make test code easy to understand. From my perspective, both `id` and `data` are OK for me. But after some research, I think `data-*` is more reasonable for this use case.
The definition and usage of `id`:

> The id attribute specifies a unique id for an HTML element (the value must be unique within the HTML document).
The id attribute is most used to point to a style in a style sheet, and by JavaScript (via the HTML DOM) to manipulate the element with the specific id.

[...]

PR #375, opened 2018-10-05
Adding WIP on #324

sfondell:
WIP: Sorry it took me so long to push this ! I'm not quite sure I did it correctly so any comments/suggestions/explanations would be wonderful so that I could do the rest.


components/Modal/CreateImage.js
```diff
@@ -96,6 +96,7 @@ class CreateImage extends React.Component {
       <div
         className="modal fade"
         id="cmpsr-modal-crt-image"
+        data-id="cmpsr-modal-crt-image"
```
jgiardino: We don't need the extra `data-id` attribute on this element. Since we already have an `id` attribute here, we can use that instead of create a new `data-id` one.

For example, the id that's defined here is currently being used in [test/end-to-end/pages/createImage.js, line 12](https://github.com/weldr/welder-web/pull/375/files#diff-886584fb195b433a216cecc7e230cd8cR12) as `#cmpsr-modal-crt-image`

components/Modal/CreateImage.js
```diff
@@ -128,12 +129,16 @@ class CreateImage extends React.Component {
                   ><FormattedMessage defaultMessage="Blueprint" /></label>
                   <div className="col-sm-9">
                     <p className="form-control-static">{this.props.blueprint.name}</p>
+                    	<label
+                    		data-id="cmpsr-form-ctrl"
+                    	>
```
jgiardino: The element we want to target is the paragraph `<p>` element that contains the blueprint name. Instead of adding a new `<label>` element here, the `data-id` attribute should be added to the existing `<p>` element. 

Also, `"cmpsr-form-ctrl"` is a little too generic for our purposes. Maybe `"cmpsr-form-blueprint"`?

So after these changes, it would be something like:
```
<p 
  className="form-control-static"
  data-id="cmpsr-form-blueprint"
>{this.props.blueprint.name}</p>
```

BTW, in looking at the existing `id` values being used in this form, they're not that great and they're all really generic. Sorry for having bad examples for you to work from. 😬

test/end-to-end/pages/createImage.js
```diff
@@ -9,7 +9,7 @@ module.exports = class CreateImagePage extends MainPage {
 
     // ---- Root element selector ---- //
     // Create Blueprint dialog root selector
-    this.dialogRootElement = 'div[id="cmpsr-modal-crt-image"]';
+    this.dialogRootElement = `${this.dialogRootElement} #cmpsr-modal-crt-image`;
```
jgiardino: I like replacing this:
`'div[id="cmpsr-modal-crt-image"]'`
with this:
`#cmpsr-modal-crt-image`

But I don't think we need the `${this.dialogRootElement}` in `${this.dialogRootElement} #cmpsr-modal-crt-image`. I think this is where `this.dialogRootElement` is getting defined to be used for other selectors below.

@henrywang Can you please confirm if this is true? Can we just use this here:
`this.dialogRootElement = '#cmpsr-modal-crt-image'`;

test/end-to-end/pages/createImage.js
```diff
@@ -20,11 +20,11 @@ module.exports = class CreateImagePage extends MainPage {
     this.labelCreateImage = `${this.dialogRootElement} #myModalLabel`;
 
     // Blueprint Name label
-    this.labelBlueprintName = `${this.dialogRootElement} p[class="form-control-static"]`;
+    this.labelBlueprintName = `${this.dialogRootElement} #cmpsr-form-ctrl`;
```
jgiardino: If we do use the the `data-id` attribute as shown below:
```
<p 
  className="form-control-static"
  data-id="cmpsr-form-blueprint"
>{this.props.blueprint.name}</p>
```
then we can't use the `#` with this attribute value as a selector.

The `#` would only precede the value provided for an `id` attribute. But most attributes would need to be identified using the [attribute css selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as shown below:

```
this.labelBlueprintName = `${this.dialogRootElement} [data-id="cmpsr-form-blueprint"]`;
```

components/Modal/CreateImage.js
```diff
@@ -128,12 +129,16 @@ class CreateImage extends React.Component {
...
                   </div>
                 </div>
                 <div className="form-group">
                   <label
                     className="col-sm-3 control-label"
                     htmlFor="textInput-modal-markup"
+                    data-id="cmpsr-modal-markup1"
```
jgiardino: Instead of adding a `data-id` attribute here, could you add an `id` to the `<select>` as noted in a [previous comment](https://github.com/weldr/welder-web/pull/375/files#diff-886584fb195b433a216cecc7e230cd8cR27)

components/Modal/CreateImage.js
```diff
@@ -147,6 +152,7 @@ class CreateImage extends React.Component {
                   <label
                     className="col-sm-3 control-label"
                     htmlFor="textInput2-modal-markup"
+                    data-id="cmpsr-modal-markup2"
```
jgiardino: Instead of adding a `data-id` attribute here, could you add an `id` to the `<select>` as noted in a [previous comment](https://github.com/weldr/welder-web/pull/375/files#diff-886584fb195b433a216cecc7e230cd8cR27)

jgiardino:
Hey @sfondell, this was a great first attempt for someone who's never worked with css selectors before! I included some comments that will hopefully point you in the right direction. Please let me know if you need additional clarification for any of these.

Also, @larskarlitski has pointed out to me that whenever we push our first commit, the last line of the commit message should say what issue it closes as:
```
(last line of commit message)

Closes #324
```
This will automatically close the issue when the PR merges, which is nice.

components/Modal/CreateImage.js
sfondell: Are `id` and `data-id` pretty much the same in this instance? If an element already has an `id` attribute does that mean nothing needs to be changed?

components/Modal/CreateImage.js
sfondell: This makes sense ! I'll fix this

test/end-to-end/pages/createImage.js
sfondell: I saw `${this.dialogRootElement}` used elsewhere in the file and figured it needed to be used here, would it be redundant to put it on the right side in this case?

test/end-to-end/pages/createImage.js
sfondell: Why does this one need to be identified using an attribute css selector while the other one is able to be referenced with `#`?

components/Modal/CreateImage.js
jgiardino: Correct!

test/end-to-end/pages/createImage.js
jgiardino: > would it be redundant to put it on the right side in this case?

I'm not sure what you mean by "right side"

test/end-to-end/pages/createImage.js
jgiardino: I'm not really sure what the history behind this is, so I'm making up my own version here that hopefully clarifies this:

CSS was invented so that developers could assign styles to elements. Elements could be selected by their `id` attributes using `#`, by `class` attributes using `.`, or by elements themselves using their tag name (e.g. `p` or `div`). But then someone realized that we should be able to style things using any attribute, so the catch-all attribute selector was added to the spec (i.e. `[{attribute}]`). And since then, it's been agreed to that selecting elements by `id` for styling with css is bad practice.

We're using the CSS selector syntax here since the tests use the CSS selectors to identify elements. And in this case, `id` is ideal if it's present because it has to be unique. But it isn't always present. It's most likely going to be used for form elements, but it might not be present for other things like buttons and links.

components/Modal/DeleteBlueprint.js
```diff
@@ -30,9 +30,11 @@ class DeleteBlueprint extends React.Component {
                 className="close"
                 data-dismiss="modal"
               >
-                <span className="pficon pficon-close"></span>
+                <span 
+                className="pficon pficon-close"
+                id="pficon-close"></span>
               </button>
-              <h4 className="modal-title" id="myModalLabel"><FormattedMessage defaultMessage="Delete Blueprint" /></h4>
+              <h4 className="modal-title" data-id="myModalLabel"><FormattedMessage defaultMessage="Delete Blueprint" /></h4>
```
jgiardino: In this case we want the `id`. 

The rules to follow for `data-id` vs `id` are:
- If we need to select an element, check if an `id` attribute already exists.
  - If `id` already exists, then use the `id`
  - If `id` doesn't exist, then check if the element is using a class that no other element in that control is using
    - If a fairly unique `class` exists, then use the class
    - If a class doesn't exist or is used by other elements, then add a `data-id` attribute with a value that identifies the element.

If an `id` is used, then it's being used to associate html elements with other elements. This is most likely going to be used in form elements. But here, the `id` value is being referenced by the modal parent `<div>` in the attribute `aria-labelledby="myModalLabel"` (i.e. the label for this modal is provided by this `<h4>` element)

components/Modal/DeleteBlueprint.js
```diff
@@ -49,11 +51,13 @@ class DeleteBlueprint extends React.Component {
               <button
                 type="button"
                 className="btn btn-default"
+                id="btn-default"
```
jgiardino: `id` shouldn't be added here

In this case it's fairly safe to assume that this is the only element that would have the class `btn-default`, so for the test, we can use the class selector: `.btn-default`

components/Modal/DeleteBlueprint.js
```diff
@@ -49,11 +51,13 @@ class DeleteBlueprint extends React.Component {
...
                 data-dismiss="modal"
               ><FormattedMessage defaultMessage="Cancel" /></button>
               <button
                 type="button"
                 className="btn btn-danger"
+                id="btn-danger"
```
jgiardino: `id` shouldn't be added here

In this case it's fairly safe to assume that this is the only element that would have the class `btn-danger`, so for the test, we can use the class selector: `.btn-danger`

test/end-to-end/pages/deleteBlueprint.js
```diff
@@ -12,7 +12,7 @@ module.exports = class deleteBlueprint extends MainPage {
 
     // ---- Root element selector ---- //
     // Root Element for this Dialog Page
-    this.rootElement = 'div[id="cmpsr-modal-delete"] .modal-dialog .modal-content';
+    this.rootElement = '#cmpsr-modal-delete .modal-dialog .modal-content';
```
jgiardino: 👍 Yay! Great job!

test/end-to-end/pages/deleteBlueprint.js
```diff
@@ -25,17 +25,17 @@ module.exports = class deleteBlueprint extends MainPage {
 
     // ---- Page element selector ---- //
     // Page Title
-    this.labelPageTitle = `${this.headerElement} h4[class="modal-title"]`;
+    this.labelPageTitle = `${this.headerElement} [data-id="myModalLabel"]`;
```
jgiardino: Related to a previous comment above, if we keep the heading element as it originally was:
`<h4 className="modal-title" id="myModalLabel">`
Then we can use the id selector here:
this.labelPageTitle = `${this.headerElement} #myModalLabel`;

test/end-to-end/pages/deleteBlueprint.js
```diff
@@ -25,17 +25,17 @@ module.exports = class deleteBlueprint extends MainPage {
...
 
     // X Close Button
-    this.btnXClose = `${this.headerElement} span[class="pficon pficon-close"]`;
+    this.btnXClose = `${this.headerElement} .pficon-close`;
```
jgiardino: Yes, this is perfect!

test/end-to-end/pages/deleteBlueprint.js
```diff
@@ -25,17 +25,17 @@ module.exports = class deleteBlueprint extends MainPage {
...
 
     // Blueprint Name label
     this.labelBlueprintName = `${this.bodyElement} p strong`;
 
     // Delete and Cancel button
-    this.btnDelete = `${this.footerElement} button[class="btn btn-danger"]`;
-    this.btnCancel = `${this.footerElement} button[class="btn btn-default"]`;
+    this.btnDelete = `${this.footerElement} #btn-danger`;
+    this.btnCancel = `${this.footerElement} #btn-default`;
```
jgiardino: Related to a previous comment, since we can probably rely on the classes in these elements to identify them, then we can use the class selectors here. For example:
```
    this.btnDelete = `${this.footerElement} .btn-danger`;
    this.btnCancel = `${this.footerElement} .btn-default`;
```

components/Modal/CreateImage.js
```diff
@@ -129,7 +129,10 @@ class CreateImage extends React.Component {
                     className="col-sm-3 control-label"
                   ><FormattedMessage defaultMessage="Blueprint" /></label>
                   <div className="col-sm-9">
-                    <p className="form-control-static">{this.props.blueprint.name}</p>
+                    <p 
+                    className="form-control-static" 
+                    id="cmpsr-form-blueprint"
```
jgiardino: This is a case where we want to use `data-id`, since we are adding an identifying attribute for the test.

Here you can just change this to `data-id="cmpsr-form-blueprint"`

test/end-to-end/pages/createImage.js
```diff
@@ -20,7 +20,7 @@ module.exports = class CreateImagePage extends MainPage {
     this.labelCreateImage = `${this.dialogRootElement} #myModalLabel`;
 
     // Blueprint Name label
-    this.labelBlueprintName = `${this.dialogRootElement} p[class="form-control-static"]`;
+    this.labelBlueprintName = `${this.dialogRootElement} #cmpsr-form-blueprint`;
```
jgiardino: Related to the previous comment, since we are using `data-id` instead of `id`, then the attribute selector should be updated to:
```
this.labelBlueprintName = `${this.dialogRootElement} [data-id="cmpsr-form-blueprint"]`;
```

test/end-to-end/pages/createImage.js
```diff
@@ -20,7 +20,7 @@ module.exports = class CreateImagePage extends MainPage {
...
 
     // Image Type select
     this.selectImageType = `${this.dialogRootElement} label[for="textInput-modal-markup"] + div select`;
```
jgiardino: Now that you've rebased, you can update this selector to not be so confusing.

The original selector is looking for a `<select>` inside a `<div>` that's preceded by a label with the html attribute `for="textInput-modal-markup"`. 

But this `<select>` has an `id` that we can use instead:
```
                    <select 
                      id="textInput-modal-markup" 
                      className="form-control" 
                      required 
                      value={this.state.imageType} 
                      onChange={this.handleChange}
                    >
```
So now we can simplify the selector to:
```
this.selectImageType = `${this.dialogRootElement} #textInput-modal-markup`;
```

components/Modal/CreateBlueprint.js
```diff
@@ -176,7 +176,9 @@ class CreateBlueprint extends React.Component {
                       onBlur={(e) => this.handleErrors(e.target.value)}
                     />
                     {createBlueprint.errorNameVisible &&
-                      <span className="help-block"><FormattedMessage defaultMessage="A blueprint name is required." /></span>
+                      <span 
+                      className="help-block"
+                      id="help-block-msg"><FormattedMessage defaultMessage="A blueprint name is required." /></span>
```
jgiardino: This might be a case where we should use `data-id` since the class `help-block` is being used for other messages that could display in the form.

I would just change `id` to `data-id` here.

Also, if you run `npm run lint`, it will probably complain about your indentation and tell you how to fix it.

test/end-to-end/pages/createBlueprint.js
```diff
@@ -41,7 +41,7 @@ module.exports = class CreateBlueprintPage extends BlueprintPage {
 
     // Help-block error message
     this.varHelpBlockMsg = 'A blueprint name is required.';
-    this.spanHelpBlockMsg = `${this.dialogRootElement} span[class="help-block"]`;
+    this.spanHelpBlockMsg = `${this.dialogRootElement} #help-block-msg`;
```
jgiardino: After making the updates in my previous comment, you'll need to replace the `id` selector with the `data-id` attribute selector.

jgiardino:
Hey @sfondell, I see you're making great progress on this! I included some comments to help you finish your updates.

sfondell:
These files should be *hopefully* all correct now that I've rebased and implemented your suggestions, going forward which other tests also need to be changed? I noticed that some files (like deleteBlueprint, createBlueprint, and createImage) have corresponding ones in test/end-to-end/pages and components/Modal but some don't.

larskarlitski:
Rebased, resolved conflicts, squashed all commits and force-pushed together with Sophia. Tests should be able to run now.

components/Modal/CreateImage.js
```diff
@@ -129,7 +129,10 @@ class CreateImage extends React.Component {
                     className="col-sm-3 control-label"
                   ><FormattedMessage defaultMessage="Blueprint" /></label>
                   <div className="col-sm-9">
-                    <p className="form-control-static">{this.props.blueprint.name}</p>
+                    <p
+                    className="form-control-static"
+                    data-id="cmpsr-form-blueprint"
+                    >{this.props.blueprint.name}</p>
```
jgiardino: Did you run `npm run lint` yet? The indentation here looks like something it would catch.

larskarlitski: I did not, sorry. Fixed the errors.

henrywang:
@sfondell, could you please update id css selector to [id="xxx"], because "#xxx" will not work in new Geckodriver (webdriver for firefox). Thanks.
https://github.com/webdriverio/webdriverio/issues/2392

sfondell:
@henrywang Done ! Is this alright?

henrywang:
@sfondell, PR #393 got merged already. Could you please rebase master? The tests will be triggered. Thanks.

larskarlitski:
> @sfondell, PR #393 got merged already. Could you please rebase master? The tests will be triggered. Thanks.

Thanks @henrywang! I've rebased and squashed the two commits so that the tests can run before @sfondell's start of day.

sfondell:
Looks like the RHEL 7.6 Chrome tests are failing, how would I go about fixing that?

martinpitt:
@sfondell : The chrome tests are known to be brittle. I retried the test.

henrywang:
@sfondell, there's a rebase conflict in Travis CI. Could you please take a look? Thanks.
```
Using index info to reconstruct a base tree...
M	test/end-to-end/pages/createBlueprint.js
M	test/end-to-end/pages/createImage.js
Falling back to patching base and 3-way merge...
Auto-merging test/end-to-end/pages/createImage.js
CONFLICT (content): Merge conflict in test/end-to-end/pages/createImage.js
Auto-merging test/end-to-end/pages/createBlueprint.js
error: Failed to merge in the changes.
Patch failed at 0001 test: Refactor css selectors
The copy of the patch that failed is found in: .git/rebase-apply/patch
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```

martinpitt:
@jgiardino : Is this still actually relevant? This PR contains some bogus temporary files, needs rebasing (with resolving a lot of conflicts), and does not describe what it's actually trying to do.

KKoukiou:
It's outdated and inactive for long time. Please re-open if needing ot work on this again.

PR #377, opened 2018-10-08
Updates Create Image modal to not have a default image type

### LINKED ISSUE #374: Default image type changes based on available types
I'd prefer if there's a specific default image type in the "Create Image" dialog.

Right now, it chooses the first one in the list that lorax-composer returns. This means it changes when that list changes. That's admittedly not very often, but confused me just now :)

jgiardino:
Instead of defaulting to the first image type in the select menu, the select menu displays "Select one" by default. This is a disabled option which means it can't be selected in the menu.

This menu is also indicated as required to prompt the user to make a selection, and the submit button is disabled until an option is selected.

Example of modal when it first displays:
image

Example of select menu when open:
image

Closes #374

### DIFF
--- a/components/Modal/CreateImage.js
+++ b/components/Modal/CreateImage.js
@@ -18,6 +18,9 @@ const messages = defineMessages({
   warningUnsaved: {
     defaultMessage: "This blueprint has changes that are not committed. " +
       "These changes will be committed before the image is created."
+  },
+  selectOne: {
+    defaultMessage: "Select one"
   }
 });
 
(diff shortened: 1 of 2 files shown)


jgiardino:
Hey @larskarlitski, I was trying to avoid a new text string and had left the default selection as a blank option. When I presented this to my UI review group, one of the suggestions was to have placeholder text in the menu, so I added "Select one".

Let me know if we can add this string.

Also let me know if you have any other suggestions. 😄

andreasn:
If we can't add the extra string at this point, I think having a blank dropdown is an OK experience as well, that we can live with for now.

PR #368, opened 2018-10-01
test: Support Cockpit test scenario (actually enable testing on Chrome)

martinpitt:
https://github.com/cockpit-project/cockpit/pull/10194 introduces test
scenarios in Cockpit CI, and defines two scenarios "chrome" and
"firefox". These directly map to the Makefile's `$BROWSER` variable.

- [x] land Cockpit CI test support for scenarios: https://github.com/cockpit-project/cockpit/pull/10191
- [x] fix failing tests when running with chrome: PR #389

PR #381, opened 2018-10-16
Fix sed issue and improve case stability on Chrome.

henrywang:
1. Move sed to vm.install because the sed will not work inside run if running test locally.
2. Go to view blueprint page by clicking blueprint name link instead of by URL.


martinpitt:
This seems to have broken the [firefox test](http://fedorapeople.org/groups/cockpit/logs/pull-381-20181016-133636-b91a890a-cockpit-rhel-7-6-firefox/log) as well? The [chrome failure](http://fedorapeople.org/groups/cockpit/logs/pull-381-20181016-133030-b91a890a-cockpit-rhel-7-6-chrome/log) is already present in master, downloading GPG keys is just 'orribly unreliable.

test/vm.install
```diff
@@ -43,6 +43,10 @@ ln -s /usr/local/bin/node /usr/local/bin/nodejs
 if type firewall-cmd >/dev/null 2>&1; then
     firewall-cmd --add-service=cockpit --permanent
 fi
+
+# mock compose creation calls to backend for testing purposes
+sed -i "s|\"/api/v0/compose\"|\"/api/v0/compose?test=2\"|" /usr/share/cockpit/welder/dist/main.*.js
```
martinpitt: In principle this approach is better indeed, as long as you can actually rely on the .js files from the installed rpm. When you build in production mode, these are usually compressed and minified, so that sed wouldn't work right away.

But why wouldl that not work when running the test locally? It seems to work fine. So the commit message should be clearer about the rationale.

test/end-to-end/specs/test_viewBlueprint.js
```diff
@@ -16,7 +17,8 @@ describe('View Blueprint Page', () => {
     beforeEach(() => {
       CreateBlueprintPage.newBlueprint(testData.blueprint.simple);
 
-      helper.goto(viewBlueprintPage)
+      browser
+        .click(BlueprintPage.blueprintNameSelector(testData.blueprint.simple.name))
```
martinpitt: This is also more in line in how a user would use the page, so good improvement.

test/end-to-end/.nyc_output/coverage-cab8ea038a83e03a0b5321fc690675855bfbe6c2650af50b2fbbc8e373cff0ae.json
```diff
@@ -0,0 +1 @@
+null
```
martinpitt: This should not be in the commit.

test/run
```diff
@@ -2,7 +2,4 @@
 # This is the expected entry point for Cockpit CI; will be called without
 # arguments but with an appropriate $TEST_OS
 
-# mock compose creation calls to backend for testing purposes
```
martinpitt: Please break the overly long line in the commit message.

martinpitt:
In a [chrome test run](https://fedorapeople.org/groups/cockpit/logs/pull-368-20181016-211313-449b59fa-cockpit-rhel-7-6-chrome/log) (from PR #368), it still fails.

test/run
henrywang: Updated. Thanks @martinpitt.

test/end-to-end/.nyc_output/coverage-cab8ea038a83e03a0b5321fc690675855bfbe6c2650af50b2fbbc8e373cff0ae.json
henrywang: Updated .gitignore. Thanks @martinpitt

PR #382, opened 2018-10-16
test: Only install node if it isn't already available

martinpitt:
Cockpit's rhel-7-6 image will soon pre-install node, as this is brittle
and time consuming when running for each test. As this code is not
reentrant, check if node is already installed before installing it.

With that we can update the cockpit image without breaking welder-web in
between. After that landed, this can go completely.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -8,35 +8,36 @@ NODE_VERSION="8.12.0"
 
 echo 'PATH=/usr/local/bin/:$PATH' >> /etc/bashrc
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Also see https://github.com/cockpit-project/cockpit/pull/10319

martinpitt:
Note, this also drops `ln -s /usr/local/bin/node /usr/local/bin/nodejs`, which is unnecessary as far as I can see.

PR #383, opened 2018-10-16
Makefile: bots is not a phony target

martinpitt:
bots/ is an actual directory and thus an intended make target. It should
*not* be checked out again if it already exists, as that breaks testing
a bots/ from a cockpit PR against welder-web.


martinpitt:
Priority as this breaks the intended cross-project testing.

PR #386, opened 2018-10-17
po: Update from Fedora Zanata

larskarlitski:


po/de.po
```diff
@@ -52,28 +73,28 @@ msgstr "Hinzugefügt"
...
-msgstr ""
+msgstr "Ein Fehler ist aufgetreten"
 
 #: build/messages/components/ListView/BlueprintContents.json
 #. [components.ListView.BlueprintContents.emptyStateErrorMessage]
 #. defaultMessage is:
 #. An error occurred while trying to get blueprint contents.
 msgid "An error occurred while trying to get blueprint contents."
-msgstr ""
+msgstr "Beim Versuch, Blaupauseninhalte zu erhalten, ist ein Fehler aufgetreten."
```
martinpitt: It would be nice to consistently translate "blueprint" as "Blueprint" *or* "Blaupause", not mix both. But oh well, that's not something which we should patch here, but fix in Zanata.

PR #393, opened 2018-10-31
Two updates to support nodejs 10 and Fedora 29.

henrywang:
1. Update wdio-mocha-framework to 0.6.4 to support nodejs 10.
2. Update grep regxp to support lorax-composer response on Fedora 29.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -18,10 +18,12 @@ done;
         machine.execute(script=WAIT_SELENIUM_RUNNING)
 
 
+# lorax-composer in Fedora retruns: "db_supported":true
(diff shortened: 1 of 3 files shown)


martinpitt:
I retried the [chrome test](https://fedorapeople.org/groups/cockpit/logs/pull-393-20181031-083333-2478be13-cockpit-rhel-x-chrome/log) three times now, it keeps failing. Can you please have a look?

Also,  the vm.install script needs to call something like `which node || yum install nodejs`, as it's not pre-installed on the Cockpit fedora-29 VMs.

Issue #394, opened 2018-11-01
Composer layout issue on Edge


henrywang:
The "Commit", "Discard Changes", and "Create Image" buttons should be at right up corner. But it's not rendered on the correct position on Edge. This issue cannot be found on Chrome and Firefox.
![edge_layout_issue](https://user-images.githubusercontent.com/672230/47857003-874d1e80-de23-11e8-840e-e6c80ed16fe9.PNG)

andreasn:
@jgiardino @garrett any idea what's up?

andreasn:
Poked around a bit on Edge and it seems that if I remove the property `display: -ms-grid;` from` .cmpsr-grid_wrapper` in `custom.css` it displays as it should.
I think display: -ms-grid; is a IE11 thing, and since we no longer support IE11, I think we can go with just `display: grid;`
I can cook up a quick PR with a fix.

andreasn:
Browser support for grid: https://caniuse.com/#feat=css-grid

PR #395, opened 2018-11-02
Remove deprecated -ms-grid properties

andreasn:
The usage of the -ms-grid property causes issues on Edge.
The ms-grid property was used to support IE, but since we decided
to not support IE, we can now use the standard grid css properties.
Edge has full support for grid properties since version 16.
See https://caniuse.com/#feat=css-grid

Fixes issue #394

### DIFF
--- a/public/custom.css
+++ b/public/custom.css
@@ -2,11 +2,8 @@
 /* Blueprint Edit layout */
 .cmpsr-grid__wrapper {
   display: grid;
-  display: -ms-grid;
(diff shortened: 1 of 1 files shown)


henrywang:
@andreasn, code coverage result looks good. and all testes passed.

henrywang:
@andreasn, code coverage result looks good. and all testes passed.
Verified on my local env. Issue got fixed. Thanks @andreasn.
remmina_win10_10 73 131 246_2018115-65439 995323

PR #397, opened 2018-11-06
Update end to end test README.md about running test on different OS.

henrywang:


test/end-to-end/README.md
```diff
@@ -43,10 +43,14 @@ In the event you wish to visually see what the browser is doing you will want to
 
     $ make debug-check
 
-By default the test will be run on Firefox. To run it on Chrome, a prefix ```BROWSER=chrome``` needs to be added, like:
+By default the cockpit-composer will be installed into RHEL 7.6 and test will be run on Firefox. To run it on Chrome, a prefix ```BROWSER=chrome``` needs to be added, like:
 
     $ BROWSER=chrome make check
 
+To test cockpit-composer in different OS, add prefx ```TEST_OS=fedora-29```, for example.
```
martinpitt: "prefx" typo, but I'd rather write

    ... set the `$TEST_OS` environment variable, for example:

henrywang: Thanks @martinpitt. Updated. One time all green!
Oh. I updated make file to set default TEST_OS to fedora-29.

PR #401, opened 2018-11-15
package.json: Update babel-polyfill package dependency

cockpituous:
Check for any outstanding update of point version of any outstanding NPM package dependency
 * [x] npm-update

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "npm": ">=5.0"
   },
   "dependencies": {
-    "babel-polyfill": "~6.9.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "npm": ">=5.0"
   },
   "dependencies": {
-    "babel-polyfill": "6.9.1",
+    "babel-polyfill": "^6.26.0",
```
cockpituous: There should be no "^" in the version.

PR #403, opened 2018-11-15
 Add .task file for npm-update

mvollmer:
This needs https://github.com/cockpit-project/cockpit/pull/10584 to run successfully.

I needed some package that npm-update can update, and babel-polyfill was the first in the list. We can use any other package, or maybe even all.


package.json
```diff
@@ -7,7 +7,7 @@
     "npm": ">=5.0"
   },
   "dependencies": {
-    "babel-polyfill": "~6.9.1",
+    "babel-polyfill": "6.9.1",
```
martinpitt: TBH I'd just pin *every* package in `dependencies`. Seems much safer to update them in the CI controlled fashion than having some random PR bump it and run into unrelated failures. These are much less obvious and harder to debug.

mvollmer: Sure, that works as well, and removes the blocker.

mvollmer:
Hmm, the bots don't want to touch this...

mvollmer:
@martinpitt ptal thx

martinpitt:
@mvollmer: LGTM! I checked two handfuls of bots, and they don't seem to crash, they are just uber-busy. My two image refreshes don't start either. Let's give them some time to catch up, until the afternoon, and then investigate if it's still broken.

mvollmer:
> Hmm, the bots don't want to touch this...

Probably because it had the "blocked" label.

PR #405, opened 2018-11-15
Trigger the po-refresh bot

mvollmer:
Depends on #403

### DIFF
--- a/.tasks
+++ b/.tasks
@@ -2,6 +2,7 @@
 
 # Open issues for things that need doing on a regular basis
 bots/npm-trigger
+bots/po-trigger
 
 # Find those issues and propose to do them
 bots/issue-scan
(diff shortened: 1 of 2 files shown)


mvollmer:
Zanata creds need to be available when po-refresh runs.  I'll check our options.

mvollmer:
> Zanata creds need to be available when po-refresh runs. I'll check our options.

Oh, cool, our bots are already a "Translation Maintainer" for welder-web.

martinpitt:
@mvollmer: If we give "cockpit" translation privs in zanata, it should work. Indeed our bots have a zanata token already. And https://fedora.zanata.org/project/view/welder-web/people?dswid=-6732 looks like cockpit should already have access?

mvollmer:
> I can't get make po-pull to work locally here with the token from our bots:

It does work in the pods, though.  I'll try to figure out what the difference is.

Issue #399, opened 2018-11-14
openstack is miscapitalized


andreasn:
Not sure if this string lives in the web frontend or if it's in the backend, but I'll try here first.

In the dropdown list of image types I get the string **openstack**, but it's most commonly spelled **OpenStack**. It also seems it's the only one that doesn't have a file extension format attached to it.

![screenshot from 2018-11-14 14-42-18](https://user-images.githubusercontent.com/209895/48486370-a1233400-e81b-11e8-93ac-c48e2e08c6a1.png)

andreasn:
Also see https://www.openstack.org/

andreasn:
I think it should be OpenStack Image, since as far as I can see, that's what OpenStack itself calls it
![screenshot from 2018-11-14 16-38-52](https://user-images.githubusercontent.com/209895/48493308-ddf72700-e82b-11e8-9740-2f45070274a7.png)

I'm building an OpenStack image right now to check what format it outputs.

andreasn:
Here is the video I took that screenshot from, if you want to see the full flow on how to import a qcow2 image into OpenStack https://youtu.be/yTNKrJjODw4

andreasn:
Just finished building an openstack image, and the image is indeed **.qcow2**

PR #406, opened 2018-11-15
Capitalize OpenStack corrently in the image create dialog.

andreasn:
OpenStack didn't have a proper name in the dropdown, but took it
directly from the API. This commit adds an entry that uses the
corrent capitalization of the product.

Fixes issue #399


andreasn:
Screenshot of fix applied.

screenshot from 2018-11-15 15-00-38

PR #413, opened 2018-11-20
end to end test improvement

henrywang:
- [x] Page Object Pattern (page object and component)
- [x] Log for page element if it cannot be found
- [x] Random blueprint name and description
- [x] Random component added into blueprint
- [ ] Run test in parallel
- [x] More info (step log) in test report
- [x] ~~Visual regression test (screen capture comparison)~~
- [x] Performance improvement (~5 times faster than before)
- [x] New HTML report (no electron dependency)


henrywang:
@martinpitt could you please review this POC if you have time? Thanks.

Makefile
```diff
@@ -113,7 +113,7 @@ test_rpmbuild_cockpit-composer: buildrpm_image
 	sudo docker run --rm --name buildrpm -v `pwd`:/welder welder/buildrpm:latest make cockpit-composer-rpm cockpit-composer-srpm
 
 local-clean:
-	rm -rf test/images tmp cockpit-composer.spec cockpit-composer*.rpm welder-web*.tar.gz
+	rm -rf test/images tmp cockpit-composer.spec cockpit-composer*.rpm welder-web*.tar.gz test/end-to-end/node_modules test/end-to-end/wdio_report
```
martinpitt: I would suggest to not kill node_modules. This is very expensive to download, and just annoying if you have to run several tests in a row.

test/end-to-end/components/AvailableComponents.component.js
```diff
@@ -0,0 +1,37 @@
+// "Available Components" component in Edit Blueprint page
+class AvailableComponents {
```
martinpitt: So the idea is that components/ contains abstraction classes for the things the tests want to handle, so that the logic to enumerate table elements, click things, etc. is centralized?

My gut feeling is that this will make the tests easier to maintain once you understand them, but harder to debug (as you can't easily follow the code when reading the selenium debug log) and understand when you aren't familiar with them. But I don't have practical experience with this, so this is just thinking aloud. If you prefer this style, then by all means.

test/end-to-end/specs/blueprints.test.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+  });
+
+  describe('New Blueprint', function(){
+    let name, description, blueprintComponent;
+    beforeEach(function(){
+      // blueprint name cannot contain space due to issue https://github.com/weldr/welder-web/issues/317
+      // use lorem.slug() ("a-accusantium-repudiandae") instead of lorem.words() ("nulla placeat qui")
+      name = faker.lorem.slug();
+      description = faker.lorem.sentence();
```
martinpitt: TBH I'm not a big fan of such fancy stuff; just static strings are easier to control, easier to understand as a reader, and more robust for repeated debugging.

henrywang: @martinpitt, the blueprint name has to be random if we run test case in parallel. Each test will run on different blueprint.

Makefile
henrywang: @martinpitt, if we keep node_modules and copy it into "machine", after npm install, the wdio (webdriver.io) cannot be run. (that's a bug of webdriver.io). So I have to clear it and re-install all dependencies.
Before, I always removed node_modules manually, not convenience.

test/end-to-end/components/AvailableComponents.component.js
henrywang: @martinpitt. I understand what you mean. I was thinking this a lot, I tried both ways(put it into page file or separate file). No one is the best. The only reason I choose this way is this's a React app, the test should follow React way.

martinpitt:
[chrome failed](https://fedorapeople.org/groups/cockpit/logs/pull-413-20181126-054848-e8fbf168-cockpit-rhel-x-chrome/log), this could be a flake. But the [edge failure](https://fedorapeople.org/groups/cockpit/logs/pull-413-20181126-054747-e8fbf168-cockpit-fedora-29-edge/log) looks more serious:
```
F/root/end-to-end/node_modules/fibers/future.js:313
						throw(ex);
						^

Error: No such driver
    at new RuntimeError (/root/end-to-end/node_modules/webdriverio/build/lib/utils/ErrorHandler.js:143:12)
    at Request._callback (/root/end-to-end/node_modules/webdriverio/build/lib/utils/RequestHandler.js:316:39)
    at Request.self.callback (/root/end-to-end/node_modules/request/request.js:185:22)
    at Request.emit (events.js:182:13)
    at Request.EventEmitter.emit (domain.js:442:20)
    at Request.<anonymous> (/root/end-to-end/node_modules/request/request.js:1161:10)
    at Request.emit (events.js:182:13)
    at Request.EventEmitter.emit (domain.js:442:20)
    at IncomingMessage.<anonymous> (/root/end-to-end/node_modules/request/request.js:1083:12)
    at Object.onceWrapper (events.js:273:13)
```

Makefile
martinpitt: Eww, how ugly -- something writes into `node_modules/` during running the tests?? Could you please add a comment with a link to the bug report? This is outrageously annoying for local development.

I suggest a better workaround, like removing only the newly created file(s) which break stuff. Or, if it's hard to determine these reliably, copy test/end-to-end/node_modules to test/end-to-end/node_modules.clean after npm install, and copy back node_modules.clean to node_modules here.

martinpitt:
Sure, you can use `cp -a` to preserve symlinks. I'm just concerned because `npm install` takes several minutes, it compiles stuff, etc.

henrywang:
Thanks @martinpitt. end-to-end/node_modules was removed from `local-clean` of make file.

henrywang:
@martinpitt, could you please review this PR again if you have time? Thanks.
It's one time all green!
This PR doubled code coverage to 66%.

PR #459, opened 2018-11-26
Add AppStream metainfo

martinpitt:
This will make welder-web appear on Cockpit's "Applications" page.

### DIFF
--- a/io.weldr.cockpit-composer.metainfo.xml
+++ b/io.weldr.cockpit-composer.metainfo.xml
@@ -1,0 +1,16 @@
+<component type="addon">
+  <id>io.weldr.cockpit-composer</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Image Builder</name>
+  <summary>
+      Build customized operating system images
+  </summary>
+  <description>
+    <p>
+      Image Builder (Composer) can generate custom images suitable for deploying
+      systems, or as images ready to upload to the cloud.
+    </p>
+  </description>
+  <extends>cockpit.desktop</extends>
+  <launchable type="cockpit-manifest">welder</launchable>
+</component>
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,13 @@ npm-install:
 install: all
 	mkdir -p /usr/share/cockpit/welder
 	cp -r public/* /usr/share/cockpit/welder
+	mkdir -p /usr/share/metainfo/
+	cp io.weldr.cockpit-composer.metainfo.xml /usr/share/metainfo/
 
 dist-gzip: NODE_ENV=production
 dist-gzip: all $(PACKAGE_NAME).spec
 	mkdir -p $(PACKAGE_NAME)-$(VERSION)
-	cp -r public/ LICENSE.txt README.md $(PACKAGE_NAME).spec $(PACKAGE_NAME)-$(VERSION)
+	cp -r public/ LICENSE.txt README.md $(PACKAGE_NAME).spec io.weldr.cockpit-composer.metainfo.xml $(PACKAGE_NAME)-$(VERSION)
 	tar -czf $(PACKAGE_NAME)-$(VERSION).tar.gz $(PACKAGE_NAME)-$(VERSION)
 	rm -rf $(PACKAGE_NAME)-$(VERSION)
 
--- a/cockpit-composer.spec.in
+++ b/cockpit-composer.spec.in
@@ -24,8 +24,11 @@ Composer GUI for Cockpit and lorax-composer
 %install
 mkdir -p %{buildroot}/%{_datadir}/cockpit/welder
 cp -r public/* %{buildroot}/%{_datadir}/cockpit/welder
+mkdir -p %{buildroot}/%{_datadir}/metainfo/
+cp io.weldr.cockpit-composer.metainfo.xml %{buildroot}/%{_datadir}/metainfo/
 
 %files
 %doc README.md
 %license LICENSE.txt
 %{_datadir}/cockpit/welder
+%{_datadir}/metainfo/*


martinpitt:
As this once again has to duplicate stuff -- has there been a decision whether to use just one name (welder-web vs. cockpit-composer) at some point?

bcl:
> As this once again has to duplicate stuff -- has there been a decision whether to use just one name (welder-web vs. cockpit-composer) at some point?

Yes. IIRC we agreed to just name it all cockpit-composer a while back. I was planning on letting you guys handle making the switch, I don't think it makes any sense for me to continue doing upstream builds -- but I'll still backport them to rhelX.

andreasn:
Tested and looks good to me!
Didn't we have something a while back that it could also show a link to the service file?
At least I recall we did in a really early prototype by @mvollmer
Would be nice adding a link to lorax-composer.service (or socket?) if we do.

andreasn:
So as a naming thing. This install thing says "Composer", but the tab in the browser that it installs says "Image Builder". @jgiardino do you recall what we decided here?

io.weldr.cockpit-composer.metainfo.xml
```diff
@@ -0,0 +1,15 @@
+<component type="addon">
+  <id>io.weldr.welder-web.metainfo.xml</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Composer</name>
+  <summary>
+      Web interface for Composer
+  </summary>
+  <description>
+    <p>
+      Composer can generate custom images suitable for deploying systems, or as images ready to upload the cloud.
```
mvollmer: "to the cloud"

io.weldr.cockpit-composer.metainfo.xml
```diff
@@ -0,0 +1,15 @@
+<component type="addon">
+  <id>io.weldr.welder-web.metainfo.xml</id>
```
mvollmer: ` io.weldr.cockpit-composer`?

io.weldr.welder-web.metainfo.xml
```diff
@@ -0,0 +1,15 @@
+<component type="addon">
+  <id>io.weldr.welder-web.metainfo.xml</id>
```
mvollmer: `io.weldr.welder-web`

mvollmer:
> Didn't we have something a while back that it could also show a link to the service file?

Yes, but I don't think it would apply here.  The idea back then was that Cockpit could install something that _only_ has a service file but no UI to go with it.  Cockpit would invent a UI on the spot, which could just be a link to the page for the service.

In this case, we install a UI and if it makes sense, that UI can contain a link to the pages for the services it needs.

io.weldr.cockpit-composer.metainfo.xml
mvollmer: But I am confused about the need to duplicate things, so maybe `welder-web` as the id is correct.

martinpitt: @mvollmer: It's still the upstream github project name, and https://weldr.io its upstream homepage. Is the ID visible to users anywhere? I thought it was just an internal disambiguation for the API?

io.weldr.welder-web.metainfo.xml
martinpitt: That's exactly what is here, or I misunderstand you.

Note that welder-web currently can build a welder-web.rpm (with this file) or a cockpit-composer.rpm (with the other one). This is also why I asked whether we can drop one name now. But for now this PR supports both.

andreasn:
> In this case, we install a UI and if it makes sense, that UI can contain a link to the pages for the services it needs.

Ah, right. Of course. I forgot we check if lorax-composer is running on the Composer main page and offer to start it there. That should make it a complete flow then. Thanks!

io.weldr.welder-web.metainfo.xml
mvollmer: I would drop the `metainfo.xml` part, but I can't say I am an export on AppStream ids.

io.weldr.cockpit-composer.metainfo.xml
larskarlitski: We do want to change the name to cockpit-composer everywhere. I haven't done so yet, because I'm afraid I'll miss and break something. I think we should go with cockpit-composer here already. The change is coming in the near future.

mvollmer: Could we have just a single file instead of two files with identical content?  That would be less confusing, I think.

martinpitt: It's not identical content, the `<Name>` is different. And yes, the duplication is confusing, but as long as we have it, we should be consistently doing it. We can also block this PR on doing the un-duplication, if that happens soon (or I need to carve out some time to do it)

martinpitt:
Pushed again with dropping the welder-web appstream stuff, only keeping it for cockpit-composer. As we don't publish welder-web.rpm anywhere except a COPR, it doesn't matter to not add it there. The metadata only really makes sense when the package is in Fedora or RHEL 8.

io.weldr.cockpit-composer.metainfo.xml
```diff
@@ -0,0 +1,15 @@
+<component type="addon">
+  <id>io.weldr.cockpit-composer</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Composer</name>
+  <summary>
+      Web interface for Composer
```
larskarlitski: Everything in Cockpit is a Web interface :wink: 

I prefer something along the lines of "Build customized operating system images"

jgiardino:
> This install thing says "Composer", but the tab in the browser that it installs says "Image Builder". @jgiardino do you recall what we decided here?

The decision was to follow the convention we currently have in Cockpit where the package name and CLI command name are specific, but the label for the tool in the UI is generic and descriptive of the functionality, such as with [Virtual Machines](https://cockpit-project.org/guide/latest/feature-virtualmachines.html).

PR #464, opened 2018-12-05
Disable https in cockpit and use http instead to ignore self-signed cert warning.

henrywang:


martinpitt:
On the face of it it looks good, but several tests now fail with "Error: Navigation caused the user agent to hit a certificate warning, which is usually the result of an expired or invalid TLS certificate.".

henrywang:
@martinpitt, oh, i just forgot change https to http in `test/check-application`. Updated and one time all green again.

PR #469, opened 2018-12-12
Add attribute `data-*` in some elements, update case accordingly, add new case

henrywang:
Add attribute `data-*` in some elements to make element locating easy.
Update test case according to new changes.
Add new case to cover "Selected Components" tab and "Dependencies" tab.

### DIFF
--- a/components/ListView/ComponentInputs.js
+++ b/components/ListView/ComponentInputs.js
@@ -97,7 +97,7 @@ class ComponentInputs extends React.Component {
     return (
       <div className="list-pf cmpsr-list-inputs cmpsr-list-pf__compacted list-pf-stacked">
         {components.map((component, i) => (
-          <div key={i} className={`list-pf-item ${component.active ? 'active' : ''}`}>
+          <div key={i} className={`list-pf-item ${component.active ? 'active' : ''}`} data-input={component.name}>
             <div
               className="list-pf-container"
               tabIndex="0"
--- a/components/ListView/LabelWithBadge.js
+++ b/components/ListView/LabelWithBadge.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 export default function LabelWithBadge(props) {
   return (
-    <span>{props.title}&nbsp;<span className="badge">{props.badge}</span></span>
+    <span>{props.title}&nbsp;<span className="badge" data-badge={props.title}>{props.badge}</span></span>
   );
 }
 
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -114,6 +114,9 @@ test_rpmbuild: buildrpm_image
 test_rpmbuild_cockpit-composer: buildrpm_image
 	sudo docker run --rm --name buildrpm -v `pwd`:/welder welder/buildrpm:latest make cockpit-composer-rpm cockpit-composer-srpm
 
+debug_install:
+	sudo rm -rf /usr/share/cockpit/welder && sudo cp -r `pwd`/public /usr/share/cockpit/welder
```
martinpitt: IMHO this is wrong, as it scribbles over package manager territory. It's much better to create a ~/.local/share/cockpit/welder symlink to `$(pwd)/public`. Maybe call it "local-install"?

Also, this new target needs to be added to `.PHONY`.

components/ListView/ComponentInputs.js
```diff
@@ -1,16 +1,16 @@
...
 import React from 'react';
-import {defineMessages, injectIntl, intlShape} from 'react-intl';
+import { defineMessages, injectIntl, intlShape } from 'react-intl';
 import PropTypes from 'prop-types';
 import ComponentTypeIcons from '../../components/ListView/ComponentTypeIcons';
 
 const messages = defineMessages({
   hideDetails: {
-    defaultMessage: "Hide Details"
+    defaultMessage: 'Hide Details'
```
martinpitt: Note that translatable strings must be marked with `_("...")`, and for xgettext it *has* to be double quotes. This PR doesn't make things any worse, but this goes into the wrong direction.

components/ListView/ComponentInputs.js
```diff
@@ -1,16 +1,16 @@
 /* global $ */
 
 import React from 'react';
-import {defineMessages, injectIntl, intlShape} from 'react-intl';
+import { defineMessages, injectIntl, intlShape } from 'react-intl';
```
martinpitt: There are a lot of these formatting changes in this PR. Would be nice if they could be done in a separate commit, as this makes it hard to review what actually changed.

henrywang:
@martinpitt, I deleted makefile target change and will move it to another PR. All of the formatting changes are removed from this PR.
The only failed case is due to [issue 471](https://github.com/weldr/welder-web/issues/471).

henrywang:
@martinpitt, yes. It's in another PR.

PR #470, opened 2018-12-13
Make cockpit-composer the only package name

martinpitt:
Drop the duality of welder-web vs. cockpit-composer and only keep the
latter name.

"welder-web" is kept for the GitHub and Zanata projects, these renames
need to happen separately.


martinpitt:
@bcl @jgiardino : After that I'd like to tag a new upstream release, to get a COPR build. It doesn't necessarily have to be uploaded to RHEL, but it should be the thing to review for Fedora inclusion. Any objections?

PR #472, opened 2018-12-14
Update eslint, fix eslint issue, format code in some of files.

henrywang:
1. Update eslint and its plugins.
2. Fix eslint issue on EmpytState.js, Layout.js, Link.js, and run-zanata.js.
2. Disable some eslint rules in this PR. These issues will be fixed in separated PRs.
4. Update end to end test case according to code change.


components/EmptyState/EmptyState.js
```diff
@@ -1,27 +1,27 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-
+import React from "react"
+import PropTypes from "prop-types"
 
 class EmptyState extends React.PureComponent {
-
   render() {
+    const { title, message, children } = this.props
```
martinpitt: This PR removes a lot of semicolons on line endings. At least personally I must say this is highly confusing, as in JS in general, statements have to be terminated by a semicolon. Does eslint enforce that with this configuration? [cockpit's](https://github.com/cockpit-project/cockpit/blob/master/.eslintrc.json) doesn't, and we don't have anything special there.

This should maybe get agreed upon between the welder-web developers  first, as it's a lot of noise.

henrywang:
Thanks @martinpitt.
There's no any eslint rule to enforce semicolon removing. I do not have strong reason to remove it. So let's discuss it here. Thanks.

components/EmptyState/EmptyState.js
jgiardino: I'm not familiar with what's required vs what's standard. I'd defer to whatever the cockpit devs prefer or would recommend. 

That said, seeing it like this is not what I'm use to. If there are some cases where semicolons are required and others where they're optional, it might take me some time to fully grasp those rules.

larskarlitski:
> [...] So let's discuss it here. If cockpit does not remove semicolon, composer should align with cockpit's style. Thanks.

Yeah I very much agree with Martin here. Please let's be as close to [cockpit's style](https://github.com/cockpit-project/cockpit/blob/master/.eslintrc.json) as possible.

Otherwise is will be confusing for people who work on both codebases (like me :smile:).

package.json
```diff
@@ -48,14 +49,14 @@
     "ejs": "^2.5.1",
     "enzyme": "^3.5.0",
     "enzyme-adapter-react-15": "^1.1.0",
-    "eslint": "^3.17.1",
-    "eslint-config-airbnb": "^9.0.1",
-    "eslint-plugin-import": "^1.12.0",
-    "eslint-plugin-jest": "^20.0.3",
-    "eslint-plugin-jsx-a11y": "^2.2.3",
+    "eslint": "^5.10.0",
```
KKoukiou: Dependency updates are not mentioned in the commit message.

henrywang: Thanks @KKoukiou. The eslint update is due to `eslint-plugin-jsx-a11y` depends on it. And due to this dependence, `eslint-plugin-jsx-a11y` does not work for a long time. So I fix it in this PR.

KKoukiou: Ah right :) thanks!

components/EmptyState/EmptyState.js
```diff
@@ -1,27 +1,27 @@
...
+import React from "react";
+import PropTypes from "prop-types";
 
 class EmptyState extends React.PureComponent {
-
   render() {
+    const { title, message, children } = this.props;
     return (
       <div className="blank-slate-pf">
-        {this.props.icon}
```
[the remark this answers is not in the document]
henrywang: This code change is to fix "react/destructuring-assignment" issue.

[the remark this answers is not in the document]
KKoukiou: this.props.icon was used before but not any more. The rest of variables are fixing the rule you mention indeed.

henrywang:
@KKoukiou, There're 4 commits in this PR. I would like to make each commit for different PR, but the following 3 commits depends on the first one. And I plan to continue working on it to keep fixing eslint issue in this PR but with different commits during the holiday. And I'll make them separate when all you guys come back and can review my PR.

package.json
henrywang: @KKoukiou, happy holidays!

components/EmptyState/EmptyState.js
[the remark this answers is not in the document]
henrywang: @KKoukiou, ah, sorry for confusing here. The reason I remove `{this.props.icon}` is this prop is not used by all of renders in this project. And I'll add this change into commit. Thanks!

package.json
```diff
@@ -22,7 +22,8 @@
     "redux": "3.6.0",
     "redux-saga": "0.15.3",
     "reselect": "3.0.1",
-    "whatwg-fetch": "1.1.1"
+    "whatwg-fetch": "1.1.1",
+    "bootstrap": "^3.3.7"
```
martinpitt: Moving this from devDependencies to dependencies is correct. It should not be hidden in this commit, though.

components/ListView/ComponentDetailsView.js
```diff
@@ -75,16 +74,17 @@ class ComponentDetailsView extends React.Component {
         this.setState({ componentData: data[0][0] });
         const filteredDependencies = data[0][0].dependencies.filter(dependency => dependency.name !== data[0][0].name);
         this.setState({ dependencies: filteredDependencies });
-        if (this.props.status === "editSelected") {
+        const { status } = this.props;
+        if (status === "editSelected") {
```
martinpitt: This looks wrong. `status` is a method parameter, and unrelated to `this.props.status`. Also, why all this excessive aliasing like `const { status } = this.props;` ? This is just noise, makes the code harder to follow, and is prone to namespace conflicts like this one.

components/ListView/ListItemImages.js
```diff
@@ -21,30 +21,34 @@ class ListItemImages extends React.Component {
 
   // maps to Remove button for FAILED
   handleDelete() {
-    this.props.deletingCompose(this.props.listItem.id);
+    const { deletingCompose, listItem } = this.props;
+    deletingCompose(listItem.id);
```
martinpitt: This is another such case. The extra aliasing adds no benefit at all here.

martinpitt: Right, but I can't say that it really makes sense here. The [intention](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md) is for render methods that directly map property names to the rendered results, not for these helper methods. If it forces us to add such irrelevant and even dangerous noise, then I'd rather vote for disabling that eslint error.

henrywang: Agree with you. I'll disable this eslint rule and keep most of props at destructuring-assignment and make some of props back to `this.props` style.

martinpitt:
@henrywang Please no merge commits (they also don't work with github). Rebase your's on top of master instead.

henrywang:
@martinpitt, the failed case in rhel-7-6 is imported by [PR 469](https://github.com/weldr/welder-web/pull/469). I got the root cause and will fix it in next PR.

martinpitt:
@henrywang : Sorry, I missed the failure as at that time we still had the broken rhel-7-6 image, so it got hidden by that.

henrywang:
@martinpitt, I removed " fix "react/destructuring-assignment" eslint error" commit. Thanks for your suggestions! :+1:
BTW: I'll fix rhel-7-6 test issue in a coming PR.

PR #481, opened 2018-12-25
No gcc-c++ as denpendence because we don't need electron any more.

henrywang:


henrywang:
@martinpitt, is there any change in rhel-7-6 image? Thanks.

PR #494, opened 2019-01-10
Remove useless files.

henrywang:
1. files in docs folder are not related with composer.
2. icons folder and some files in public folder are non-cockpit.
3. database.rules.json, firebase.json, import-metadata.sh and markdown-loader.js are not used by composer.


import-metadata.sh
```diff
@@ -1,43 +0,0 @@
-#!/bin/bash
-
```
martinpitt: I'm not sure about this one. It's not necessary with lorax-composer, but the original plan was that lorax-composer is just an implementation backend for RHEL 7, and bdcs is the intended backend for the future. Did that change? AFAIK, bdcs still needed a metadata.db (although that is certainly a bug, too).

henrywang:
Thanks @martinpitt. I'll land it when tests are all green.

PR #495, opened 2019-01-10
Remove unit test because the cases have been covered by end to end.

henrywang:
1. remove unit folder.
2. remove unit test in Travis.
3. remove unit test related dependencies.
4. remove jest related eslint configurations.
5. add "expect": true in "globals" of eslint configuration because "jest/valid-expect": "off" has been removed.


henrywang:
Thanks @martinpitt. Yeah! And no code coverage decrease.

PR #497, opened 2019-01-11
Move raw code into src folder

henrywang:
1. Move components, core, data, pages folders and main.js, routes.json files into src folder.
2. Update package.js, run.js and webpack.config.js according to location change.


henrywang:
@martinpitt, agree, it'll add more effort when typing file names. I put the main files in a /src/ file because it is neater and better than putting everything in the root. And It'll be easy to configure babel and webpack.

PR #500, opened 2019-01-16
Refactor api calls and data structure

jgiardino:
This pulls in updates made in #345 and includes additional updates.

This PR primarily focuses on trying to move functionality from BlueprintApi and MetadataApi files to redux saga.

Some of the related UI updates include:

- visual indication when a blueprint component is a dependency (see abrt-dbus in example below)
- allowing dependencies to be explicitly added (and also not displaying a remove action)
- keeping Undo/Redo history after discarding workspace changes (e.g. the user can discard changes, but still redo to get them back)
- handling component version updates when more than one version is available

image


jgiardino:
There are a few weird things that I've noticed and am either not sure how to best handle them, or want to address them in a separate PR:

- On the Edit Blueprint page, if a list item in the Blueprint Components panel is expanded, and the list gets updated, the contents that display in the expanded section are fetched again. I’m not sure what the best way to handle this is (i.e. keeping that data and avoid making that api call). Whenever blueprint contents are updated, I reload the contents (i.e. depsolve the blueprint using the updated list of selected components)
- When viewing the list of dependencies in the Component Details view, the metadata we get with the list of dependencies doesn’t include summary. I don’t perform a second api call just to get summary when loading the list, but I do make this call per list item when it’s expanded. It’s weird that summary suddenly pops into the list item row. I’m not sure what the best approach to this is (i.e. modify the design to accommodate this data, or modify the api, or just make the second api call to get a bunch of extra data that we may never need).
- I made some changes to the display of info in the Pending Changes modal to only show name and version, since this is the info that gets posted, but this could probably use additional refinement in a separate PR.
- I haven’t updated the UI that displays when there are multiple options available, since this could be a separate PR. Currently, in this case, a select menu displays with version-release as the menu options. When there are multiple sources available (e.g. Fedora 28 and Fedora 29) and the same version is available in more than one release, I see the older release as an option, and can currently select it, but the blueprint will immediately commit the contents to the workspace, which only includes the package name and version (and not release), and then reload showing the release provided by the api, which is always the latest (I think). I definitely want to update the UI to move the select menus to the Details section, but I’m not sure what the best/ultimate solution for handling many available sources should be.

components/Modal/PendingChanges.js
```diff
@@ -211,44 +210,74 @@ class PendingChanges extends React.Component {
...
-                                      {Object.values(componentUpdated.new)[0].name}-
-                                      {Object.values(componentUpdated.new)[0].version}
-                                    </strong>
-                                  </li>
-                                ))}
-                              </ul>
+                      {this.props.blueprint.workspacePendingChanges.map(change => (
+                        <li
+                          className="list-group-item"
+                          key={!change.componentNew ? change.componentOld : change.componentNew}
```
henrywang: Object inside `workspacePendingChanges` array has "name" and "version" keys. 
But object inside `localPendingChanges`. array does not. It would be better to have same object structure.
componentnew

Should be changed to
```
key={
   !change.componentNew
   ? `${change.componentOld.name}-${change.componentOld.version}`
   : `${change.componentNew.name}-${change.componentNew.version}`
}
```

core/reducers/blueprints.js
```diff
@@ -250,10 +221,34 @@ const blueprints = (state = [], action) => {
               return Object.assign({}, blueprint, {
                 present: Object.assign({}, blueprint.past.shift(), {
                   localPendingChanges: [],
-                  workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
+                  workspacePendingChanges: []
```
henrywang: The same change needs to be applied to all `webspacePendingChanges`. Otherwise this prop will get old value and `this.props.blueprint.workspacePendingChanges.map` in `PendingChanges.js` will throw an error and pending changes cannot get rendered in Pending Changes page.
workspacependingchanges
```shell
diff --git i/core/reducers/blueprints.js w/core/reducers/blueprints.js
index 5a4521d4..698ace86 100644
--- i/core/reducers/blueprints.js
+++ w/core/reducers/blueprints.js
@@ -27,7 +27,7 @@ const blueprints = (state = [], action) => {
             past: [],
             present: Object.assign({}, action.payload.blueprint, {
               localPendingChanges: [],
-              workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
+              workspacePendingChanges: []
             }),
             future: []
           }
@@ -50,7 +50,7 @@ const blueprints = (state = [], action) => {
                 past: [],
                 present: Object.assign({}, action.payload.blueprint, {
                   localPendingChanges: [],
-                  workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
+                  workspacePendingChanges: []
                 }),
                 future: []
               }
@@ -131,7 +131,7 @@ const blueprints = (state = [], action) => {
                 past: [],
                 present: Object.assign({}, action.payload.blueprint, {
                   localPendingChanges: [],
-                  workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
+                  workspacePendingChanges: []
                 }),
                 future: []
               });
```

henrywang:
@jgiardino, All selected component were gone when apply a version-release change.
Reproduce steps:
1. add 0ad and 0ad-data to blueprint.
2. commit
3. change 0ad version-realese to a new one.
4. commit
5. 0ad and 0ad-data disappeared from Blueprint Component. And you it always keep blank whether you add a new one or remove old or not.
If you want selected component back, you have to refresh the whole page.

components/Modal/PendingChanges.js
jgiardino: Yes, this is definitely a change I'd like to make. I'd also like to separate the name and version as you mention below. I was thinking these changes could be a separate PR, since this is already pretty big.

core/reducers/blueprints.js
jgiardino: Thanks for catching that!

jgiardino:
@henrywang Thanks for reviewing! I addressed a couple of things you commented on:

- missing updates for workspacePendingChanges prop change
- handling depsolve error
  - For this update, I moved the blueprint errorState into the blueprint history. This allows the user to hit Undo to return to the previous state of the blueprint before they made a change that caused this error. This also allows them to hit Redo, which would show the error message again, which might be odd, but also might be helpful for some people. There might be some additional UI refinements needed here, like not showing Add/Remove actions in the list of inputs. Let me know if you have any suggestions.

Thank you for your suggestions on the other items that I mentioned in my comment. I'd like to get further feedback on some of these, and I'm also not sure that any of these updates need to happen as part of this PR, or not. Unless anyone sees them as critical, I'd like to break them out as separate PRs.

For the failing tests, I'm not sure what needs to happen to get those to pass. Do you have any suggestions?

martinpitt:
Sorry for the botched test tasks (long story). You need at least one more force-push to get the tests running again.

henrywang:
@jgiardino, could you please make a force-push to make test running? Thanks!

jgiardino:
The fedora tests still fail, but it isn't obvious to me what changes are needed so they can pass. Any suggestions would be helpful. 😄

core/sagas/blueprints.js
```diff
@@ -55,40 +55,129 @@ function* fetchBlueprints() {
...
-        deletedChanges: deletedChanges
-      };
-      blueprintPast = [
-        Object.assign({}, blueprintResponse, {
-          localPendingChanges: [],
-          workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
-        })
-      ];
+    const response = yield call(fetchBlueprintContentsApi, blueprintId);
+    console.log("blueprintData", response);
```
henrywang: Could you please remove this debug output? Thanks.

jgiardino:
@henrywang Thanks for providing info about what causes the tests to fail. I'm not sure that your solution is possible. This issue with the package isn't surfaced until `get_blueprints_deps: "/api/v0/blueprints/depsolve/",` is called, which requires first adding the package to the blueprint. I'm not aware of any way to get this level of info about each available package that's listed before the user adds the package. 

@bcl  Can you comment on this issue that @henrywang noted above? After I use /api/v0/modules/list to get a list of available packages, is there any other api call I can make that would let me know if any packages in that list would return an error when added to the blueprint?

If this info isn't available after getting the list of packages, and we have to add the package to the blueprint so that we can depsolve, would the test still fail if we add the package, catch the error so that we can display it in the UI, then automatically remove the package?

jgiardino:
Per chat with @larskarlitski and @henrywang, we will run x number of tests, and if more than n% errors are returned then the test should fail.

Additionally, ideally we would test when the blueprint correctly depsolves and correctly returns a depsolve error by having a local repo with a controlled set of packages that are known to either depsolve or not.

jgiardino:
Also, additional UI refinements related to how errors are handled can be added in a separate PR.

henrywang:
@jgiardino, test got updated in PR #513. Just rebase on top of master when this #513 got merged. Thanks.

components/ListView/BlueprintContents.js
```diff
@@ -41,19 +41,28 @@ const BlueprintContents = props => {
...
     <div>
       {(fetchingState === true && <Loading />) ||
-        ((Object.keys(errorState).length && (
+        ((Object.keys(errorState).length > 0 && (
           <EmptyState
             title={formatMessage(messages.emptyStateErrorTitle)}
             message={formatMessage(messages.emptyStateErrorMessage)}
-          />
+          >
+            <p>{errorState.msg}</p>
```
henrywang: The `EmpytState` component has children prop with an React element type. But here's an array, not an element. Here're two solutions:
1. put `<p>` and `<button>` into a `<div>`
2. update children type to:
```javascript
children: PropTypes.oneOfType([
  PropTypes.element,
  PropTypes.arrayOf(PropTypes.element)
])
```
empty_state

henrywang:
According to https://github.com/weldr/welder-web/pull/500#discussion_r248183899, one more left.
```
diff --git i/core/reducers/blueprints.js w/core/reducers/blueprints.js
index f052754d..c30deded 100644
--- i/core/reducers/blueprints.js
+++ w/core/reducers/blueprints.js
@@ -50,7 +50,7 @@ const blueprints = (state = [], action) => {
                 past: [],
                 present: Object.assign({}, action.payload.blueprint, {
                   localPendingChanges: [],
-                  workspacePendingChanges: { addedChanges: [], deletedChanges: [] }
+                  workspacePendingChanges: []
                 }),
                 future: []
               }
```

components/ListView/BlueprintContents.js
jgiardino: Thanks for the suggestion! I made this update.

PR #503, opened 2019-01-18
Add generating translation.json before building.

henrywang:


main.js
```diff
@@ -39,11 +39,13 @@ if (userLanguage in translations) {
   messages = translations[userLanguage];
 }
 
+const locale_lang = userLanguage.includes("_") ? userLanguage.split("_")[0] : userLanguage;
```
martinpitt: This seems a bit too simplistic. "zh" by itself isn't a precise enough specification to e. g. tell apart simplified from traditional Chinese, and neither is e. g. "pt" to tell apart Portugal from Brazilian Portugese.

Can you please check if changing the _ to a - works? AFAIK, usually locales are written like "pt-br" or "pt-BR" in the JavaScript  world.

PR #509, opened 2019-01-22
Update babel and webpack and change settings according to update.

henrywang:
1. Update babel to 7 and webpack to 4 and add babel.config.js
because babel 7 does not look setting in package.json.
2. Remove run.js to make webpack to take care of it.
3. Remove package.json inside components/Layout and Link folders
and update "import" in BlueprintListView, index.js in all page
folders.
4. Remove po/po.empty.js, po2json. Thay're useless right now.
5. Put all final files into public/dist and update spec file.
6. Update index.ejs, so jquery is not sourced here, but by webpack.


main.js
```diff
@@ -8,42 +7,41 @@ import enLocaleData from "react-intl/locale-data/en";
 import FastClick from "fastclick";
 import { Provider } from "react-redux";
 import "bootstrap";
+import cockpit from "cockpit"; // eslint-disable-line global-require, import/no-unresolved
```
martinpitt: For a follow-up cleanup: To avoid the eslint override, you can put a `module.exports.externals = { "cockpit": "cockpit" }`

https://github.com/cockpit-project/cockpit/blob/master/webpack.config.js#L249

But actually our webpack config already has that, so it's strange that eslint complains about that. starter-kit doesn't use these hacks.

package.json
```diff
@@ -7,7 +7,7 @@
     "npm": ">=5.0"
   },
   "dependencies": {
-    "babel-polyfill": "6.26.0",
+    "@babel/polyfill": "^7.2.5",
```
martinpitt: Please don't use the `^` syntax in `dependencies`, only in `devDependencies`. Our bot will continuously try to update npm modules in package.json, but run them through PRs with CI, which is a lot more reliable.

package.json
```diff
@@ -16,7 +16,7 @@
     "patternfly": "3.59.0",
     "patternfly-react": "1.19.1",
     "prop-types": "^15.6.2",
-    "react": "15.6.2",
+    "react": "^15.6.2",
```
martinpitt: Here too

henrywang:
@martinpitt, fixed eslint error. :-)

PR #517, opened 2019-01-28
WIP: bos-tasks-x5wh2: Update translations from Fedora Zanata

cockpituous:
 * [ ] po-refresh

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -1,4 +1,5 @@
 # Ludek Janda <ljanda@redhat.com>, 2018. #zanata
+# Xiaofeng Wang <xiaofwan@redhat.com>, 2019. #zanata
 #, fuzzy
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


martinpitt:
Look @henrywang , your translation updates are coming in! :-)

martinpitt:
Got [timeouts/test failures](https://fedorapeople.org/groups/cockpit/logs/pull-517-20190128-080000-17da46fa-cockpit-fedora-29-chrome/log) on all Fedora 29 tests, retrying. If that doesn't help, likely some external change crept in, as I don't suppose we actually test the UI in Chinese?

henrywang:
@martinpitt, went through the screenshot, the failure is caused by "create image". Sometimes the UI does not get updated after clicking "create image" in 120 seconds (test timeout). Actually UI should be updated with "In Progress" and Stop button rendered.
Right there's no Chinese test case for UI.

PR #520, opened 2019-01-29
Move coverage collector to afterSuite hook from afterTest hook.

henrywang:
Because the test code in "after" hook cannot get covered if we collect coverage result in "afterTest" hook.


martinpitt:
I wonder if this actually works.. There is no coverage report here?

henrywang:
@martinpitt, thanks for review.
I have no idea why there's no coverage report in recent PR (not only this one). But it works.
You can find coverage result from https://codecov.io/gh/weldr/welder-web/commits.

PR #530, opened 2019-02-07
Enables user to specify wildcard version value

jgiardino:
Originally, we had discussed adding the easiest fix for this issue, which was assumed to be always saving the version as "*". But after considering the following:

- Updates in #500 included the ability to select one out of a list of all available versions that are returned  
- The effort involved in removing all of that seemed more complicated than just adding on the ability to select a wildcard version
- In the cli, the user can specify a specific version or any variation of a wildcard version (e.g. 1.2.3, 1.2.*, 1.*, *)

It seemed like the easiest path was to just update the list of available versions to include wildcard variations of the versions.

Updates included in this PR:

- By default, the wildcard version "*" is selected/added to a blueprint. 
  - Previously, clicking the + icon in the list of available components would add the latest version. 
- The user can optionally view all versions that are available and choose a specific version or a
wildcard variation of a specific version (e.g. "10.*"). (see image below)
  - Previously, when viewing component details, the select menu options were version-release. 

image

### DIFF
--- a/core/sagas/inputs.js
+++ b/core/sagas/inputs.js
@@ -64,6 +64,8 @@ function* fetchInputs(action) {
 function flattenInput(response) {
   // each response item is a different build (version, release, arch)
   // flatten the response to a single item with an array of builds
+  // only keep the latest release per version
+  // for each version, include wildcard options
   let previousBuilds = {};
   let flattened = Object.assign({}, response[0], { builds: [] });
   response.forEach(item => {
@@ -73,21 +75,45 @@ function flattenInput(response) {
       release: item.builds[0].release,
       arch: [item.builds[0].arch]
     };
-    if (previousBuilds.hasOwnProperty(build.version + build.release)) {
-      // if this item has the same values for release and version of a
-      // previousBuild
-      // then push arch to the array of arch's, and filter this item out
-      previousBuild = previousBuilds[build.version + build.release];
+    if (previousBuilds.hasOwnProperty(build.version)) {
+      // if this item has the same version value as a
+      // previousBuild replace the release value (this is assumed to be the latest)
+      // then push arch to the array of arch's and filter this item out
+      previousBuild = previousBuilds[build.version];
       previousBuild.arch = previousBuild.arch.concat(build.arch);
+      previousBuild.release = build.release;
       return false;
     } else {
       // else push the build to the array of builds
       flattened.builds = [build].concat(flattened.builds);
     }
-    previousBuilds[build.version + build.release] = build;
+    previousBuilds[build.version] = build;
     return true;
   });
+  flattened.builds = addWildcardVersions(flattened.builds);
   return flattened;
+}
+
+function addWildcardVersions(builds) {
+  let wildcardBuilds = {};
+  builds.forEach(item => {
+    item.depsolveVersion = item.version;
+
+    let parts = item.version.split(".");
+    for (let i = 0; i < parts.length; i += 1) {
+      let wildcard = parts
+        .slice(0, i)
+        .concat("*")
+        .join(".");
+
+      if (!(wildcard in wildcardBuilds)) {
+        wildcardBuilds[wildcard] = Object.assign({}, item, { version: wildcard });
+      }
+    }
+    wildcardBuilds[item.version] = Object.assign({}, item);
+  });
+
+  return Object.values(wildcardBuilds);
 }
 
 // when ComponentDetailsView loads, get component details
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -34,7 +34,8 @@ class ComponentDetailsView extends React.Component {
   constructor() {
     super();
     this.state = {
-      selectedBuildIndex: undefined
+      selectedBuildIndex: undefined,
+      savedVersion: undefined
     };
     this.setBuildIndex = this.setBuildIndex.bind(this);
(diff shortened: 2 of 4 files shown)


jgiardino:
There are a couple of things that are a little strange in the UI after these updates. I'm not sure that these are critical to include in this PR, though.

- When listing the selected components for a blueprint, the version is presented as the depsolved version, and not the one saved with the blueprint. This makes it difficult to know what the actual saved value is.
- Originally, the plan was to remove the select menu from the top of the Component Details panel, and move it down to be adjacent to the Version label listed with the other details. But I also like showing what the current depsolved value for version would be based on the selected value if it is a wildcard value, which is why I kept it there for now. Probably the same solution we provide for the previous bullet would work here.
- In the list of available components, the tooltip for the Add button shows what the depsolved version of the default "*" version value would be. This could be confusing, and maybe just having it say "Add Component" or "Add Latest Version" would be better.

Related to the last bullet above, should the first menu option include "latest version", like this, or is that obvious?:
  ```
* (latest version)
10.*
10.15.*
...
```

**Comment by andreasn** [comment]
The star is not obvious to me as "wildcard", so the _latest version_ string sounds like a very good idea to me.

**Comment by atodorov** [comment]
@jgiardino, @henrywang can we get a UI test for this option (since we have a bug for it) ? 

How does selecting a wildcard version (*) affect the export output (e.g. do I get glibc-*-.rpm or not when I export the depsolved contents of the recipe)? Not sure if relevant though.

**Comment by atodorov** [comment]
@jgiardino do we progressively replace version numbers to the left, e.g. what if a package has 1.2.3.4 version numbering does this produce *, 1.*, 1.2.*, 1.2.3.* ? 

```
* (latest version)
10.*
10.15.*
```

**Comment by jgiardino** [comment]
I updated the option "*" to include "(latest version)".

> How does selecting a wildcard version (_) affect the export output (e.g. do I get glibc-_-.rpm or not when I export the depsolved contents of the recipe)? Not sure if relevant though.

@atodorov Are you referring to the following modal? Depsolving the blueprint contents will always return version and release with the package.
screen shot 2019-02-08 at 8 55 57 am

> @jgiardino do we progressively replace version numbers to the left, e.g. what if a package has 1.2.3.4 version numbering does this produce `*`, `1.*`, `1.2.*`, `1.2.3.*` ?

Yes, for each period-separated value, a variation of the version is provided that inserts a wildcard. I included screencaps of how "1.2.3.4" vs just "1" are handled:

screen shot 2019-02-08 at 8 53 28 am
screen shot 2019-02-08 at 8 55 27 am

File: core/sagas/inputs.js [hunk 1]
```diff
@@ -73,23 +75,68 @@ function flattenInput(response) {
...
   });
+  flattened.builds = addWildcardVersions(flattened.builds);
   return flattened;
 }
 
+function addWildcardVersions(builds) {
+  // create a new array of builds that includes wildcard versions
+  let wildcardBuilds = [];
+  builds.forEach(item => {
+    item.depsolveVersion = item.version;
```
larskarlitski: Is `depsolveVersion` ever different from `version`? If it is, why do you overwrite it here?

core/sagas/inputs.js
```diff
@@ -73,23 +75,68 @@ function flattenInput(response) {
...
+      const newWildcardVersion = createWildcardVersion(newBuilds[0].version);
+      const versionIndex = wildcardBuilds.findIndex(build => build.version === newWildcardVersion);
+      if (versionIndex === -1) {
+        const newWildcardBuild = Object.assign({}, item, {
+          version: newWildcardVersion
+        });
+        newBuilds.unshift(newWildcardBuild);
+      }
+    }
+    wildcardBuilds = wildcardBuilds.concat(newBuilds);
```
larskarlitski: This is a bit complex and inefficient, because it's iterating over the list of versions many times. If I understand what this is doing correctly, I think it can be simplified to something like this:
```javascript
function addWildcardVersions(builds) {
  // create a new array of builds that includes wildcard versions
  let wildcardBuilds = {}; 
  builds.forEach(item => {
    item.depsolveVersion = item.version;

    let parts = item.version.split(".");
    for (let i = 0; i < parts.length; i += 1) {
      let wildcard = parts
        .slice(0, i)
        .concat("*")
        .join(".");

      if (!(wildcard in wildcardBuilds)) {
        wildcardBuilds[wildcard = Object.assign({}, item, { version: wildcard }); 
      }   
    }   
  }); 

  return Object.values(wildcardBuilds);
}
```

larskarlitski:
> When listing the selected components for a blueprint, the version is presented as the depsolved version, and not the one saved with the blueprint. This makes it difficult to know what the actual saved value is.

This is a bit confusing indeed. If this is hard to fix, I'd be fine changing it in a separate pull request, though.

> In the list of available components, the tooltip for the Add button shows what the depsolved version of the default "*" version value would be. This could be confusing, and maybe just having it say "Add Component" or "Add Latest Version" would be better.

I don't think "Add Component" would add (haha) enough information to be useful as a tooltotip. Maybe something like "Add latest version (currently X.Y.Z)"?

> Related to the last bullet above, should the first menu option include "latest version", like this, or is that obvious?:

Yes!

core/sagas/inputs.js
jgiardino: Yes, the version could be "10.*" but the depsolveVersion is "10.11.0". Keeping the depsolveVersion associated with the wildcard versions allow me to show what the current depsolved version of "10.*" would be.

PR #538, opened 2019-02-15
Hostname config

jgiardino:
Adds ability to edit the hostname defined for a blueprint, which is similar to the ability to edit the blueprint description.

Also refactors the edit blueprint description functionality:

- Uses inline edit css available from PF3
- Moves input value out of the redux store to the component state to avoid lagging during typing
- refactors the function to cause a version bump when description is updated (previously, version would not change)
- Improves the keyboard interaction of the inline edit functionality (e.g. places focus on the input when it becomes visible, enables Escape key to exit out of inline edit)

image

### DIFF
--- a/pages/blueprint/index.js
+++ b/pages/blueprint/index.js
@@ -18,8 +18,13 @@ import EmptyState from "../../components/EmptyState/EmptyState";
 import BlueprintToolbar from "../../components/Toolbar/BlueprintToolbar";
 import ListView from "../../components/ListView/ListView";
 import ListItemImages from "../../components/ListView/ListItemImages";
-import { fetchingBlueprintContents, setBlueprintDescription, fetchingCompDeps } from "../../core/actions/blueprints";
 import TextInlineEdit from "../../components/Form/TextInlineEdit";
+import {
+  fetchingBlueprintContents,
+  setBlueprintDescription,
+  setBlueprintHostname,
+  fetchingCompDeps
+} from "../../core/actions/blueprints";
 import {
   clearSelectedInput,
   setSelectedInput,
@@ -37,7 +42,7 @@ import {
   setModalDeleteImageVisible,
   setModalDeleteImageState
 } from "../../core/actions/modals";
-import { setEditDescriptionVisible, setActiveTab } from "../../core/actions/blueprintPage";
+import { setEditDescriptionVisible, setEditHostnameVisible, setActiveTab } from "../../core/actions/blueprintPage";
 import {
   componentsSortSetKey,
   componentsSortSetValue,
@@ -85,6 +90,12 @@ const messages = defineMessages({
   },
   descriptionInputLabel: {
     defaultMessage: "Description"
+  },
+  hostnameButtonLabel: {
+    defaultMessage: "Edit hostname"
+  },
+  hostnameInputLabel: {
+    defaultMessage: "Hostname"
   }
 });
 
@@ -103,6 +114,7 @@ class BlueprintPage extends React.Component {
     this.handleHideModalStop = this.handleHideModalStop.bind(this);
     this.handleHideModalDeleteImage = this.handleHideModalDeleteImage.bind(this);
     this.handleEditDescription = this.handleEditDescription.bind(this);
+    this.handleEditHostname = this.handleEditHostname.bind(this);
     this.handleStartCompose = this.handleStartCompose.bind(this);
     this.downloadUrl = this.downloadUrl.bind(this);
   }
@@ -115,6 +127,7 @@ class BlueprintPage extends React.Component {
       this.props.fetchingComposes();
     }
     this.props.setEditDescriptionVisible(false);
+    this.props.setEditHostnameVisible(false);
     this.props.setModalExportBlueprintVisible(false);
   }
 
@@ -157,6 +170,14 @@ class BlueprintPage extends React.Component {
     this.props.setEditDescriptionVisible(state);
     if (!state && action === "commit") {
       this.props.setBlueprintDescription(this.props.blueprint, value);
+    }
+  }
+
+  handleEditHostname(action, value) {
+    const state = !this.props.blueprintPage.editHostnameVisible;
+    this.props.setEditHostnameVisible(state);
+    if (!state && action === "commit") {
+      this.props.setBlueprintHostname(this.props.blueprint, value);
     }
   }
 
@@ -230,8 +251,12 @@ class BlueprintPage extends React.Component {
       setSelectedInputParent,
       clearSelectedInput
     } = this.props;
-    const { editDescriptionVisible } = this.props.blueprintPage;
+    const { editDescriptionVisible, editHostnameVisible } = this.props.blueprintPage;
     const { formatMessage } = this.props.intl;
+    let hostname = "";
+    if (blueprint.customizations !== undefined && blueprint.customizations.hostname !== undefined) {
+      hostname = blueprint.customizations.hostname;
+    }
 
     return (
       <Layout className="container-fluid" ref={c => (this.layout = c)}>
@@ -313,11 +338,21 @@ class BlueprintPage extends React.Component {
                     <TextInlineEdit
                       className="col-sm-10"
                       editVisible={editDescriptionVisible}
-                      handleChange={this.handleChangeDescription}
                       handleEdit={this.handleEditDescription}
                       buttonLabel={formatMessage(messages.descriptionButtonLabel)}
                       inputLabel={formatMessage(messages.descriptionInputLabel)}
                       value={blueprint.description}
+                    />
+                  </form>
+                  <form className="form-group" onSubmit={() => this.handleEditHostname("commit")}>
+                    <span className="col-sm-2 control-label">{formatMessage(messages.hostnameInputLabel)}</span>
+                    <TextInlineEdit
+                      className="col-sm-10"
+                      editVisible={editHostnameVisible}
+                      handleEdit={this.handleEditHostname}
+                      buttonLabel={formatMessage(messages.hostnameButtonLabel)}
+                      inputLabel={formatMessage(messages.hostnameInputLabel)}
+                      value={hostname}
                     />
                   </form>
                 </div>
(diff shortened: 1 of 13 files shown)


jgiardino:
@henrywang The tests are failing, but it's weird that the failures don't seem to be related to the updates I made. Also, I'm not sure how to update the tests based on the changes I made. The html structure is different, and I'm wondering if I should add something specific to the html that makes it easier to select? Let me know if you have any suggestions. 

For this case, I think I would just replace `dd` with `span`?:
```
  get detailsTabBlueprintDescriptionLabel() {
    const selector = `span=${this.description}`;
```
Would I do this same for this one?
```
  updatedBlueprintDescriptionLabel(selector) {
    browser.waitUntil(
      () => browser.isExisting(`dd=${selector}`),
```
For this one, the button class name is `form-control-pf-value`, but so is the button for edit hostname. I can add a data attribute to this button so that we can distinguish between them. Let me know what you recommend.
```
  get editBlueprintDescriptionButton() {
    const selector = ".form-control-pf-value";
```
There are a few more related to editing the description, but I think I can figure the rest out from here if you provide feedback on these.

henrywang:
@jgiardino, never mind, I can take care of failed test issue. But I cannot load blueprint page. Do I need to update lorax-composer? Or what should I do? Thanks.
ooops

jgiardino:
Doh! I completely missed that issue. Thanks for catching that, @henrywang! It should be fixed now.

pages/blueprint/index.js
```diff
@@ -313,52 +339,33 @@ class BlueprintPage extends React.Component {
...
-                      {blueprint.description}
-                      <button className="btn btn-link" type="button">
-                        <span className="pficon pficon-edit" />
-                      </button>
-                    </dd>
-                  )}
-                </dl>
+              <div className="col-sm-12">
+                <div className="form-horizontal">
+                  <form className="form-group" onSubmit={() => this.handleEditDescription("commit")}>
```
henrywang: It'll be convenient to add `data-form="description"` here. :+1: 
There're two same forms, same edit icon, same text input. So It's hard to find description one.

pages/blueprint/index.js
```diff
@@ -313,52 +339,33 @@ class BlueprintPage extends React.Component {
...
+                      className="col-sm-10"
+                      editVisible={editDescriptionVisible}
+                      handleChange={this.handleChangeDescription}
+                      handleEdit={this.handleEditDescription}
+                      buttonLabel={formatMessage(messages.descriptionButtonLabel)}
+                      inputLabel={formatMessage(messages.descriptionInputLabel)}
+                      value={blueprint.description}
+                    />
+                  </form>
+                  <form className="form-group" onSubmit={() => this.handleEditHostname("commit")}>
```
henrywang: And add `data-form="hostname"` here. Thanks.

henrywang:
@jgiardino, never mind the test. I'll give you a diff file, you can update test code according to it.

henrywang:
@jgiardino, I have a suggestion here. From my perspective, I love the old style better. It's just my personal thought. :smiley: 

description_only
hostname

jgiardino:
@henrywang Thanks for the suggestions! I'll make those updates. 

As for the inline edit style, the original version was what I had implemented before this feature was finalized in [patternfly](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/inline-edit-for-form.html). I appreciate that you liked my version better 😄but for consistency I should use what's defined in patternfly for this.

henrywang:
@jgiardino, Here's diff for test update. It should work after update, I tested in locally already.
```shell
diff --git i/test/end-to-end/pages/ViewBlueprint.page.js w/test/end-to-end/pages/ViewBlueprint.page.js
index a5bfb8d2..39e553cc 100644
--- i/test/end-to-end/pages/ViewBlueprint.page.js
+++ w/test/end-to-end/pages/ViewBlueprint.page.js
@@ -95,18 +95,8 @@ class ViewBlueprintPage {
     return $(selector);
   }

-  get detailsTabBlueprintNameLabel() {
-    const selector = `dd=${this.name}`;
-    browser.waitUntil(
-      () => browser.isExisting(selector),
-      timeout,
-      `blueprint name under "Details" tab in View Blueprint page cannot be found by selector ${selector}`
-    );
-    return $(selector);
-  }
-
   get detailsTabBlueprintDescriptionLabel() {
-    const selector = `dd=${this.description}`;
+    const selector = `span=${this.description}`;
     browser.waitUntil(
       () => browser.isExisting(selector),
       timeout,
@@ -117,15 +107,15 @@ class ViewBlueprintPage {

   updatedBlueprintDescriptionLabel(selector) {
     browser.waitUntil(
-      () => browser.isExisting(`dd=${selector}`),
+      () => browser.isExisting(`span=${selector}`),
       timeout,
       `blueprint description under "Details" tab in View Blueprint page cannot be found by selector ${selector}`
     );
-    return $(`dd=${selector}`);
+    return $(`span=${selector}`);
   }

   get editBlueprintDescriptionButton() {
-    const selector = ".pficon-edit";
+    const selector = '[data-form="description"] .pficon-edit';
     browser.waitUntil(
       () => browser.isExisting(selector),
       timeout,
@@ -135,7 +125,7 @@ class ViewBlueprintPage {
   }

   get descriptionInputBox() {
-    const selector = '[id="blueprint-tabs-pane-details"] .form-control';
+    const selector = '[data-form="description"] .form-control';
     browser.waitUntil(
       () => browser.isExisting(selector),
       timeout,
@@ -145,7 +135,7 @@ class ViewBlueprintPage {
   }

   get okButton() {
-    const selector = ".fa-check";
+    const selector = '[data-form="description"] .form-control-pf-save';
     browser.waitUntil(
       () => browser.isExisting(selector),
       timeout,
@@ -155,7 +145,7 @@ class ViewBlueprintPage {
   }

   get cancelButton() {
-    const selector = ".btn-link .pficon-close";
+    const selector = '[data-form="description"] .form-control-pf-cancel';
     browser.waitUntil(
       () => browser.isExisting(selector),
       timeout,
diff --git i/test/end-to-end/specs/viewBlueprint.test.js w/test/end-to-end/specs/viewBlueprint.test.js
index 4e150440..f1881ed3 100644
--- i/test/end-to-end/specs/viewBlueprint.test.js
+++ w/test/end-to-end/specs/viewBlueprint.test.js
@@ -53,10 +53,6 @@ describe("View Blueprint Page", function() {
     });

     describe('"Details" tab', function() {
-      it(`blueprint name under "Details" tab should be "${name}"`, function() {
-        expect(viewBlueprintPage.detailsTabBlueprintNameLabel.getText()).to.equal(name);
-      });
-
       it(`blueprint description under "Details" tab should be "${description}"`, function() {
         expect(viewBlueprintPage.detailsTabBlueprintDescriptionLabel.getText()).to.equal(description);
       });

```

henrywang:
@jgiardino, the hostname string should have a check, because not all characters are valid for hostname.
http://man7.org/linux/man-pages/man7/hostname.7.html

henrywang:
> @henrywang Thanks for the suggestions! I'll make those updates.
> 
> As for the inline edit style, the original version was what I had implemented before this feature was finalized in [patternfly](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/inline-edit-for-form.html). I appreciate that you liked my version better but for consistency I should use what's defined in patternfly for this.

Understood. Thanks for making it clear.

jgiardino:
@henrywang Thanks for the test updates. I included those in the latest commit. And thanks for sharing the info about valid characters in hostname. I'll work on adding a check for that.

jgiardino:
@henrywang I included an update that checks the value entered for hostname and prevents the user from submitting an invalid value.

I also included help text below the input, which turns red when the value is invalid. I grabbed this from the [documentation](http://man7.org/linux/man-pages/man7/hostname.7.html) that @henrywang shared, although I'm allowing dots "." and am realizing that this text doesn't mention they are allowed. 😕 Any suggestions on that are welcome.

image

pages/blueprint/index.js
```diff
@@ -183,6 +193,12 @@ class BlueprintPage extends React.Component {
     event.preventDefault();
   }
 
+  handleEditHostnameValue(value) {
+    const validCharacters = /^[a-zA-Z0-9-.]{1,1000}$/.test(value);
```
jgiardino: @larskarlitski I was fumbling my way through this one. Any suggestions are much appreciated. :-)

henrywang:
Thanks @jgiardino. Yeah!! The checking works locally.
I tried to write a regx to validate hostname, but I really do not know how to cover `1 to 63 characters long for each element`. Here's my regx.
```javascript
^(?=.{1,253}$)(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$
```

pages/blueprint/index.js
larskarlitski: I think this could be good enough:
```regex
^(\d|\w)(\d|\w|-){0,62}(\.(\d|\w)(\d|\w|-){0,62})*$
```

It checks that the first letter is a digit or a word character (and not a `-`), followed by up to 62 digits, word characters. Then optionally a dot with the same. You'd only need to check that the whole string is shorter than 253 characters.

jgiardino: I included this change in my latest commit, but when I type in 63 characters my browser becomes unresponsive. Sometimes it becomes unresponsive after typing almost 63 characters and then a dot. Could you check this and see if you have the same issues?

jgiardino:
Thanks for reviewing, @henrywang. That's interesting that you aren't able to reproduce the issue I described above. 
I did come across this issue opened for Chrome which sounds similar to what I'm seeing: https://bugs.chromium.org/p/chromium/issues/detail?id=178790
I'm also able to reproduce this issue in Firefox.
Maybe this issue only happens on a mac, but I feel like it's worth addressing. I'll play around with other ideas to see what works.

jgiardino:
@henrywang @larskarlitski I just updated the regex I'm using to simplify it so that it doesn't crash. I'm not sure that the issue I saw would be reproducible on your machines, maybe it's just on a mac? Let me know if you have any suggestions on how to handle this. 

And then for squashing commits, I think they'd need to be reordered to do that, but I had issues when trying to do that before, so I would probably need some help there.

Thanks!

henrywang:
> @henrywang @larskarlitski I just updated the regex I'm using to simplify it so that it doesn't crash. I'm not sure that the issue I saw would be reproducible on your machines, maybe it's just on a mac? Let me know if you have any suggestions on how to handle this.
> 
> And then for squashing commits, I think they'd need to be reordered to do that, but I had issues when trying to do that before, so I would probably need some help there.
> 
> Thanks!
It works for me. Thanks for update. :+1:

jgiardino:
> What happens when I leave the field blank? Could we give some guidance that a default is used in this case instead of an empty hostname?

I'm not sure what best practice is for this. I was thinking that if the user leaves the field blank, then hostname has no value. Is a default value weird if this default would not be used for blueprints created from the cli?

Yesterday I was able to reorder and squash commits, but wanted to wait to push those until you had a chance to review the regex variations ;-)

jgiardino:
> I'm getting this when clicking the "Edit" button:

I think I captured this in the squashed changes I was waiting to push, but will double-check.

larskarlitski:
> I'm not sure what best practice is for this. I was thinking that if the user leaves the field blank, then hostname has no value. Is a default value weird if this default would not be used for blueprints created from the cli?

Sorry, I didn't mean to put a default, but to make it clearer (somehow) that leaving the field blank does not mean that the hostname will be empty. It means "use the default of the OS".

I can reproduce the crash with long hostnames. The regular expression was too computationally complex. This one works for me:
```regex
^[a-zA-Z0-9][a-zA-Z0-9-]{0,62}(\.[a-zA-Z0-9][a-zA-Z0-9-]{0,62})*$
```
Your changes work as well though and are more explicit.

> I think I captured this in the squashed changes I was waiting to push, but will double-check.

Great, thanks!

jgiardino:
Thanks for the feedback! I love this phase "computationally complex". 

I can add info about what happens when no hostname is defined. Does this sound accurate?

> If no hostname is provided, the hostname will be determined by the OS.

larskarlitski:
> I can add info about what happens when no hostname is defined. Does this sound accurate?
> 
> > If no hostname is provided, the hostname will be determined by the OS.

Perfect, thanks!

jgiardino:
@larskarlitski I just pushed updates that should address your comments. For the additional info, I have it so that nothing displays when a value is defined:

image
image



And then the text displays in either state if the value is empty:
image
image

PR #540, opened 2019-02-17
Remove browser-sync in devDependencies.

henrywang:
Two reasons:
1. It is not used any more.
2. It caused severity vulnerability due to lower version.


henrywang:
@larskarlitski, @martinpitt, I love you guys. :+1:

PR #542, opened 2019-02-18
Update lorax-composer using the updates-testing repository to test it before it enters into stable repo.

henrywang:
This update only gets enabled on Fedora and running test on firefox.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -1,7 +1,12 @@
 #!/bin/sh
 # image-customize script to enable cockpit in test VMs
 # The application RPM will be installed separately
-set -eu
+set -eux
+
+# Update lorax-composer using the updates-testing repository
+if grep -isq fedora /etc/redhat-release && grep -isq firefox /tmp/BROWSER; then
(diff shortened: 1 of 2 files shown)


test/vm.install
```diff
@@ -1,7 +1,12 @@
 #!/bin/sh
 # image-customize script to enable cockpit in test VMs
 # The application RPM will be installed separately
-set -eu
+set -eux
+
+# Update lorax-composer using the updates-testing repository
+if [[ $(grep -ish "Fedora" /etc/redhat-release) == Fedora* ]] && [[ $(grep -ish "firefox" /tmp/BROWSER == "firefox") ]]; then
```
martinpitt: `[[` is a bashism, and this is a /bin/sh script. Let's please not break it on Debian-based systems. :-)

This can be done more elegantly with `if grep -iq fedora /etc/redhat-release && ...`.

henrywang: Thanks @martinpitt, updated!

PR #545, opened 2019-02-22
 Add API response time and sanity functional test.

henrywang:
1. Update .gitignore and add .eslintignore 

- Make git ignore composer.covered.
- Do not run eslint on all node_modules and wdio_report. Eslint ignores node_modules folder in root folder by default, but does not ignore /test/end-to-end/node_modules. There's a app.js inside wdio_report. Eslint should ignore that.
2.  Add API response time and sanity functional test.

- Check API response time to find API issue ASAP.
- Reload browser before test suite starts to clear browser session.
- Add "cockpit" into eslintConfig global to avoid eslint error.

### DIFF
--- a/test/end-to-end/specs/api.test.js
+++ b/test/end-to-end/specs/api.test.js
@@ -1,0 +1,45 @@
+const commands = require("../utils/commands");
+
+// acceptable API response time
+// the 1 second comes from https://www.nngroup.com/articles/response-times-3-important-limits/
+const acceptable = 1000;
+
+describe("lorax-composer api sanity test", function() {
+  before(function() {
+    commands.login();
+    // Edge does not support the following command so have to keep default timeout(3 seconds)
+    // browser.timeouts({ script: timeout });
+  });
+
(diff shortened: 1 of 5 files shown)


henrywang:
In this PR, I added one API test for review. The rest of APIs will be added in the following PRs.

test/end-to-end/specs/api.test.js
```diff
@@ -0,0 +1,44 @@
+const commands = require("../utils/commands");
+
+// acceptable API response time
+// the 0.1 second comes from https://www.nngroup.com/articles/response-times-3-important-limits/
+const acceptable = 100;
```
larskarlitski: Great to have this kind of test. I'm a bit worried that this might become flaky on overloaded test servers though.

martinpitt: Right, that's my worry as well. But we can see how it goes, and bump it if it fails noticeably on the infra.

henrywang:
@larskarlitski and @martinpitt, thanks for review. I agree with you. I'll change to 1 second.

PR #564, opened 2019-02-28
Fix two dependencies issue

henrywang:
1. Remove hightlight.js because we do not need it any more and it has a bug which caused installation issue.
2. Update bootstrap to 3.4.1 to fix CVE-2019-8331.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "7.2.5",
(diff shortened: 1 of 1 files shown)


henrywang:
@martinpitt, I think we can close [PR 547](https://github.com/weldr/welder-web/pull/547) because they are doing the same thing. But I have to say, the cockpit bot is really good. It send update 4 days before CVE-2019-8331 reported in Github UI.

PR #594, opened 2019-03-13
Include ability to add a user to the blueprint

jgiardino:
- User account creation allows for defining name, description, password, ssh key and also enabling the user to be a server admin (i.e. groups=["wheel"]).
  - When an ssh key is provided, no validation occurs. When the account is listed on the Details page, if the key includes an email address, that is the only part of the key that displays in the table.
  - When a password is provided, an encrypted password is stored in the blueprint. When the account is listed on the Details page, only a checkmark icon displays to indicate that a password is set. When editing a user account, the password can be deleted or a new password can be set.
- User accounts can also be updated or removed.
- User accounts that are created for a blueprint are listed on the Details tab of a blueprint.
- Text strings and functionality match Cockpit as much as possible, but there are some differences given that the workflow is slightly different

image

image

image

### DIFF
--- a/components/Form/Password.js
+++ b/components/Form/Password.js
@@ -1,0 +1,218 @@
+import React from "react";
+import { FormattedMessage, defineMessages, injectIntl, intlShape } from "react-intl";
+import PropTypes from "prop-types";
+import cockpit from "cockpit";
+
+const messages = defineMessages({
+  weakPassword: {
+    defaultMessage: "Password is too weak"
+  },
+  rejectedPassword: {
+    defaultMessage: "Password is not acceptable"
+  }
+});
+
+class Password extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      passwordOne: "",
+      passwordTwo: "",
+      displayWarningMatch: false,
+      passwordQuality: "",
+      warningQuality: "",
+      displayWarningQual: false
+    };
+    this.handleChangePasswordOne = this.handleChangePasswordOne.bind(this);
+    this.handleChangePasswordTwo = this.handleChangePasswordTwo.bind(this);
+    this.handleWarnings = this.handleWarnings.bind(this);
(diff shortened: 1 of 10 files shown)


jgiardino:
Thanks for the review!

> 1. The same user can be added.

When I enter the name of an existing user, the dialog will display as shown below. But I also included the same functionality in cockpit where the user name is based on the full name, and I did not include that same check in that case, so I will fix that. 

Can you confirm if that's how you were able to create a duplicate user account? The names shown in your example are not consistent with that workflow, so I'm curious what steps you took to produce those duplicate entries. 

image

> 2. The user which only has username can be created. Is this by design?

Yes, that's by design.

henrywang:
@jgiardino, it's easy to reproduce. :-) Just fill in all items, then the Create button will be clickable.

jgiardino:
I fixed both the issue that @henrywang found and the one I found. I also changed a class on the buttons that display in the table so that they match a similar example in [PF3 react](https://patternfly-react.surge.sh/patternfly-3/?selectedKind=patternfly-react%2FContent%20Views%2FTable%20View&selectedStory=Table%20With%20Inline%20Edit%20Row&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).

henrywang:
@jgiardino, this issue cannot be found after applying your patch. Thanks for fixing this issue!
I found another issue. When you create a new blueprint and then add user, the user cannot be created. This issue can be reproduced only on new created blueprint.
How to reproduce:
1. Create a new blueprint, add component into blueprint.
2. Click blueprint name link to go to view blueprint page.
3. Create user. Filled in all items, then click Create button.
There's no user in the UI. And after I check the `.toml` file, there's no user added in that file.
After that, user can be created.
Could you please take a look? Thanks.

jgiardino:
@henrywang Thanks for reviewing again, and catching that issue! I just pushed a fix for that.

PR #599, opened 2019-03-20
cockpituous: Release to Fedora Rawhide

martinpitt:
cockpit-composer is in rawhide now, so upload new releases automatically
to there.

### DIFF
--- a/utils/cockpituous-release
+++ b/utils/cockpituous-release
@@ -9,7 +9,14 @@ RELEASE_SOURCE="_release/source"
 RELEASE_SPEC="cockpit-composer.spec"
 RELEASE_SRPM="_release/srpm"
 
+# Authenticate for pushing into Fedora dist-git
+cat ~/.fedora-password | kinit cockpit@FEDORAPROJECT.ORG
+
 job release-source
 job release-srpm
+
+# Do fedora builds for the tag, using tarball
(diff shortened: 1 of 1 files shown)


utils/cockpituous-release
```diff
@@ -9,7 +9,14 @@ RELEASE_SOURCE="_release/source"
...
 RELEASE_SRPM="_release/srpm"
 
+# Authenticate for pushing into Fedora dist-git
+cat ~/.fedora-password | kinit cockpit@FEDORAPROJECT.ORG
+
 job release-source
 job release-srpm
+
+# Do fedora builds for the tag, using tarball
+job release-koji master
```
larskarlitski: Cockpit's release script passes `-k` here to not fail the release on build failures. Is not doing this here intentional?

martinpitt: I did that deliberately, yes. Unlike cockpit, the composer.rpm build is little more than a `cp -a`.  The only realistic case where that fails is that libappstream-glib is not installable in rawhide (c-c's only BuildRequires).

Once we have a Fedora 30 branch, we could switch rawhide to `-k`, I don't have a strong opinion about that, though.

larskarlitski: Thanks!

PR #601, opened 2019-03-22
package.json: Explicitly include js-yaml 3.13.0

larskarlitski:
Earlier versions have a security vulnerability:

  https://www.npmjs.com/advisories/788

It is a dependency of eslint. Pull in 3.13.0 explicitly until eslint is
updated.

Also remove front-matter, which had js-yaml as a dependency as well.
We're not using it.


larskarlitski:
> But if it pacifies the github page, sure :-)

And travis. I wonder if there's a way to only `npm audit` the actual dependencies...

PR #604, opened 2019-03-22
fixes issue with fetching contents for Export modal

jgiardino:
Unlike the other pages, the Export modal on the Blueprints page requires an api call to fetch the contents of the modal. This was broken in #500 when the api calls were refactored. This update
moves the functionality from the modals saga to the blueprints saga to reuse functions that handle flattening the list of components after the api call.

### DIFF
--- a/core/sagas/blueprints.js
+++ b/core/sagas/blueprints.js
@@ -38,6 +38,7 @@ import {
   FETCHING_COMP_DEPS,
   setCompDeps
 } from "../actions/blueprints";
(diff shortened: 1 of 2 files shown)


jgiardino:
@henrywang I see there is a test for the Export, but since the contents are retrieved using different methods between the Blueprints page versus other pages (ie. fetching contents vs using already fetched contents), do you think it would be possible to include a test for both cases? It isn't obvious to me when looking at the test files where the Export modal is called (i.e. which page).

martinpitt:
[edge test failed](https://209.132.184.41:8493/logs/pull-604-20190325-060606-38ce49e1-cockpit-fedora-29-edge/log) twice in a row, retrying once more and comparing logs.

PR #613, opened 2019-04-01
core: Don't hide errors in sagas

stefwalter:
While testing Composer on RHEL 8 Betas I've noticed various
failures in the javascript console, but the actual error content
was hidden. This impedes support, especially of production systems.

I've added full error output to go to the javascript console
in each of these cases.


jgiardino:
Thanks for making these changes!

PR #615, opened 2019-04-01
Remove unused dependencies from package.json

KKoukiou:


KKoukiou:
@jgiardino @henrywang Are these really leftovers from the past?

KKoukiou:
looks like cjs is used by travis ci.. https://travis-ci.org/weldr/welder-web/builds/514185579?utm_source=github_status&utm_medium=notification

jgiardino:
Thanks for making these updates! I'm not aware of any need for these dependencies.

martinpitt:
The [Travis failure](https://travis-ci.org/weldr/welder-web/jobs/514185581) looks relevant -- perhaps `postcss` was a transitive dependency of the removed ones, and now needs to be added explicitly?

henrywang:
> The [Travis failure](https://travis-ci.org/weldr/welder-web/jobs/514185581) looks relevant -- perhaps `postcss` was a transitive dependency of the removed ones, and now needs to be added explicitly?

Correct! `pixrem` depends on `postcss`.

KKoukiou:
> The [Travis failure](https://travis-ci.org/weldr/welder-web/jobs/514185581) looks relevant -- perhaps `postcss` was a transitive dependency of the removed ones, and now needs to be added explicitly?

Yeah, probably. Still, since stylelint uses it, I don't get why it's not dependency of that.. I repushed with adding postcss back.

PR #620, opened 2019-04-04
test: add Export test on bluepirnts page to address PR#604

henrywang:
1. Add blueprint export test on blueprints page.
2. Move apiFetchTest() function to utils/commands.js because it will be re-used by api.test.js and blueprints.test.js

### DIFF
--- a/test/end-to-end/specs/api.test.js
+++ b/test/end-to-end/specs/api.test.js
@@ -14,25 +14,9 @@ describe("lorax-composer api sanity test", function() {
     blueprintsPage.loading();
   });
 
-  function apiFetchTest(endpoint, options = { body: "" }) {
-    options.path = endpoint;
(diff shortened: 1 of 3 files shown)


martinpitt:
@henrywang : I manually triggered a RHEL 7.7 and 8.1 test, to see if that works. The code doesn't have any special-cases for rhel-8-0, but we do have a special case for rhel-7-6 (depending on which lorax version/image formats are available). I'll send a separate PR if that needs adjustments.

henrywang:
@martinpitt, Thanks.
And one more special for rhel-7-x is repo setting in [vm.install](https://github.com/weldr/welder-web/blob/e4eb06088eb2971e9922f9f9a9b63898ed1418ea/test/vm.install#L25)

PR #621, opened 2019-04-04
test: Generalize image format condition for all RHEL 7.x

martinpitt:
This also applies to RHEL 7.7.


martinpitt:
I manually triggered 7.7 and 8.1 tests to validate that this works everywhere now. Once it does, I'll enable these in Cockpit for all PRs.

PR #622, opened 2019-04-04
test: add case to test hostname setting

henrywang:


henrywang:
@martinpitt, could you please review this PR if you have time? Thanks! :-)

test/end-to-end/pages/ViewBlueprint.page.js
```diff
@@ -154,6 +154,56 @@ class ViewBlueprintPage {
...
+
+  get editHostnameButton() {
+    const selector = '[data-form="hostname"] .pficon-edit';
+    browser.waitUntil(
+      () => browser.isExisting(selector),
+      timeout,
+      `edit hostname button under "Details" tab in View Blueprint page cannot be found by selector ${selector}`
+    );
+    return $(selector);
+  }
```
martinpitt: This is a lot of red tape and redundancy. Could this be factorized into a helper function that waits for a given selector, and then gets used directly in the tests? That would make  them easier to maintain (you directly see the selectors in the tests, and avoid the abstract `somethingItem()` names, so it matches up much better with what you see in the Inspector); it would also avoid having to add so much boilerplate for each new test.

henrywang: Thanks @martinpitt. It's already on my todo list. My plan is to do that when I update webdriver.io to v5.

PR #624, opened 2019-04-06
Refactor CreateImage && CreateBlueprint modal dialogs

KKoukiou:
This commit introduces the following changes:
* The dialog uses now Modal component from patternfly-react
* All properties that can be derived from others were removed from props
* The Button with the Modal for Create image belong now to the same
component. This removes the need for handlers getting passed all
around
* Stop using jquery for handling the modal state
* Remove from global state all properties that are Modal state specific
* The list of images types was moved out from the modals object in the
global state, since it's not really related to the that

### DIFF
--- a/components/ListView/BlueprintListView.js
+++ b/components/ListView/BlueprintListView.js
@@ -2,6 +2,7 @@ import React from "react";
 import { FormattedMessage } from "react-intl";
 import PropTypes from "prop-types";
 import Link from "../Link/Link";
+import CreateImage from "../Modal/CreateImage";
 
 class BlueprintListView extends React.PureComponent {
   constructor() {
@@ -9,7 +10,7 @@ class BlueprintListView extends React.PureComponent {
   }
 
   render() {
-    const { blueprints } = this.props; // eslint-disable-line no-use-before-define
+    const { blueprints, imageTypes, layout } = this.props;
     return (
       <div className="list-group list-view-pf list-view-pf-view">
         {blueprints.map(blueprint => (
@@ -18,13 +19,7 @@ class BlueprintListView extends React.PureComponent {
               <Link to={`/edit/${blueprint.name}`} className="btn btn-default">
                 <FormattedMessage defaultMessage="Edit Blueprint" />
               </Link>
-              <button
-                type="button"
-                className="btn btn-default"
-                onClick={e => this.props.handleShowModalCreateImage(e, blueprint)}
-              >
-                <FormattedMessage defaultMessage="Create Image" />
-              </button>
+              <CreateImage blueprint={blueprint} imageTypes={imageTypes} layout={layout} />
               <div className="dropdown pull-right dropdown-kebab-pf">
                 <button
                   className="btn btn-link dropdown-toggle"
@@ -80,14 +75,17 @@ BlueprintListView.propTypes = {
   handleShowModalDelete: PropTypes.func,
   blueprints: PropTypes.arrayOf(PropTypes.object),
   handleShowModalExport: PropTypes.func,
-  handleShowModalCreateImage: PropTypes.func
+  imageTypes: PropTypes.arrayOf(PropTypes.object).isRequired,
+  layout: PropTypes.shape({
+    setNotifications: PropTypes.func
+  })
 };
 
 BlueprintListView.defaultProps = {
   handleShowModalDelete: function() {},
   blueprints: [],
   handleShowModalExport: function() {},
-  handleShowModalCreateImage: function() {}
+  layout: {}
 };
 
 export default BlueprintListView;
--- a/components/Modal/CreateBlueprint.js
+++ b/components/Modal/CreateBlueprint.js
@@ -3,7 +3,6 @@ import { Modal } from "patternfly-react";
 import { FormattedMessage } from "react-intl";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import { setModalCreateBlueprintBlueprint } from "../../core/actions/modals";
 import { creatingBlueprint } from "../../core/actions/blueprints";
 
 class CreateBlueprint extends React.Component {
@@ -39,8 +38,6 @@ class CreateBlueprint extends React.Component {
         {this.state.showModal && (
           <CreateBlueprintModal
             blueprintNames={this.props.blueprintNames}
-            setModalCreateBlueprintBlueprint={this.props.setModalCreateBlueprintBlueprint}
-            createBlueprint={this.props.createBlueprint}
             creatingBlueprint={this.props.creatingBlueprint}
             close={this.close}
           />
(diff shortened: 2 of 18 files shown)


KKoukiou:
Sorry for making it a single commit, it's quite hard to separate the changes. @henrywang Could you please help me understand the test failures? make check, clearly shows some failures, but I am having trouble locate the failing line in the test file.

KKoukiou:
@jgiardino You 'will see in the commit I change a bit the redux store structure.
IMO the store should contain information that come from the API (blueprints, image list) and nothing else. Everything else should be handled in the React component state.
So the store now looks like this:
redux-state

I would do some efffort this week to remove this modals object from the store, doing one modal per commit. Then we can continue to different objects.

You will see that the code it much easier to read and also less :)

KKoukiou:
> Sorry for making it a single commit, it's quite hard to separate the changes. @henrywang Could you please help me understand the test failures? make check, clearly shows some failures, but I am having trouble locate the failing line in the test file.

update: @henrywang looks like I fixed it, tests are green now.

package.json
```diff
@@ -178,7 +178,8 @@
       "jsx-a11y/no-onchange": "off",
       "jsx-a11y/label-has-for": "off",
       "redux-saga/no-unhandled-errors": 0,
-      "jsx-a11y/anchor-is-valid": "off"
+      "jsx-a11y/anchor-is-valid": "off",
+      "jsx-a11y/no-autofocus": "off"
```
jgiardino: 👍 I agree with your commit comments on this one:

> a11y complains about autoFocus, so this commit turns off that rule as
well.
> There are cases where autoFocus decreases accessibility, but let's judge it buy ourselves, since this use of it does not cause any harm. 

A modal is definitely a case where it's important to set focus for accessibility. 

Related to this, the [wai-aria authoring spec for modals](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) recommends placing focus back to the modal toggle when it closes. I noticed that before your changes, this was the case with the Create Blueprint button on the Blueprints page, but nowhere else. And after your changes, this is the case with the Create Image button on the Blueprints page, but nowhere else. Do you happen to know why that changed? And if it's easy to implement that behavior for all of them?

Also, I tried to do this exact same thing (i.e. `<input>` with `autofocus`) in the User Account modal, and could never get it to work. 😢

components/Modal/CreateBlueprint.js
```diff
@@ -92,216 +114,151 @@ class CreateBlueprint extends React.Component {
...
-              >
-                <span className="pficon pficon-close" />
-              </button>
-              <h4 className="modal-title" id="myModalLabel">
-                <FormattedMessage defaultMessage="Create Blueprint" />
-              </h4>
+      <Modal show id="cmpsr-modal-crt-blueprint">
+        <Modal.Header>
+          <Modal.CloseButton onClick={this.props.close} />
+          <FormattedMessage defaultMessage="Create Blueprint" />
```
jgiardino: Weird, I posted a comment here, and it disappeared.

The font weight isn't as heavy as it should be. I think it's because you're missing `<Modal.Title>` here, which is being used in the [PF3 component example](https://patternfly-react.surge.sh/patternfly-3/?selectedKind=patternfly-react%2FForms%20and%20Controls%2FModal%20Overlay%2FComponents&selectedStory=Modal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).

jgiardino: This is also the case with the Create Image modal.

core/reducers/index.js
```diff
@@ -14,7 +15,8 @@ const rootReducer = combineReducers({
   modals,
   sort,
   filter,
-  composes
+  composes,
+  images
```
jgiardino: What are your thoughts on combining `images` and `composes` into one, and having all things related to image creation together?

package.json
KKoukiou: You mean to focus on the button that actually opened the modal when closing the modal?
So if you want this behavior I just need to add this logic in the close function.

Regarding the autoFocus, in react it's autoFocus not autofocus. Maybe you did this wrong?

components/Modal/CreateBlueprint.js
KKoukiou: Right, good catch, I will add it back.

core/reducers/index.js
KKoukiou: Oh right, composes object has the list of the created images. We should really stick to either composes or images for naming vars, I know realized it's the same things.
Your comment makes sense. I will update accordingly.

package.json
jgiardino: Yes, by "toggle" I meant the button that opens the modal. If you can add that behavior, that would be great. 😄 👍

core/reducers/index.js
jgiardino: 👍 BTW, I like "images" but am fine with either label.

KKoukiou: I would just move the previous imageTypes object into the existing composes object, so that I will not introduce huge change. But feel free to change it in future.

package.json
KKoukiou: I was not able to find a way to implement this. :/ If this is not a regression, can we merge without this feature?

jgiardino:
The updates you made related to my comments look great!

There's one thing I missed in my first review. Currently, when the dialogs are open, I can hit Esc to dismiss them. This is missing in your updates. Is that something you can add?

components/ListView/BlueprintListView.js
```diff
@@ -80,14 +75,17 @@ BlueprintListView.propTypes = {
   handleShowModalDelete: PropTypes.func,
   blueprints: PropTypes.arrayOf(PropTypes.object),
   handleShowModalExport: PropTypes.func,
-  handleShowModalCreateImage: PropTypes.func
+  imageTypes: PropTypes.arrayOf(PropTypes.object).isRequired,
+  layout: PropTypes.shape({
+    setNotifications: PropTypes.func
+  })
```
larskarlitski: Curious: why is `setNotifications` in a separate `layout` object?

The `isRequired` is added in the first commit and removed in the second.

KKoukiou:
> The updates you made related to my comments look great!
> 
> There's one thing I missed in my first review. Currently, when the dialogs are open, I can hit Esc to dismiss them. This is missing in your updates. Is that something you can add?

Just fixed the Esc closing dialog thingy.

components/ListView/BlueprintListView.js
KKoukiou: Oh, this is a method that is implemented in the Layout component so basically whenever we want this pop up notification we need a reference to that component to call the setNotifications function. It's not good, but that's how it was before. 
I was thinking already how to make it better.

Reposting to fix this double change of the isRequired field.

KKoukiou:
> This is great cleanup, thanks!
> 
> Please squash the commit that renames variables you introduced in the previous one.

I didn't really introduce them, they were in the store, and I just put them in the component state. But I can squash them definitely.

jgiardino:
I tested the Esc functionality. Thanks for adding that!

Issue #625, opened 2019-04-06
Should not create an "all empty" user account


henrywang:
How to reproduce:
This issue is only able to reproduce when you first login cockpit-composer and enter into one of bluepirnt's view blueprint page, and then click Create User Account. You can see the following dialog.
![Screenshot_2019-04-06 Image Builder - localhost localdomain](https://user-images.githubusercontent.com/672230/55671611-6aca7c00-58c4-11e9-88cb-c95f0cdce8ee.png)

KKoukiou:
@henrywang  I don't seem to be able to reproduce. Whenever I am trying to create a new user, the user input field, is auto filled with my current user.

henrywang:
@KKoukiou, aha! Sorry for making you confused. The issue is an all-empty user can be created without any error message.
How to reproduce:
1. click "Create User Account" button.
2. click "Create" button.
Then an all-empty user will be created. And when you delete this user, an internal error was reported.

Issue #626, opened 2019-04-06
The div id property in Create User Account dialog should not be the same as Create Blueprint dialog's


henrywang:
https://github.com/weldr/welder-web/blob/4358b9a4d2cfd2a0ca0a5bf54293850e3e74e7b4/components/Modal/UserAccount.js#L199
It's the same as Create Blueprint dialog's.
https://github.com/weldr/welder-web/blob/4358b9a4d2cfd2a0ca0a5bf54293850e3e74e7b4/components/Modal/CreateBlueprint.js#L116

PR #627, opened 2019-04-09
User account issues

jgiardino:
Closes issues #625 and #626. Also addresses an issue where the user enters a value for the first password field then delete the value. Before this update, the Create button would remain disabled, but instead, it should become enabled again.

### DIFF
--- a/components/Form/Password.js
+++ b/components/Form/Password.js
@@ -53,6 +53,9 @@ class Password extends React.Component {
     }
     if (event.target.value === this.state.passwordTwo) {
       this.setState({ displayWarningMatch: false });
+      if (event.target.value === "") {
+        this.props.setValidPassword(event.target.value);
+      }
     }
   }
 
--- a/components/Modal/UserAccount.js
+++ b/components/Modal/UserAccount.js
@@ -191,7 +191,7 @@ class UserAccount extends React.Component {
     const disabledSubmit =
       userAccount.showDuplicateUser ||
       userAccount.showInvalidName ||
-      userAccount.length === 0 ||
+      userAccount.name.length === 0 ||
       userAccount.showInvalidPassword;
     return (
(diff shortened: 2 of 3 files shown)


henrywang:
@KKoukiou, do you have any comment on this PR? Thanks.

components/Form/Password.js
```diff
@@ -53,6 +53,9 @@ class Password extends React.Component {
     }
     if (event.target.value === this.state.passwordTwo) {
       this.setState({ displayWarningMatch: false });
+      if (event.target.value === "") {
```
KKoukiou: I don't understand this change. I think the 'Create' button should be disabled whenever some mandatory field is missing from the modal.
That said, having the password erased should always disable the Create button (except if the password is not mandatory field)

jgiardino: I'm on the fence about that. My general inclination is to provide the user the freedom to do what they want, and not place restrictions on user actions based on our assumptions about how things should/shouldn't work. I would suggest we only require fields that would result in an api failure. And in cases like this where we know the primary use case for defining a user account would fail, we can always inform them with an inline warning about the consequences. But I don't think we should prevent them from adding information to a user account and saving those changes. 

cc'ing @larskarlitski and @andreasn for their input on this

For this case, should we display an inline warning? (and would that be a separate PR?)

jgiardino:
Thanks for reviewing this, @KKoukiou! I included a reply to your comment.

components/Form/Password.js
KKoukiou: I don't really have strong opinion on this. We can introduce a warning in a seperate PR definitely.

larskarlitski: I don't feel strongly about whether we're showing a warning or disallowing it completely.

PR #629, opened 2019-04-10
test: update test case to support two more image types.

henrywang:
It's rhel-8-1 ONLY. Two new types are not supported on Fedora.
This's a quick fix. Case will be updated when UI has update according to new types.


henrywang:
@martinpitt, Thanks for reviewing! I totally agree with you. In this case, the purpose is not a "test", but a change checking. If lorax-composer has any changes, the UI should get updated. So we can know that ASAP.

PR #630, opened 2019-04-10
test: add test case for Create User Account

henrywang:
1. Add data-xx property in table element in Users section to make locate page element easy.
2. Add test case for Create User Account function.


henrywang:
Case failed on rhel-7-7 is due to #631.

test/end-to-end/pages/createUserAccount.page.js
```diff
@@ -0,0 +1,148 @@
+// Create User Account
+class CreateUserAccount {
+  constructor() {
+    this.containerSelector = '[id="cmpsr-modal-user-account"]';
+  }
+
+  loading() {
+    // sometimes the style attribute is style="display: block; padding-right: 12px;"
```
KKoukiou: Why we need this comment here? Is it related anyhow to the test?

henrywang:
@KKoukiou, @jgiardino Thanks for review. I'll not land this PR until #631 got fixed. Otherwise the following test on rhel-7-7 will fail.

martinpitt:
@henrywang : The conflicts aren't trivial -- can you please rebase this, so that we can land this soon? This should be unblocked now. Thanks!

PR #633, opened 2019-04-12
WIP: Refactor DeleteBlueprint modal dialog to use Modal from PF react

KKoukiou:
The following changes where also introduced:
* The link that triggers the dialog was moved to the same component as
the modal
* All delete blueprint modal related fields where removed from the redux
store, and similarly all related saga actions
* Stop using jquery to show/hide the modal; instead use event handlers on
the buttons


martinpitt:
Nice! I'll squash these together.

PR #639, opened 2019-04-17
error state with option to start lorax-composer

### LINKED ISSUE #460: No good error state if the lorax-composer.service is not running
If lorax-composer.service isn't running, I get this:
screenshot from 2018-11-27 11-38-31

The error state is very vague, and offers little guidance on how to get out of this state.

Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1708387

jgiardino:
Closes #460 

These updates include a custom error message when the error "not-found" is returned. The error state and actions are based on [similar UI in cockpit](https://github.com/cockpit-project/cockpit-podman/commit/24665af7d0f7d2529b34f0797eeba380ec60e4a0).

The second commit also prevents the empty state message from displaying while waiting for the blueprint contents to load.

### DIFF
--- a/core/reducers/blueprints.js
+++ b/core/reducers/blueprints.js
@@ -46,7 +46,8 @@ const blueprints = (state = [], action) => {
       });
     case FETCHING_BLUEPRINT_NAMES_SUCCEEDED:
       return Object.assign({}, state, {
-        fetchingBlueprints: false
+        fetchingBlueprints: false,
+        blueprintNames: action.payload.blueprints
       });
     // The following reducers filter the blueprint out of the state and add the new version if
     // the blueprint contains component data or is not found in the state
@@ -55,6 +56,7 @@ const blueprints = (state = [], action) => {
         !state.blueprintList.some(blueprint => blueprint.present.id === action.payload.blueprint.id)
         ? Object.assign({}, state, {
             fetchingBlueprints: false,
+            blueprintNames: [],
             blueprintList: [
               ...state.blueprintList.filter(blueprint => blueprint.present.id !== action.payload.blueprint.id),
               {
--- a/core/reducers/blueprints.js
+++ b/core/reducers/blueprints.js
@@ -1,3 +1,4 @@
+import { combineReducers } from "redux";
 import {
   UNDO,
   REDO,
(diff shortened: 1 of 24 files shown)


pages/blueprints/index.js
```diff
@@ -114,6 +122,24 @@ class BlueprintsPage extends React.Component {
     e.stopPropagation();
   }
 
+  startService(e) {
+    if (!e || e.button !== 0) return;
+
+    let argv;
+    if (this.state.enableService) argv = ["systemctl", "enable", "--now", "lorax-composer.socket"];
+    else argv = ["systemctl", "start", "lorax-composer.socket"];
```
larskarlitski: Please put the bodies of the `if` and `else` keywords on separate, indented lines.

pages/blueprints/index.js
```diff
@@ -43,6 +43,9 @@ const messages = defineMessages({
   errorTitle: {
     defaultMessage: "An Error Occurred"
   },
+  errorTitleInactive: {
+    defaultMessage: "Image Building Service (lorax-composer) is Not Active"
```
larskarlitski: This message seems a bit clunky to me. Do we need to mention that it's called `lorax-composer` in the title?

Do we usually use title case?

pages/blueprints/index.js
```diff
@@ -147,9 +173,33 @@ class BlueprintsPage extends React.Component {
...
+                        type="checkbox"
+                        checked={this.state.enableService}
+                        onChange={e => this.setState({ enableService: e.target.checked })}
+                      />
+                      Automatically start lorax-composer on boot
+                    </label>
+                  </div>
+                  <div className="blank-slate-pf-main-action">
+                    <button className="btn btn-primary btn-lg" type="button" onClick={this.startService}>
+                      Start Service
```
larskarlitski: Would "Start" be enough here?

pages/blueprints/index.js
jgiardino: Here is PF3 documentation for the [Empty State pattern](https://www.patternfly.org/pattern-library/communication/empty-state/). They use headline style capitalization. But in PF4 they are making a change to recommend sentence style for most strings. I'd opt to follow what other Cockpit pages use now—which I think is headline—and update later when we migrate to PF4.

jgiardino:
This is what I have after making some of the updates you suggest. It seems a little odd to me, but I'm fine with it if it's consistent with other pages in cockpit. Please let me know if it's not.

image

jgiardino:
> Should probably also be shown when the service stops. Would love to have the the warning icon on here as well, to be consistent with other pages in Cockpit. Both of these can be done as a follow-up though.

Thanks for catching that! I totally missed the icon, and also some strings for translation, so I'm adding that to this PR.

I'm not sure how I would show this when the service stops, so let's handle that later. And also, these are other things I noticed:
- If I delete all blueprints, stop the service, then restart it, the oob blueprints come back, which is not what I was expecting
- In one of my vms, I tried to start the service, and got the error message "Server has closed the connection". I could never figure out what happened there and if there's anything that could be done in the UI to resolve that.

core/reducers/blueprints.js
```diff
@@ -38,6 +39,11 @@ const blueprints = (state = [], action) => {
           }
         ]
       });
+    case FETCHING_BLUEPRINTS:
+      return Object.assign({}, state, {
+        fetchingBlueprints: true,
+        errorState: null
```
KKoukiou: I don't understand why the FETCHING_BLUEPRINTS reducer should clean up the errorState variable?
Generally, I tend to think of redux actions as single operation components, meaning that they should not affect sometihng extra than what their name says.
In this case, I would expect that the FETCHING_BLUEPRINTS action to be just setting the fetchingBlueprints to true and nothing else. (Even better would be FETCHING_BLUEPRINTS_INIT so that it's clear that this is just initialization action and not actually returning any blueprints list. But since this action existed no need to change that here)

The question about errorState remains though.

KKoukiou:
> This is what I have after making some of the updates you suggest. It seems a little odd to me, but I'm fine with it if it's consistent with other pages in cockpit. Please let me know if it's not.
> 
> image

For comparison here is the screenshot from the cockpit-machines page in case the libvirt service is not running.
service-libvirtd-cockpit-machines


I am not suggesting any changes, just to have it compared.

pages/blueprints/index.js
```diff
@@ -114,6 +131,25 @@ class BlueprintsPage extends React.Component {
...
+      argv = ["systemctl", "start", "lorax-composer.socket"];
+    }
+    cockpit
+      .spawn(argv, { superuser: "require", err: "message" })
+      .then(() => this.props.fetchingBlueprints())
+      .catch(err => console.error("Failed to start lorax-composer.socket:", JSON.stringify(err)));
+  }
+
+  goToServicePage(e) {
+    if (!e || e.button !== 0) return;
```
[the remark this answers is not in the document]
jgiardino: @martinpitt I copied this from https://github.com/cockpit-project/cockpit-podman/commit/24665af7d0f7d2529b34f0797eeba380ec60e4a0

Could you answer @KKoukiou's question here?

jgiardino:
In response to the previous comment from @KKoukiou, I had copied that from this example that 
@martinpitt shared: https://github.com/cockpit-project/cockpit-podman/commit/24665af7d0f7d2529b34f0797eeba380ec60e4a0

I'm fine with either layout. Adding @andreasn here to see if he can shed some light on which is the most current layout. 😄

pages/blueprints/index.js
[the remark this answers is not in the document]
marusak: @jgiardino the check guards that only primary mouse button (left one) clicks trigger the following code. When you do right click (e.button === 1) or wheel click (e.button === 2) it is immediately returned and the code of `goToServicePage` is not executed. Unless it is valid usecase that those other clicks should redirect you as well than this condition should be not here, but I doubt that, so this is right.

KKoukiou:
> In response to the previous comment from @KKoukiou, I had copied that from this example that
> @martinpitt shared: [cockpit-project/cockpit-podman@24665af](https://github.com/cockpit-project/cockpit-podman/commit/24665af7d0f7d2529b34f0797eeba380ec60e4a0)
> 
> And then I changed the primary button label based on a suggestion from @larskarlitski, which made it more inconsistent smile
> 
> I'm fine with either layout. Adding @andreasn here to see if he can shed some light on which is the most current layout.

I guess it's fine with the current design actually. Let's not nitpick here.

andreasn:
The verb _Start_ might feel a bit short compared to the verb _Troubleshoot_, so for that sake, it could use the verb _Start Service_ instead, even though it ends up being a bit redundant. But that's nitpicking and over-optimizing, and I would be OK merging as-is.

If I recall correctly, @martinpitt mentioned that enabling a systemd service doesn't only make it start automatically on boot only. This is why we call it "Automatically starts", without any mentioning of _boot_ in https://github.com/cockpit-project/cockpit/issues/11846. The term on the current service page is _Automatic startup_

KKoukiou:
> The verb _Start_ might feel a bit short compared to the verb _Troubleshoot_, so for that sake, it could use the verb _Start Service_ instead, even though it ends up being a bit redundant. But that's nitpicking and over-optimizing, and I would be OK merging as-is.
> 
> If I recall correctly, @martinpitt mentioned that enabling a systemd service doesn't only make it start automatically on boot only.

Well, systemctl enable x-service will only mark it for autostart at boot, if you use systemctl enable --now it will mark it for autostart at boot, but also start it immediately. 

This is why we call it "Automatically starts", without any mentioning of _boot_ in [cockpit-project/cockpit#11846](https://github.com/cockpit-project/cockpit/issues/11846). The term on the current service page is _Automatic startup_

andreasn:
I was convinced on IRC by several parties that we should keep _on boot_, so disregard the suggestion above.

pages/blueprints/index.js
```diff
@@ -40,9 +40,21 @@ const messages = defineMessages({
...
+  errorInactiveTitle: {
+    defaultMessage: "Image Building Service is Not Active"
+  },
+  errorInactivePrimary: {
+    defaultMessage: "Start"
+  },
+  errorInactiveSecondary: {
+    defaultMessage: "Troubleshoot"
+  },
+  errorInactiveCheckbox: {
```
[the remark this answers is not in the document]
jgiardino: These are really good points. I used "error" because this is the result of an error returned by the api. I used "inactive" because the error is returned when the service is not "active". I see your point about the variable name not reflecting the checkbox purpose. My thinking here is that all string variables that are used in the EmptyState component when this error is returned should be prefaced with the name of that state (e.g. "errorInactive") so that it's easy to see they're all related, and the rest of the string is just the UI element in that state (e.g. "Title", "Primary", "Checkbox") so that it's easy to visualize how these strings are being rendered.

pages/blueprints/index.js
```diff
@@ -114,6 +131,25 @@ class BlueprintsPage extends React.Component {
...
+    let argv;
+    if (this.state.enableService) {
+      argv = ["systemctl", "enable", "--now", "lorax-composer.socket"];
+    } else {
+      argv = ["systemctl", "start", "lorax-composer.socket"];
+    }
+    cockpit
+      .spawn(argv, { superuser: "require", err: "message" })
+      .then(() => this.props.fetchingBlueprints())
+      .catch(err => console.error("Failed to start lorax-composer.socket:", JSON.stringify(err)));
```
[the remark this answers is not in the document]
jgiardino: Makes sense. I'll work on this next...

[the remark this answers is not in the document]
martinpitt: In that case we shouldn't show the button at all, just the error message. We know this through the [cockpit.permission API](https://cockpit-project.org/guide/latest/cockpit-login.html).

martinpitt:
Looks by and large ok to me, aside from the design points that Lars and Katerina already mentioned. Thank you! This should also get a test case, but maybe @henrywang can work on that?

henrywang:
@martinpitt, sure I'll add test case on this feature when this CR got landed. What I'm concerned is about error message if lorax-composer starts fail or what if lorax-composer does not installed. We need to make them different.

pages/blueprints/index.js
```diff
@@ -147,9 +183,39 @@ class BlueprintsPage extends React.Component {
           handleShowModalManageSources={this.handleShowModalManageSources}
         />
         {(blueprintsLoading === true && <Loading />) ||
-          ((blueprintsError !== null && (
-            <EmptyState title={formatMessage(messages.errorTitle)} message={blueprintsError.message} />
-          )) ||
+          ((blueprintsError !== null &&
+            ((blueprintsError.message === "not-found" && (
+              <EmptyState title={formatMessage(messages.errorInactiveTitle)} icon="fa fa-exclamation-circle">
+                <>
```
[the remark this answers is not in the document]
jgiardino: This should be updated now.

core/reducers/blueprints.js
jgiardino: I refactored the reducers for `blueprints` so that separate reducers control the different things inside the `blueprints` state. I was kinda following this example, where one action can be used by more than one reducer: https://egghead.io/lessons/javascript-redux-displaying-error-messages

pages/blueprints/index.js
```diff
@@ -147,9 +183,39 @@ class BlueprintsPage extends React.Component {
           handleShowModalManageSources={this.handleShowModalManageSources}
         />
         {(blueprintsLoading === true && <Loading />) ||
-          ((blueprintsError !== null && (
-            <EmptyState title={formatMessage(messages.errorTitle)} message={blueprintsError.message} />
-          )) ||
+          ((blueprintsError !== null &&
```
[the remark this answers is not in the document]
jgiardino: I will leave this change for a future update. 😄 (we need a better message for when the user doesn't have permission to use image builder)

jgiardino:
I think I've addressed all the comments. Please let me know if there's anything else I should address.

> What I'm concerned is about error message if lorax-composer starts fail or what if lorax-composer does not installed. We need to make them different.

@henrywang Yes, I agree that we need better messaging if lorax-composer is not installed. I'll save that for a separate PR. Regarding what if lorax-composer start fails, these updates will display an inline Error message above the EmptyState if an error is returned after the user clicks Start (like the example below):
image

And then there was a suggestion on IRC yesterday to disable the Start action and display a tooltip if the user doesn't have permission to start. Just fyi that the default state is to enable the action, in case getting the user permissions fails, because even if the user doesn't have permission and is able to click the button, then an error like the one above will display. 
image

henrywang:
@KKoukiou, sure, I'll push test to @jgiardino's branch. I hope I can get it done in this week. :-)

henrywang:
@jgiardino, @KKoukiou: The Manage Sources dropdown item should be disabled if lorax-composer is not started.

jgiardino:
> The Manage Sources dropdown item should be disabled if lorax-composer is not started.

Thanks for catching that, @henrywang! That should be fixed now.

henrywang:
@jgiardino, @KKoukiou: I tried one scenario, the repo cannot get fetched due to internet or repo issue when starting lorax-composer. 
lorax-composer error log:
`Could not resolve host: xxx.xxx.xxx.xxx; Unknown error"`
The UI reports an unrelated error. Could you please take a look if you have time? Thank you!
Screenshot_2019-05-23 Image Builder - localhost localdomain

jgiardino:
> lorax-composer error log:
Could not resolve host: xxx.xxx.xxx.xxx; Unknown error"
The UI reports an unrelated error.

The message "Server has closed the connection" is the message that's provided by the api. That's odd that this message is different from the one in the log. Would that be an issue for the api to address? Or should the UI provide an option to display the log? Either way, I don't think that's something that can be handled with this PR, but definitely worth following up on.

jgiardino:
And thanks for adding the tests, @henrywang! 👍

test/end-to-end/specs/viewBlueprint.test.js
```diff
@@ -14,6 +14,7 @@ describe("View Blueprint Page", function() {
 
   before(function() {
     commands.login();
+    commands.startLoraxIfItDoesNotStart();
```
KKoukiou: You have added this extra step to enable and start lorax-composer in each test.
If that single component will break, all tests will break in this way.
For this reason it would make more sense to have it by default enabled, and don't enable/start it from the UI for all the tests, apart from the single test that it's checking this feature. And for that one you could stop it with 'systemctl stop' in the setUp of the test.

henrywang: @KKoukiou appreciate your suggestion! :+1: 
The "start lorax-composer" is only got run with the first test, in this case it's api.test.js. The rest of tests will not run this extra step. If I add this test, i'll add it in api.test.js. So It's almost the same thing for the test. The only reason I choose this solution is saving one more step - stop lorax-composer.

henrywang:
Re-run rhel-7-7 after #677 got merged.
It's green now.
Thanks @jgiardino, @KKoukiou. Merged!

PR #641, opened 2019-04-18
po: Add cleanup commands

marusak:
New version of po-refresh bot also needs `clean-po` command.
This command is used for dropping languages that do not have sufficient
translation coverage anymore - currently all languages that have at least
one translated word in welder-web are pulled. So non languages will drop
anyway, but if once introduced minimal translations rule then languages
could be dropped (but seems that zanata-js does not support min-translation)

See also: https://github.com/cockpit-project/cockpit/pull/11638

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,8 @@ bots:
 update-po:
 upload-pot: po-push
 download-po: po-pull
+clean-po:
+	rm po/*.po
 
 .PHONY: tag local-clean vm check devel-install
 


marusak:
Also let's keep this open to test https://github.com/cockpit-project/cockpit/pull/11581

marusak:
I cannot ask for review the "normal" way, so @martinpitt can you please review? And then retrigger https://github.com/weldr/welder-web/issues/645, which failed because of this PR has not yet landed.

PR #645, opened 2019-04-25

cockpituous:
 * [x] po-refresh


marusak:
`po: Add 'sk' language`
ou yes :smile:

PR #650, opened 2019-04-26
test: Add test case for manage sources page

henrywang:
1. Add data-* and id attributes to make search page element easy.
2. Add test case for manage sources page.

### DIFF
--- a/components/ListView/SourcesListItem.js
+++ b/components/ListView/SourcesListItem.js
@@ -38,7 +38,7 @@ class SourcesListItem extends React.Component {
     const type = source.type.substring("yum-".length);
     const { formatMessage } = this.props.intl;
     return (
-      <div className="list-pf-item" key={key}>
+      <div className="list-pf-item" key={key} data-source={source.name}>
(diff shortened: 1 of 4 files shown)


henrywang:
The failed case on edge is due to issue #651

martinpitt:
@henrywang : Would it be possible to skip this new test on Edge until this gets fixed properly? Then this feature is at least covered on the other two browsers, to avoid regressing there.

henrywang:
@martinpitt, cloud you please do not merge this PR until the rhel-8-1 failed issue gets fixed. Thanks!

henrywang:
@martinpitt, the add source in rhel-8-1 has been resolved. I think we can land this PR. Thanks.
The copy and paste test on edge still got failed, looks the Crtl-V does not work on edge sometimes.

martinpitt:
@henrywang : I retried three times, edge just doesn't want to get green :-(

henrywang:
@martinpitt, it's failed on copy and paste case. That's what I need #694 to debug copy and paste case.

Issue #631, opened 2019-04-10
openssl passwd in rhel 7 does not support sha512 based password algorithm


henrywang:
Adding user account does not work on rhel 7 because openssl passwd in rhel 7 does not support sha512 based password algorithm.

https://github.com/weldr/welder-web/blob/12584fb192dd4ce9f8f874e07d5ebf37e98c6fed/components/Modal/UserAccount.js#L185

`openssl passwd` output:
```shell
[root@m1 ~]# openssl passwd -6 -stdin
Usage: passwd [options] [passwords]
where options are
-crypt             standard Unix password algorithm (default)
-1                 MD5-based password algorithm
-apr1              MD5-based password algorithm, Apache variant
-salt string       use provided salt
-in file           read passwords from file
-stdin             read passwords from stdin
-noverify          never verify when reading password from terminal
-quiet             no warnings
-table             format output as table
-reverse           switch table columns
```

martinpitt:
Unfortunately it's ridiculously complicated to encrypt a password on the command line; `openssl passwd` is pretty much the only way; `chpasswd` knows how to encrypt, but only operates on the /etc/shadow file, not on stdout. There are a [few ways using python or perl](https://serverfault.com/questions/330069/how-to-create-an-sha-512-hashed-password-for-shadow), but these all mean additional dependencies, or mucking around with python2-only vs. python3-only distros.

With that much trouble, I think it would actually make more sense to do it on the JavaScript side with [bcrypt](https://www.npmjs.com/package/bcrypt).

martinpitt:
Fixed in PR #661, just got wrongly referenced in the commit message. (@henrywang : This should be "Fixes #xxxx")

PR #661, opened 2019-05-07
Fix "passwd in rhel 7 does not support sha512" issue

henrywang:
Address issue [613](https://github.com/weldr/welder-web/issues/631)
Use python to generate sha512 hashed password in all platforms

### DIFF
--- a/components/Modal/UserAccount.js
+++ b/components/Modal/UserAccount.js
@@ -182,7 +182,12 @@ class UserAccount extends React.Component {
   }
 
   encryptPassword(password) {
-    return cockpit.spawn(["openssl", "passwd", "-6", "-stdin"], { err: "message" }).input(password);
+    return cockpit
+      .script(
+        "$(which /usr/libexec/platform-python 2>/dev/null || which python3 2>/dev/null || which python) -c 'import sys, crypt; print(crypt.crypt(sys.stdin.readline().strip(), crypt.mksalt(crypt.METHOD_SHA512)))'",
+        { err: "message" }
+      )
+      .input(password);
   }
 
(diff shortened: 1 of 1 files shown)


components/Modal/UserAccount.js
```diff
@@ -182,7 +182,17 @@ class UserAccount extends React.Component {
...
-    return cockpit.spawn(["openssl", "passwd", "-6", "-stdin"], { err: "message" }).input(password);
+    // inspired by https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/python.js
+    const pyinvoke = [
+      "sh",
+      "-ec",
+      "exec $(which /usr/libexec/platform-python 2>/dev/null || which python3 2>/dev/null || which python) $@",
+      "--",
+      "-",
+      "-c"
+    ];
```
martinpitt: This is unsafe now -- the password can legitimately contain `"`, or newlines and such. This wasn't a problem with feeding the password through stdin. I think it's probably better to keep that behaviour, and don't feed the Python source code through stdin, but on the CLI instead. This is easier than with python.js, as you have complete control over the entire command line. Can you please try with

```js
return cockipt.script(
    "$(which /usr/libexec/platform-python 2>/dev/null || which python3 2>/dev/null || which python) -c 'import sys, crypt; print(crypt.crypt(sys.stdin.readline().strip(), crypt.mksalt(crypt.METHOD_SHA512)))' ",
    { err: "message" }).input(password);
```

henrywang: Thank you so much @martinpitt! :+1: :+1:  Updating now.

PR #667, opened 2019-05-10
test: use real image building API to build image

henrywang:
1. move image building test to a separate file. So the tier-x test can ignore this test by excluding this file. 
2. add all configurable settings before build image to increase test coverage
3. remove fedora repos from rhel-7-7 test scenario because we do not need them


martinpitt:
Nice! Crossing fingers that this works.

glibc seems a strange choice, as there's no way to *not* have it installed -- it should be in every standard image already? How about "zsh" or a similarly small-but-nonstandard package?

test/end-to-end/specs/viewBlueprint.test.js
```diff
@@ -220,6 +220,8 @@ describe("View Blueprint Page", function() {
       });
 
       it('should show "Complete"', function() {
+        // set test timeout to 30 minutes for image building case ONLY
+        this.timeout(1800000);
         expect(viewBlueprintPage.completeLebel.getText()).to.equal("Complete");
```
martinpitt: This now builds an actual image, but that doesn't get verified at all. As an initial step, I suggest to ensure that the file exists at the expected place, and that running `qemu info` on it gives the expected virtual size.

martinpitt: To clarify: As a further step, the test should actually boot the generated image, with qemu. For that it needs to install and enable some network-accessible service ssh, so that we can validate from outside that the image boots and works, and even login and inspect it. But this is fine for a follow-up PR.

henrywang:
@martinpitt, I think we can merge this PR. I tried 3 full rounds of tests, all of them were passed. The edge failed issue is not related in this case. I need land this PR to add test cae for #678. Thanks!

test/end-to-end/specs/viewBlueprint.test.js
martinpitt: I'm still convinced that there should be *some* validation of the created image, to cover a complete end-to-end workflow. But this can happen in a follow-up.

PR #678, opened 2019-05-24
Add link to show the anaconda.log from the compose build process

KKoukiou:
Especially when some image build failed it's usefull to be able to see
from the composer UI the anaconda logs.
This commit enables this functionality, by having a link for showing the
logs in each image.

### DIFF
--- a/components/ListView/ListItemImages.js
+++ b/components/ListView/ListItemImages.js
@@ -9,14 +9,17 @@ import {
   setModalDeleteImageVisible,
   setModalDeleteImageState
 } from "../../core/actions/modals";
+import * as composer from "../../core/composer";
 
 class ListItemImages extends React.Component {
   constructor() {
     super();
+    this.state = { logsExpanded: false };
     this.handleDelete = this.handleDelete.bind(this);
     this.handleCancel = this.handleCancel.bind(this);
     this.handleShowModalStop = this.handleShowModalStop.bind(this);
     this.handleShowModalDeleteImage = this.handleShowModalDeleteImage.bind(this);
+    this.handleLogsShow = this.handleLogsShow.bind(this);
   }
 
   // maps to Remove button for FAILED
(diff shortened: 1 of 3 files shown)


public/custom.css
```diff
@@ -609,3 +609,8 @@ textarea.form-control[readonly] {
   background-color: inherit;
   color: inherit;
 }
+
+.show-logs-button {
```
KKoukiou: This is ugly. Not sure how to make it nice though.

components/ListView/ListItemImages.js
```diff
@@ -43,10 +46,43 @@ class ListItemImages extends React.Component {
     this.props.setModalDeleteImageVisible(true);
   }
 
+  handleLogsShow() {
+    this.setState(prevState => ({ logsExpanded: !prevState.logsExpanded, fetchingLogs: !prevState.logsExpanded }));
+    fetchComposeLogApi(this.props.listItem.id).then(
+      logs => {
+        this.setState({ logsContent: logs, fetchingLogs: false });
+      },
+      () => {
```
KKoukiou: I totally hate this style, but eslint rules of this project was forcing me to have this syntax.

KKoukiou:
@jgiardino currently it looks like this.
I am not super satisfied but I couldn't came up with a better design. Can you maybe think of some better solution?
logs-anaconda

components/ListView/ListItemImages.js
KKoukiou: Normally we would want to catch the exception thrown by the API here and print the message. However the error message from the API is useless, so at this point this doesn't make any sense.

components/ListView/ListItemImages.js
```diff
@@ -43,10 +46,43 @@ class ListItemImages extends React.Component {
...
+
   render() {
     const { listItem } = this.props;
     const timestamp = new Date(listItem.job_created * 1000);
     const formattedTime = timestamp.toDateString();
+    const moreButton = (
+      <a className="show-logs-button" role="presentation" onKeyDown={this.handleLogsShow} onClick={this.handleLogsShow}>
+        {this.state.logsExpanded && <FormattedMessage defaultMessage="Hide logs" />}
+        {!this.state.logsExpanded && <FormattedMessage defaultMessage="Show logs" />}
+      </a>
```
larskarlitski: I'm getting a `text` cursor instead of a `pointer` here.

public/custom.css
larskarlitski: Agreed. Let's wait for @jgiardino's design input before fixing it, though.

components/ListView/ListItemImages.js
larskarlitski: Yeah this is fine.

KKoukiou:
@jgiardino before I rebase and fix Lars' comment can you give me a review as well?

components/ListView/ListItemImages.js
jgiardino: Try using `<button class="btn btn-link" ...>` instead of `<a className="show-logs-button" role="presentation" ... >` here. The `.btn.btn-link` element will have more padding around it, and you could remove the custom class and css. And it also has the correct cursor.

KKoukiou: Ah cool, thanks for the hint! Otherwise are you satisfied with the placement of thhis 'show more' button/link?

Do you think it's belongs next to the status?

jgiardino: Plus, a `<button>` is semantically more correct than an `<a>` element in this case (e.g. if an `<a>` doesn't have an `href` value, then use a button instead)

jgiardino:
This is great! Thanks for adding this! After spending some time thinking about the layout of this view, and how it might evolve over time, I'd say what you have with the "Show logs" and "Hide logs" toggle is a good choice. And it seems right that it would be adjacent to the status. 

One refinement that would be nice to add is a sticky position to the row so that the ability to click "Hide logs" is always available after the user has scrolled down.

In the example below, I had added these styles to `.list-pf-container`:
```
    position: sticky;
    top: 0;
    background: #fff;
```
Although these could probably use some refinement, too (like a border that's visible between the list item and log), but that refinement could happen now or later.

image

components/ListView/ListItemImages.js
```diff
@@ -43,10 +46,43 @@ class ListItemImages extends React.Component {
...
 
+  handleLogsShow() {
+    this.setState(prevState => ({ logsExpanded: !prevState.logsExpanded, fetchingLogs: !prevState.logsExpanded }));
+    fetchComposeLogApi(this.props.listItem.id).then(
+      logs => {
+        this.setState({ logsContent: logs, fetchingLogs: false });
+      },
+      () => {
+        this.setState({
+          logsContent: <FormattedMessage defaultMessage="No anaconda log available" />,
```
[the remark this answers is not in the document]
KKoukiou: Ups I forgot this comment, repushing.

core/composer.js
```diff
@@ -188,3 +191,7 @@ export function getFinishedComposes() {
 export function getFailedComposes() {
   return get("/api/v0/compose/failed").then(data => data.failed);
 }
+
+export function getComposeLog(uuid) {
+  return get("/api/v0/compose/log/" + encodeURIComponent(uuid), undefined, { format: "raw" });
```
KKoukiou: @larskarlitski that's why I had all args to get to be named args before, to avoid this undefined variable here.

larskarlitski:
Update composer.js API, as discussed on irc. Sorry again for the back and forth.

PR #681, opened 2019-05-27
Pin babel-plugin-react-intl to 3.0.1

larskarlitski:
3.0.2 is broken for us because of

    https://github.com/formatjs/babel-plugin-react-intl/pull/165

`formatMessafe()` calls are not recognized as coming from injectIntl().

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "babel-eslint": "^10.0.1",
     "babel-loader": "^8.0.5",
     "babel-plugin-istanbul": "^5.1.0",
-    "babel-plugin-react-intl": "^3.0.1",
+    "babel-plugin-react-intl": "3.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "history": "3.3.0",
     "jquery": "3.4.1",
     "patternfly": "3.59.1",
-    "patternfly-react": "1.19.1",
+    "patternfly-react": "2.34.3",
```
martinpitt: It's a bit strange to update this dependency in this commit - it should be split out.

larskarlitski: Oops, this snuck in from a `npm audit fix`. I removed it completely from the PR.

larskarlitski:
Updated to remove the erroneous version change. Thanks for the review!

PR #683, opened 2019-05-28
Properly encode URLs (and refactor api call logic)

larskarlitski:
I only wanted to ensure that URLs are encoded properly, but I accidentally refactored quite a bit of the API call functions. I think it is more consistent now and reads much better.

Please see individual commits.


core/apiCalls.js
```diff
@@ -108,18 +108,7 @@ export function commitToWorkspaceApi(blueprint) {
...
-      .apiFetch("/api/v0/blueprints/diff/" + blueprintId + "/NEWEST/WORKSPACE")
-      .then(data => {
-        resolve(data);
-      })
-      .catch(e => {
-        console.log("Error fetching diff", e);
-        reject();
-      });
-  });
-  return p;
```
KKoukiou: Just for clarity, here they were calling reject inside the catch, so there was not an issue of not propagating the exception up. But I agree that it's useless and well removed.

data/BlueprintApi.js
```diff
@@ -1,88 +1,48 @@
...
-          console.log(`Error committing blueprint: ${e}`);
-          NotificationsApi.closeNotification(undefined, "committing");
-          NotificationsApi.displayNotification(blueprint.name, "commitFailed");
-          reject();
-        });
-    });
-    return p;
-  }
-
-  postedBlueprintData(blueprint) {
```
jgiardino: This is still being used in sagas/blueprints.js, (e.g. `BlueprintApi.postedBlueprintData`).

jgiardino:
These updates are great! Thanks for making them. I only had one comment above so far but will continue reviewing later today.

data/BlueprintApi.js
larskarlitski: Oops. Thanks for the catch. I've uploaded a fix.

PR #686, opened 2019-05-29
Simplify export blueprint modal dialog

KKoukiou:


KKoukiou:
@henrywang some tests are failing, I really can't understand from the logs what's going on because there is not exact line of the test failing.
From manual it works as expected so if you don't mind I will leave it up to you to give me some hints how to make tests pass.

henrywang:
@KKoukiou, all GREEN now! Two issues failed your test:
1. the new code update dropdown menu item from `Export` to `Export Blueprint`.
2. the new code makes `span=Close` not unique, it matches both Close button and X button at the up right corner of Export Blueprint dialog.

Here's how I fix them:
1. easy to fix, just update selector.
2. add `data-btn` in `components/Modal/ExportBlueprint.js` to make `Close` button unique.

components/Modal/ExportBlueprint.js
```diff
@@ -23,110 +39,111 @@ class ExportBlueprint extends React.Component {
...
-                    </p>
-                  </div>
+      <React.Fragment>
+        <a href="#" onClick={this.open}>
+          <FormattedMessage defaultMessage="Export Blueprint" />
+        </a>
+        <Modal show={this.state.showModal} onHide={this.close} id="cmpsr-modal-export">
+          <Modal.Header>
+            <Modal.CloseButton onClick={this.close} />
+            <Modal.Title>
```
jgiardino: Oh, I missed this in my review just now...

Can you add an `id` here, similar to what's used in Manage Sources, like `id="title-export"?
And then add `aria-labelledby="title-export"` to the `<Modal ... >`?

These updates provide an accessible name for the modal for screen readers.

KKoukiou:
> @KKoukiou, all GREEN now! Two issues failed your test:
> 
>     1. the new code update dropdown menu item from `Export` to `Export Blueprint`.

This was renamed by mistake.

Bringing it back to it's original title.

> 
>     2. the new code makes `span=Close` not unique, it matches both Close button and X button at the up right corner of Export Blueprint dialog.
> 
> 
> Here's how I fix them:
> 
>     1. easy to fix, just update selector.
> 
>     2. add `data-btn` in `components/Modal/ExportBlueprint.js` to make `Close` button unique.

components/Modal/ExportBlueprint.js
KKoukiou: @jgiardino thanks, fixed.

However these should all be done by patternly-react I believe.

PR #699, opened 2019-06-12
cockpituous: Fix srpm Release:

martinpitt:
Use release-srpm's `-V` option [1] so that the generated srpm will get a
proper changelog and Release "1" instead of "2".

[1] https://github.com/cockpit-project/cockpituous/commit/abb2bdb5


martinpitt:
The previous few releases I didn't realize this mistake, so I had to do the bodhi releases manually (as the release machinery expected "Release: 2"), and the current Fedora srpm has a [poor changelog](https://src.fedoraproject.org/rpms/cockpit-composer/blob/master/f/cockpit-composer.spec).

I locally ran cockpituous release scripts with this change, and it looks fine now.

PR #700, opened 2019-06-14
Use cockpit's PatternFly CSS

martinpitt:
Cockpit's patternfly.css already sets the correct font paths, and does
not have any other modifications right now (Cockpit specific
modifications live in the separate cockpit.min.css).

This will automatically pick up future adjustments to Cockpit PatternFly
styling, such as in https://github.com/cockpit-project/cockpit/pull/11987,
nd thus ensure that all Cockpit pages will have a consistent look and
feel.

This also avoids shipping the PatternFly css and fonts and thus makes
the package much smaller.

 - [x] Land cockpit PF4 update: https://github.com/cockpit-project/cockpit/pull/11987
 - [x] Fix kebabs and other UI elements in Cockpit's PF: https://github.com/cockpit-project/cockpit/pull/12211
 - [ ] Test all pages and add composer-local fixes

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "fastclick": "1.0.6",
     "history": "3.3.0",
     "jquery": "3.4.1",
-    "patternfly": "3.59.3",
     "patternfly-react": "1.19.1",
     "prop-types": "15.7.2",
     "react": "16.8.6",
--- a/public/custom.css
+++ b/public/custom.css
@@ -47,6 +47,7 @@
 }
 .cmpsr-panel__body--main {
   grid-area: edit;
+  background: var(--pf-global--BackgroundColor--100);
 }
 /* Blueprint Edit panel contents */
 .cmpsr-grid__wrapper .cmpsr-panel__body {
@@ -139,6 +140,7 @@
 }
 .cmpsr-title__item {
   margin: 0 10px 0 0;
+  font-size: var(--pf-global--FontSize--2xl);
 }
 
 /* List view */
@@ -303,6 +305,10 @@ the "-" that would display instead of the sr-only "to" */
   top: 10px;
   margin: 0 20px;
 }
+.label-info {
+  background-color: var(--pf-global--link--Color);
+}
+
 /* Expanded list item contents (i.e. component summary) */
 .list-pf-container .list-pf {
   max-height: 300px;
@@ -317,9 +323,9 @@ the "-" that would display instead of the sr-only "to" */
 /*Component details*/
 .pf-icon-small {
     font-size: 1.7rem;
-    height: 30px;
-    line-height: 30px;
-    width: 30px;
+    height: 2.2rem;
+    line-height: 2.2rem;
+    width: 2.2rem;
     display: -webkit-inline-box;
     display: inline-flex;
     -webkit-box-pack: center;
@@ -580,9 +586,12 @@ body {
 }
 
 /* nav-tabs-pf */
- .nav-tabs .badge {
-   display: inline;
- }
+#blueprint-tabs .nav-tabs {
+  margin-top: var(--pf-global--spacer--sm);
+}
+.nav-tabs .badge {
+  display: inline;
+}
 
 /* disabled state */
 /* disable pointer events for both <a> and <button> */
--- a/public/custom.css
+++ b/public/custom.css
@@ -32,9 +32,11 @@
 .cmpsr-panel__title--sidebar {
   grid-area: inputsTitle;
   margin-right: -1px;
+  margin-left: 20px;
 }
 .cmpsr-panel__title--main {
   grid-area: editTitle;
+  margin-right: 20px;
 }
 .cmpsr-panel__body {
   border-width: 0 1px 1px;
@@ -44,10 +46,12 @@
 .cmpsr-panel__body--sidebar {
   grid-area: inputs;
   margin-right: -1px;
+  margin-left: 20px;
 }
 .cmpsr-panel__body--main {
   grid-area: edit;
   background: var(--pf-global--BackgroundColor--100);
+  margin-right: 20px;
 }
 /* Blueprint Edit panel contents */
 .cmpsr-grid__wrapper .cmpsr-panel__body {
@@ -623,4 +627,12 @@ textarea.form-control[readonly] {
   position: sticky;
   top: 0;
   background: #fff;
+}
+
+/* PF4 styles for PF3 classes */
+body {
+  background: var(--pf-global--BackgroundColor--100);
+}
+.container-fluid {
+  padding: 0 20px;
 }
--- a/public/custom.css
+++ b/public/custom.css
@@ -144,6 +144,8 @@
 }
 .cmpsr-title__item {
   margin: 0 10px 0 0;
+}
+h1.cmpsr-title__item {
   font-size: var(--pf-global--FontSize--2xl);
 }
 
@@ -635,4 +637,11 @@ body {
 }
 .container-fluid {
   padding: 0 20px;
+}
+
+.fields-status-pf {
+  color: var(--pf-global--Color--200);
+}
+.text-muted {
+  color: var(--pf-global--Color--200);
 }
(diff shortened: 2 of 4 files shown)


martinpitt:
When running against currently released cockpit, this makes no difference. But when running against the [PF4 styling branch](https://github.com/cockpit-project/cockpit/pull/11987#issuecomment-501663275), it will automatically adopt to the new style, and at first sight it looks just fine:

bplist

bpedit

bpuser

martinpitt:
So this is at least useful to explore Composer's UI with a PF4 look and feel, and discover CSS bugs.

I still have a bad gut feeling about this, though -- it seems to me that long-term this is more brittle than necessary. I don't yet know how this will behave with PF4 components once we start migrating them one by one -- if they still require a global patternfly.css `<link>` or whether this will be handled by local CSS via webpack.

So long-term I think we should copy cockpit's PF CSS adjustments and build them ourselves in Composer. I'll work on such a branch next.

@larskarlitski , @jgiardino , do you have an opinion about this?

martinpitt:
@garrett: FYI

martinpitt:
On second thought, cockpit-composer just takes the verbatim ./node_modules/patternfly/dist/css/patternfly.min.css and ships it as-is. Cockpit does the same plus:

 - [adjusting font paths](https://github.com/cockpit-project/cockpit/blob/master/tools/patternfly.sed) (which is fine for Composer),
 - and soon, the the PF4 look (which we want for composer)

and nothing else. All other customizations are *not* shipped via patternfly.css, but via cockpit.min.css.

Also, this patternfly.css will always stay PF3 (PF4 has its own independent files), so what we do about PF4 widgets is not influenced by the decision here.

After these investigations, I retract my claim that this is more brittle, and it's actually fairly elegant. It even avoids shipping a duplicate patternfly.css and the fonts, i. e. makes cockpit-composer smaller.

jgiardino:
Thanks for doing this, @martinpitt! 

Just looking at the examples you posted, there are a few issues I notice. I'm guessing these are related to either custom css that I included or pf3 css that I am using that cockpit is not. I don't imagine these issues would be difficult to fix, and I'm happy to work on these.

Would it make sense for me to pull these changes and make further updates for these issues as part of this PR?

martinpitt:
@jgiardino : by "examples I posted" you mean the screenshots from the PF4 look? Indeed there are still a lot of little bugs with that, see the long TODO list in https://github.com/cockpit-project/cockpit/pull/11987. These screenshots merely demonstrated that with this PR c-composer will adjust to cockpit's patternfly.css. However, the PF4 look will still take some time to land, and part of it will be to also validate composer.

This PR by itself should not change anything in the UI really. Did you notice any visual regressions there? I didn't. This PR would merely pave the way so that we *can* move to the PF4 look.

So we mostly need to decide whether we conceptually want to share Cockpit's PF CSS with cockpit-{ostree,podman,composer} (the former two already do), or whether cockpit-composer wants to retain full control of its own PatternFly, and then update to the PF4 look in lockstep.

martinpitt:
As far as I understood @larskarlitski and @jgiardino , we want to go with this approach, but postpone it until after the cockpit changes are done and land. Then we can add any possibly required fixups here, and don't risk having an intermediate broken layout. Did I get this right?

Marking as blocked, and added a TODO  list.

larskarlitski:
> As far as I understood @larskarlitski and @jgiardino , we want to go with this approach, but postpone it until after the cockpit changes are done and land. Then we can add any possibly required fixups here, and don't risk having an intermediate broken layout. Did I get this right?

Indeed. Thanks!

martinpitt:
Unblocking, as the PF4 PR landed now. Force-pushed to fix the trivial conflict.

I can't build cockpit-composer right now due to dozens of errors like
```
/home/martin/upstream/cockpit-composer/components/ListView/ComponentDetailsView.js
   76:19   error  'component.builds' is missing in props validation                    react/prop-types
```
I'll investigate that now.

martinpitt:
I was finally able to build locally with PR #710. I updated the screenshots above. For sure the kebab vs. button alignment/spacing/frames need work, but at least it's usable. Someone with a more trained "pixel error" eye than mine now needs to go through with a finer comb.

garrett:
Oh, it's looking pretty good! Yeah, the kebab button needs work. I know where to change it and how to change it; I should do that.

There are some widgets (like kebab) that are used in Composer but not in Cockpit; these would need to be added to the overrides file (in this case, integrated into the existing button/dropdown CSS). There are probably other widgets and groups of widgets which might also not be covered yet too.

I need to get Composer working locally. :wink:

martinpitt:
> I need to get Composer working locally.

I just did `dnf install lorax-composer`, and then I can just run cockpit-composer out of the build tree. What breaks for you?

So you are saying we should fix the kebabs straight in Cockpit's PF overrides, not in composer's CSS? That makes sense indeed. Thanks for looking into it!

garrett:
@martinpitt: When I try to run Composer, this is what I see:

Screenshot_2019-07-03 Image Builder - Sunny

As to not have a bunch of noise (from extensions and whatnot), I also attempted this in Epiphany and checked the console there. This is what I see:

composer-error

garrett:
Here's the console from Chrome:

chrome-console

jgiardino:
Thanks for making these updates, @martinpitt! They look amazing, @garrett! Aside from the kebab, these are some other things I noticed in the screencaps above (but still need to review across all the pages):

- List view icon circle isn't sized right relative to the icon. This is based on the [PF3 list view](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-view-rows.html)
  image

- On the edit blueprint page, in the Blueprint Components, the select menu before the filter input is behind the filter input:
  image
  and what it looks like in PF3:
  image

martinpitt:
Force-pushed to pick up the empty state error message fix, so that this branch doesn't have regressions.

garrett:
I've fixed most of the above issues — and several other small ones that affect Composer — in https://github.com/cockpit-project/cockpit/pull/12211.

The notable exception is the list view icon circle. I'll have to address it next week in another round of fixups. (There are a few other small ones I ran out of time to address this week.)

martinpitt:
I updated the screenshots for https://github.com/cockpit-project/cockpit/pull/12211 . I'll merge this later tonight, and then build updated RPMs for @jgiardino . Thanks @garrett!

garrett:
I just noticed there's a bug with the dropdowns when they have a selected state. I'll have to fix next week. (On Monday.) It's a very simple fix, but it's EOB/EOW for me now. We'll get it fixed before release. :wink:

martinpitt:
@jgiardino : I updated the RPMs in https://piware.de/tmp/cpf4/ to include the latest bug fixes, if you want to play around with it.

martinpitt:
@jgiardino : I updated the RPMs in https://piware.de/tmp/cpf4/ again with the latest of @garrett's fixes. Do you have some time today to test this? It would be nice to do a composer release with this tomorrow, so that it updates in lockstep with cockpit in Fedora 30.

martinpitt:
@jgiardino : You can test these in your RHEL 8.1 VM with:
```
wget -r -np https://piware.de/tmp/cpf4/
rpm -F --verbose piware.de/tmp/cpf4/*.rpm
systemctl stop cockpit
```

After that, `rpm -qa cockpit*` should show all 197.x-* versions.

martinpitt:
Copying @jgiardino's comment from PR #725 here for having a full record here. @garrett, these may be interesting to apply to Cockpit itself as well?

[I] add[ed] a few miscellaneous css updates in the context of Composer custom css updates.

There were a couple of items provided in the cockpit css that didn't have sufficient color contrast. I updated those here, but it might be worth checking if these are used in cockpit, and if the color contrast issues are present there as well.
```
.fields-status-pf { 	
  color: #8b8d8f;
  margin-bottom: 15px;
}
```
 I updated this in composer to `var(--pf-global--Color--200);`

```
.label-info { 
  background-color: #73bcf7;
}
```
 I updated this in composer to `var(--pf-global--link--Color)`

jgiardino:
Thanks, @martinpitt!

bitmoji

PR #712, opened 2019-07-03
Fix crash if no language is set

martinpitt:
If `cockpit.language` is undefined (which can happen in practice), just
default to English instead of crashing on calling
`undefined.includes()`.


martinpitt:
This helped @garrett to be able to run composer.

Issue #689, opened 2019-05-31
Refreshing the page when already inside a blueprint will reset the image list to empty


KKoukiou:
### Description of problem
When inside a specific blueprint which has images, if you refresh the browser page the image list will be incorrect when going to the images tab.

### Component (web, API, etc)
web

### How often reproducible
always

### Steps to Reproduce

1. Go to http://localhost:9090/welder#/blueprint/example-development and create some image
2. Refresh page
3. Go to images tab and see that there are not images

### Actual results
Image list appears empty

### Expected results
Correct image list after refreshing the page.

PR #713, opened 2019-07-03
Fix images not loading on refresh

jkozol:
This fixes issue #689. 

The fetchComposeTypes saga was blocking other compose sagas from running
on a page refresh. Instead of default calling fetchComposeTypes on the
application loading this saga is now called by a dispatched action.

### DIFF
--- a/components/ListView/BlueprintListView.js
+++ b/components/ListView/BlueprintListView.js
@@ -12,7 +12,7 @@ class BlueprintListView extends React.PureComponent {
   }
 
   render() {
-    const { blueprints, imageTypes, layout } = this.props;
+    const { blueprints, layout } = this.props;
     return (
       <div className="list-group list-view-pf list-view-pf-view">
         {blueprints.map(blueprint => (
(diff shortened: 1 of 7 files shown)


jkozol:
It does make more sense to have it called when the modal is accessed and will reduce code clutter. It should be quick to refactor so I will do that before merging.

jgiardino:
Thanks for fixing this issue! The images tab works as expected now.

When I was jumping around between pages, I have a blueprint that isn't depsolving. On the Edit Blueprint page, the following error displays. But when I look at the latest that's on master, the error message that's provided was getting surfaced. In this case, the error message that's provided is "example-http-server: The following package(s) had problems: openssh-server-7.* (No match)." Can you check that this message wasn't affected in your updates? 

image

jgiardino:
Regarding my previous comment, I don't see where in your updates the BLUEPRINT_CONTENTS_FAILURE action would be affected. But in the inspector, I can see that this action is providing the error message, but when I click on the "Diff" tab, it isn't showing any difference in state with this action. And I'm fairly certain I was seeing this message as expected before pulling these changes. 🤷‍♀

Issue #646, opened 2019-04-25
Source repo deletion should have a notification


henrywang:
How to reproduce:
1. Add a new source.
2. Clicking Delete for new added source.
3. The source is still there for at least 2 seconds until it get deleted from UI.

There should be a notification when deleting source. That will tell user it's deleting now but not finished.

jkozol:
Closed by PR #714

PR #714, opened 2019-07-03
Add notification for source repo deletion

jkozol:
This fixes issue #646

After deleting a source the options to edit  will be replaced by a
spinner and text saying "Deleting source".

removingSourceScreenshot

### DIFF
--- a/components/ListView/SourcesListItem.js
+++ b/components/ListView/SourcesListItem.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { FormattedMessage, defineMessages, injectIntl, intlShape } from "react-intl";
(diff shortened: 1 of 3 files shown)


KKoukiou:
@henrywang this needs a test please.

jkozol:
Oops thanks for the catch. I will fix the commit message. So I should include "Fixes #689" directly in the commit message?

KKoukiou:
> Oops thanks for the catch. I will fix the commit message. So I should include "Fixes #689" directly in the commit message?

No need. We can edit the commit message inline before merging . Tests passed so let's keep the green results, no need for reposting now :)
For future its Fixes #XYZ for issues and Closes #XYZ for the PR containing the commit. One commit per PR is enough to contain the Closes.

PR #715, opened 2019-07-04
updates prop being passed to EmptyState contents

jgiardino:
The message that was being passed to the EmptyState for the blueprint components on the Edit Blueprint page was accidentally changed in a previous PR. This PR reverts that change.

### DIFF
--- a/components/ListView/BlueprintContents.js
+++ b/components/ListView/BlueprintContents.js
@@ -56,7 +56,7 @@ const BlueprintContents = props => {
             title={formatMessage(messages.emptyStateErrorTitle)}
             message={formatMessage(messages.emptyStateErrorMessage)}
           >
-            <p>{errorState.message}</p>
+            <p>{errorState.msg}</p>
             {pastLength > 0 && (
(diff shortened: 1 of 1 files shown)


components/ListView/BlueprintContents.js
```diff
@@ -56,7 +56,7 @@ const BlueprintContents = props => {
             title={formatMessage(messages.emptyStateErrorTitle)}
             message={formatMessage(messages.emptyStateErrorMessage)}
           >
-            <p>{errorState.message}</p>
```
KKoukiou: Oh that's weird if you consider that the proptypes bellow are like this
```
BlueprintContents.propTypes = {
...
 message: PropTypes.string,
...
}
```

martinpitt:
So that reverts commit 56f688dde48dc64e. If that broke something, then sorry about that! I just don't understand where this come from? `msg` is not being used anywhere else in the code, just `message`. Anyway, let's land this and then re-fix the propTypes then, I'll let you review that.

FWIW, I still don't understand why prop-types are being checked for @garrett and me, but not for you or CI..

martinpitt:
I fixed the propType in PR #716

PR #717, opened 2019-07-04
Add additional blueprint name validation

### LINKED ISSUE #317: Error handling for blueprint names
When creating a new blueprint, an api call will return an error if any characters that are not allowed are used. If that occurs:

- The modal remains open
- An error message should display to the user in the Create Blueprint modal explaining what the issue is with the name.

NOTE: this issue requires updates to the api

jkozol:
Fixes #317

A user attempting to create a blueprint with spaces or other invalid
charaters in the name will be notified about the invalid character and
prevented from creating the blueprint.

### DIFF
--- a/components/Modal/CreateBlueprint.js
+++ b/components/Modal/CreateBlueprint.js
@@ -255,7 +255,7 @@ class CreateBlueprintModal extends React.Component {
             id="create-blueprint-modal-create-button"
             type="button"
             className="btn btn-primary"
-            disabled={this.state.name.length === 0 || this.nameContainsError()}
+            disabled={this.nameContainsError()}
             onClick={() => this.handleCreateBlueprint()}
           >
             <FormattedMessage defaultMessage="Create" />
--- a/components/Modal/CreateBlueprint.js
+++ b/components/Modal/CreateBlueprint.js
@@ -57,7 +57,7 @@ class CreateBlueprintModal extends React.Component {
       errorNameDuplicate: false,
       errorNameSpace: false,
       errorNameInvalid: false,
-      errorNameInvalidChar: "",
+      errorNameInvalidChar: [],
       errorInline: false,
       name: "",
       description: "",
@@ -114,7 +114,7 @@ class CreateBlueprintModal extends React.Component {
       errorNameSpace: false,
       errorInline: false,
       errorNameInvalid: false,
-      errorNameInvalidChar: ""
+      errorNameInvalidChar: []
     });
   }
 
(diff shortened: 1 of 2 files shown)


components/Modal/CreateBlueprint.js
```diff
@@ -232,7 +255,7 @@ class CreateBlueprintModal extends React.Component {
             id="create-blueprint-modal-create-button"
             type="button"
             className="btn btn-primary"
-            disabled={this.state.name == "" || this.state.errorNameDuplicate}
+            disabled={this.state.name.length === 0 || this.nameContainsError()}
```
KKoukiou: errorNameEmpty in nameContainesError should make the first check redundant

components/Modal/CreateBlueprint.js
```diff
@@ -105,26 +108,45 @@ class CreateBlueprintModal extends React.Component {
   }
 
   dismissErrors() {
-    this.setState({ errorNameEmpty: false, errorNameDuplicate: false, errorInline: false });
+    this.setState({
+      errorNameEmpty: false,
+      errorNameDuplicate: false,
+      errorNameSpace: false,
+      errorInline: false,
```
KKoukiou: This has no reason to be in the state. The function you introduced nameContainsError can return it's value to a normal variable in the render function and the result can be reused there.
But since you didn't introduce this, you don't have to change it now.

components/Modal/CreateBlueprint.js
```diff
@@ -105,26 +108,45 @@ class CreateBlueprintModal extends React.Component {
...
 
-  handleErrors(blueprintName) {
-    this.handleErrorDuplicate(blueprintName);
-    this.handleErrorName(blueprintName);
-  }
+  handleErrorNameInvalid(blueprintName) {
+    if (blueprintName.length === 0 && this.state.checkErrors) {
+      this.setState({ errorNameEmpty: true });
+    } else {
+      const invalidCharsRegex = /[^a-zA-Z0-9_,.:+*-]/;
```
KKoukiou: Can a name contain + * , : 
I find it weird but if the backend allows it I am totally fine.

jkozol: I agree that it is a little strange but the backend seems to allow those characters.

components/Modal/CreateBlueprint.js
jkozol: When the modal initially opens nameContainsError will return false because we do not check for an error before the user interacts with the modal. However, the user should still be prevented from creating a blueprint with an empty name. This duplicate check is an easy fix to disable the button and still not show errors as soon as the modal opens.

jkozol: I was wrong. I forgot about the onBlur call to handleErrorNameInvalid.

components/Modal/CreateBlueprint.js
```diff
@@ -105,26 +108,45 @@ class CreateBlueprintModal extends React.Component {
...
   }
 
-  handleErrorName(blueprintName) {
-    if (blueprintName === "" && this.state.checkErrors) {
-      setTimeout(() => {
-        this.setState({ errorNameEmpty: true });
-      }, 200);
+  nameContainsError() {
+    if (
+      this.state.errorNameEmpty ||
```
KKoukiou: It's not nice to return true/false confitionaly
Just return the result of the condition inside the if

jgiardino:
These updates work great!

One thing I noticed is that only one issue is displayed at a time. In the example below, I have two invalid characters in the name, but the message that displays only mentions one invalid character. In a scenario where the user is copying and pasting the name, and several invalid characters are present, it would be really tedious for them to fix the name one character at a time. For example, the user removes all the spaces, only then to realize they need to remove all the slashes. 
`user` = 😠 

What are your thoughts on having one message that lists all the characters. Maybe something like:
> Blueprint names cannot include spaces or the characters: / , & $

I'm not sure if it's better/simpler to list all invalid characters or just the invalid characters present in the input field. If the latter is preferred, then the message could be one of these variations:
1. > Blueprint names cannot include spaces.
2. > Blueprint names cannot include the character: /
3. > Blueprint names cannot include the characters: / $
4. > Blueprint names cannot include spaces or the character: / 
5. > Blueprint names cannot include spaces or the characters: / $


image
image

jkozol:
@jgiardino I think the second option would be better since there are many invalid characters (all of them except ._-). I can update the error message. I am also updating the test to attempt to click the the Create button before checking if it is disabled.

PR #723, opened 2019-07-09
test: Add blueprint name character verification case

henrywang:
1. blueprint name with invalid character should not be allowed
2. all invalid characters should be reported with error message

### DIFF
--- a/test/end-to-end/specs/createBlueprint.test.js
+++ b/test/end-to-end/specs/createBlueprint.test.js
@@ -44,10 +44,10 @@ describe("Create Blueprints Page", function() {
     expect(createBlueprintPage.createButton.getAttribute("disabled")).to.equal("true");
   });
 
-  it("Create button should be enabled when create blueprint dialog has a valid name", function() {
+  it("Create button should be enabled without a help box notification when create blueprint dialog has a valid name", function() {
     const validName = "-_.";
     createBlueprintPage.nameBox.setValue(validName);
-    createBlueprintPage.descriptionBox.click();
+    $(createBlueprintPage.helpBlockSelector).waitForVisible(timeout, true);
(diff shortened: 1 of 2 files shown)


test/end-to-end/specs/createBlueprint.test.js
```diff
@@ -44,6 +44,15 @@ describe("Create Blueprints Page", function() {
     expect(createBlueprintPage.createButton.getAttribute("disabled")).to.equal("true");
   });
 
+  it("Invalid character for blueprint name should not be allowed", function() {
+    const invalidName = "`~!@#$%^&*()=+[]{}\\|;:'\",<>/?";
+    createBlueprintPage.nameBox.setValue(invalidName);
```
martinpitt: IMHO this is too strong a test case. This should already fail with a name that contains a single invalid character -- it's almost guaranteed to fail on a name that's nothing but garbage :grin: 

I'd test a name which has some funny, but still valid characters (which should be accepted), and then the edge case from the other side by adding one invalid character; and maybe have a few cases of that with different invalid ones.

I realize this is a bit of bikeshedding at this point. But this is a good opportunity for @jkozol to get familiar with the tests, run it locally, and adjust it as above?

henrywang:
Thanks @martinpitt! @jkozol, would you like to adjust this PR according to @martinpitt's comment? :)

jkozol:
@martinpitt @henrywang Certainly! I will add tests to test for each of the 4 possible error messages and one to make sure we don't throw an error for the characters we allow: -_.

henrywang:
Thanks @jkozol! Just ping me if you have any questions.

henrywang:
@jkozol, Looks good to me. The Codecov works finally! And we have 0.32% coverage improvement just because of [error message](https://codecov.io/gh/weldr/cockpit-composer/pull/723/changes). Great!

PR #725, opened 2019-07-10
Use cockpit's PatternFly CSS with some css updates for composer

jgiardino:
This takes updates made in https://github.com/weldr/cockpit-composer/pull/700 and adds a few miscellaneous css updates in the context of Composer custom css updates.

There were a couple of items provided in the cockpit css that didn't have sufficient color contrast. I updated those here, but it might be worth checking if these are used in cockpit, and if the color contrast issues are present there as well.
```
.fields-status-pf { 	
  color: #8b8d8f;
  margin-bottom: 15px;
}
```
 I updated this in composer to `var(--pf-global--Color--200);`

```
.label-info { 
  background-color: #73bcf7;
}
```
 I updated this in composer to `var(--pf-global--link--Color)`

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "fastclick": "1.0.6",
     "history": "3.3.0",
     "jquery": "3.4.1",
-    "patternfly": "3.59.3",
     "patternfly-react": "1.19.1",
     "prop-types": "15.7.2",
     "react": "16.8.6",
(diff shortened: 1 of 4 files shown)


martinpitt:
Many thanks, @jgiardino! As this and the original PR #700 are identical now, I copied your comment to #700 (as that has more history), and close this one.

PR #731, opened 2019-07-16

cockpituous:
 * [x] po-refresh

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -9,7 +9,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "MIME-Version: 1.0\n"
 "X-Generator: Zanata 4.6.2\n"
-"PO-Revision-Date: 2019-06-11 10:48+0000\n"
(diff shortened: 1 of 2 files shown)


po/fr.po
```diff
@@ -230,12 +230,55 @@ msgstr "Modèle"
...
+"Le noms de modèles (topologie) ne peuvent pas contenir les caractères : "
+"{invalidChar}"
+
 #: build/messages/pages/blueprints/index.json
 #. [pages.blueprints.blueprintsTitle]
 #. defaultMessage is:
 #. Blueprints
 msgid "Blueprints"
-msgstr "Modèles"
+msgstr "Modèles "
```
marusak: Interesting change. Could this have some reason or just typo that should be changed back in Zanata?

larskarlitski: Looks like a typo to me. The original string doesn't have it and there shouldn't be a space where this appears in the UI.

Nice catch!

marusak:
@larskarlitski updated the string, lets close this and let update zanata once more.

PR #736, opened 2019-07-22
test: Work around off screen element issue on Edge

henrywang:
1. run scrollIntoView() from javascript to make element visible
2. click element directly by javascript if scrollIntoView() does not work in some cases


martinpitt:
This keeps [failing on edge](https://209.132.184.41:8493/logs/pull-736-20190722-174545-58a7b4af-weldr-cockpit-composer-cockpit-fedora-30-edge/log) unfortunately, I retried several times.

henrywang:
@martinpitt, failed case got fixed now. Sorry for that.
The test result on edge totally depends on screen resolution on running machine. I run the test on my home machine with really high resolution without any issue. But the tests failed on CI env and my laptop. The root cause is all about off-screen element issue.

PR #747, opened 2019-08-01
Remove checkboxes from ListItemComponents component since they were n…

### LINKED ISSUE #744: List items on Edit page should not have checkboxes
Currently on Edit Blueprint page, the list items that display in the Blueprint Components tab have checkboxes, but there are no actions associated with selected items.

At a minimum, these checkboxes should be removed. However, we could keep them on the list items under the Selected Components tab and include a Remove action in the toolbar. We would still want to remove the checkboxes from Dependency list items (you can't remove a dependency).

KKoukiou:
…ot used

Fixes #744

### DIFF
--- a/components/ListView/ListItemComponents.js
+++ b/components/ListView/ListItemComponents.js
@@ -49,9 +49,6 @@ class ListItemComponents extends React.Component {
           <div className="list-pf-chevron">
             <span className={`fa ${expanded ? "fa-angle-down" : "fa-angle-right"}`} />
           </div>
-          <div className="list-pf-select">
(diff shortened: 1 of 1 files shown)


martinpitt:
Some trouble with the [Edge test](https://209.132.184.41:8493/logs/pull-747-20190806-092626-5e202675-weldr-cockpit-composer-fedora-30-edge/log), retrying a few times.

henrywang:
@martinpitt, Thanks! 
I tried on my local environment. No this kind of issue. Still working on this issue.

martinpitt:
Green now!

PR #751, opened 2019-08-02
Fix wrong PropType for homepage

KKoukiou:
Homepage is always returned as a single string from the API.


martinpitt:
We can still land this already, but guarding against that in the future would be nice.

Issue #743, opened 2019-07-29
Text and terminology refinements


jgiardino:
This issue is based on findings from a recent usability study where participants where tasked with modifying the hostname and creating a user account. Before these features were added, the only parts of a blueprint that could be updated in the UI were the description and package list. The terminology defined at the time was not found to be problematic. But now that we have customizations that can also be defined for a blueprint within the UI, the terminology is problematic. "Edit Blueprint" was not obvious that you were only modifying the package list. Other edits pertaining to customizations were hard to find. 

Note that the suggestions below are just a first pass at addressing these issues, focusing mainly only on text changes. Further research activities are planned to gain better insight into what improvements can be made to the workflow overall.

**Updates needed for this issue:**

1. Blueprints page (listing all blueprints)
    a.  For each list item, change the text "Edit Blueprint" to "Edit Packages"
      - Note that the panel that lists packages in the Edit page refers to these things as "components" because we want to be able to list things other than packages in these panels. But the fundamental task associated with this page results in defining the list of packages for the blueprint.
2. Blueprint page
    a. Move the ability to edit the blueprint description to the page header.
    b. Change the text "Details" to "Customizations" in the primary tabs
    c. Change the text "Selected Components" to "Packages" in the primary tabs **only**
    d. Change the text "Edit Blueprint" to "Edit Packages"
3. Edit Blueprint page
    a. Change the text "Edit Blueprint" to "Edit Packages" in the breadcrumb

PR #752, opened 2019-08-02
Cleanup Tabs in Blueprints page

KKoukiou:
…tabs

There is not any benefit to keep the active tabs in the store.
These information should be kept in the state of the component that
creates the Tabs and not somewhere in the global state, but we are using
Tabs as uncontrolled component here, so we don't need to maintain the
state of the active tabs at all.

### DIFF
--- a/pages/blueprint/index.js
+++ b/pages/blueprint/index.js
@@ -310,6 +310,8 @@ class BlueprintPage extends React.Component {
     if (blueprint.customizations !== undefined && blueprint.customizations.user !== undefined) {
       users = blueprint.customizations.user;
     }
+    const pathSuffix = cockpit.location.path[cockpit.location.path.length - 1];
+    const activeKey = ["customizations", "packages", "images"].includes(pathSuffix) ? pathSuffix : undefined;
 
     return (
       <Layout className="container-fluid" ref={c => (this.layout = c)}>
@@ -374,7 +376,11 @@ class BlueprintPage extends React.Component {
             </p>
           </div>
         </header>
-        <Tabs id="blueprint-tabs">
+        <Tabs
+          activeKey={activeKey}
+          onSelect={eventId => cockpit.location.go(["blueprint", blueprint.name, eventId])}
+          id="blueprint-tabs"
+        >
           <Tab eventKey="customizations" title={formatMessage(messages.customizationsTitle)}>
             <div className="tab-container row">
               <div className="col-sm-12">
--- a/pages/blueprint/index.js
+++ b/pages/blueprint/index.js
@@ -47,8 +47,7 @@ import {
 import {
   setEditDescriptionVisible,
   setEditHostnameVisible,
-  setEditHostnameInvalid,
-  setActiveTab
+  setEditHostnameInvalid
 } from "../../core/actions/blueprintPage";
 import {
   componentsSortSetKey,
@@ -128,7 +127,6 @@ const messages = defineMessages({
 class BlueprintPage extends React.Component {
   constructor() {
     super();
-    this.handleTabChanged = this.handleTabChanged.bind(this);
     this.handleComponentDetails = this.handleComponentDetails.bind(this);
     this.handleComponentListItem = this.handleComponentListItem.bind(this);
     this.handleDepListItem = this.handleDepListItem.bind(this);
@@ -162,12 +160,6 @@ class BlueprintPage extends React.Component {
 
   componentWillUnmount() {
     this.props.clearSelectedInput();
-  }
-
-  handleTabChanged(key) {
-    if (this.props.blueprintPage.activeTab !== key) {
-      this.props.setActiveTab(key);
-    }
   }
 
   handleComponentDetails(event, component) {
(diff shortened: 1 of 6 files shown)


jgiardino:
Thanks for working on this! There are a few things I notice when testing the updates:

1. When I first navigate to the blueprint page, I don't see a url for the Details tab in the address bar, but when I click the Details tab, I do see a url include `/details`. I don't think this is an issue, but just wanted to mention it.
2. After I navigate to the blueprint page and click each tab, when I hit the Back button for the browser, it takes me to the previous tab I was on. If I hit the Back button again, I go forward to the last tab I was on. Hitting Back only takes me back and forth between the last two tabs.
3. If I copy and paste a url for a tab in another browser window, the contents for that page aren't loaded.
4. In the original issue, I had included this comment: 
    > This should be completed after #743, so that the urls match the updated tab labels.
    My thinking on this order is the urls we add will not change when we implement the text string changes described in #743. I'm not sure if this is a real concern or not.

jkozol:
@jgiardino Thank you for the review. I updated the urls to reflect the text string updates.

> When I first navigate to the blueprint page, I don't see a url for the Details tab in the address bar, but when I click the Details tab, I do see a url include /details. I don't think this is an issue, but just wanted to mention it.

Thanks for pointing this out. I don't think this is an issue because it doesn't affect the ability to click on the details tab and we will be changing our default view soon.

> After I navigate to the blueprint page and click each tab, when I hit the Back button for the browser, it takes me to the previous tab I was on. If I hit the Back button again, I go forward to the last tab I was on. Hitting Back only takes me back and forth between the last two tabs.

What would your desired functionality be here? Do you think that if a user starts on the edit blueprint page then goes to the blueprint details page then changes tabs a few times they should go back to the edit blueprint page on a single Back button click?

> If I copy and paste a url for a tab in another browser window, the contents for that page aren't loaded.

I am not able to replicate this issue.

jgiardino:
Thanks for working to finish this one up. I was able to test this out locally. I'm not seeing the issue anymore regarding ability to copy/paste a url into another browser window. And I'm fine with not seeing "/customizations" in the url when you first navigate to a blueprint page for the reasons you mention.

I am noticing some funky behavior when using some of the features built into the browser, related to the Back button, but also when I try to right click a tab and open the link associated with that tab into a new window (this issue might be due to not having an `href` value defined for the link). I captured some steps to reproduce for the first issue, that you asked about above. But also, I'm totally fine with capturing these in a separate issue to keep the scope on this PR from creeping out.

> What would your desired functionality be here? Do you think that if a user starts on the edit blueprint page then goes to the blueprint details page then changes tabs a few times they should go back to the edit blueprint page on a single Back button click?

Here is a comparison of similar actions between the PF website and Image Builder.

Steps in PF site:

1.  I open a new browser window and enter the following in the address bar: https://www.patternfly.org/v4/
2. I clicked Get started
3. I clicked Design guidelines
4. I clicked Documentation
5. I hit the Back button in the browser
6. I right-clicked the Back button, and that's the example you see below, where you can see the history of urls that existed in the address bar for this browser window, and where I'm currently at in that history.

image

Steps in Image Builder:

1.  I open a new browser window and enter the following in the address bar: https://192.168.56.102:9090/welder
2. I click on a blueprint name
3. I click the Packages tab
4. I click the Images tab
5. I click the Edit Packages button
6. I click the Back button in the browser
7. I right-clicked the Back button, and that's the example you see below.

In this case, what you can't see is that the Forward button is disabled and the Edit Packages page isn't represented in this history. But also, there are two instances for each URL listed, and hitting the Back button doesn't traverse this history as expected. I would suggest trying this yourself and hovering over each item in the browser history for the Back button to see the actual urls being listed. As I mentioned above, I'm totally fine capturing this as a separate issue, but wanted to mention it here in case it's obvious what the fix is and it's easy to include.  

image

pages/blueprint/index.js
```diff
@@ -311,7 +311,7 @@ class BlueprintPage extends React.Component {
       users = blueprint.customizations.user;
     }
     const pathSuffix = cockpit.location.path[cockpit.location.path.length - 1];
-    const activeKey = ["details", "selected-components", "images"].includes(pathSuffix) ? pathSuffix : undefined;
+    const activeKey = ["customizations", "packages", "images"].includes(pathSuffix) ? pathSuffix : undefined;
```
KKoukiou: This change should be done when introducing this line in the previous commit and not here https://github.com/weldr/cockpit-composer/pull/752/commits/e9359dbb62a82d8571f55195569b9b99edb0d84a

pages/blueprint/index.js
```diff
@@ -381,7 +381,7 @@ class BlueprintPage extends React.Component {
           onSelect={eventId => cockpit.location.go(["blueprint", blueprint.name, eventId])}
           id="blueprint-tabs"
         >
-          <Tab eventKey="details" title="Details">
+          <Tab eventKey="customizations" title={formatMessage(messages.customizationsTitle)}>
```
KKoukiou: Same for this.

PR #756, opened 2019-08-08
Text string updates

jgiardino:
This closes updates described in #743. Additionally, after discussion with @andreasn, the inline edit pattern wasn't the best fit to use for the description that displayed in the blueprint page header. Instead, the description displays as a link in the header and opens a modal for modifying the description. An action is also included in the kebab which will open a modal. This allows the user to provide a description if one isn't defined.

image

image

### DIFF
--- a/components/ListView/BlueprintListView.js
+++ b/components/ListView/BlueprintListView.js
@@ -19,7 +19,7 @@ class BlueprintListView extends React.PureComponent {
           <div className="list-group-item" key={blueprint.name} data-blueprint={blueprint.name}>
             <div className="list-view-pf-actions">
               <Link to={`/edit/${blueprint.name}`} className="btn btn-default">
-                <FormattedMessage defaultMessage="Edit Blueprint" />
+                <FormattedMessage defaultMessage="Edit Packages" />
               </Link>
               <CreateImage blueprint={blueprint} layout={layout} />
               <div className="dropdown pull-right dropdown-kebab-pf">
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -288,7 +288,7 @@ class ComponentDetailsView extends React.Component {
         )}
         <div>
           <Tabs id="blueprint-tabs">
-            <Tab eventKey="customizations" title="Customizations">
+            <Tab eventKey="details" title="Details">
               <h4 className="cmpsr-title">{component.summary}</h4>
               <p>{component.description}</p>
               <dl className="dl-horizontal">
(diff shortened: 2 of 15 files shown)


andreasn:
Looks good to me!

jgiardino:
@henrywang I've made some updates to the Edit Description functionality and moved it to another part of the page, so that the first tab can include only the Customizations that are defined in the blueprint.

I reviewed the tests to try to understand what updates are needed, but was having trouble understanding some of the updates needed. Any help is greatly appreciated. 😄

henrywang:
@jgiardino, never mind, I can take care of that. Before that I have some questions.

The new changes confused me. There're more "terms" on the UI, like `blueprint`, `package`, `component`, `customization`. What's the different between them?

Create a blueprint, but edit `package`? blueprint === package?
Screenshot_2019-08-12 Image Builder - localhost localdomain(3)

Edit package, but choose `component`? package === component?
Screenshot_2019-08-12 Image Builder - localhost localdomain(1)

Component details tab, but there's not details, but customization? details === customization?
Screenshot_2019-08-12 Image Builder - localhost localdomain(2)

That does not make sense to me. And if the UI changed, how about composer-cli? They have to be identical, otherwise, the user will be confused on them.

jgiardino:
Thanks for reviewing @henrywang! These updates are related to findings from the latest [usability test](https://docs.google.com/presentation/d/1OAteyrNGTDAq3AxFSYmfcg9aEUj2rgciFgrLdzxsIfo/edit?usp=sharing). I would like to do more research activities this week to better understand how we can make the workflow easier to follow, and then I would like to make more updates to the UI as a result. This PR is just meant to be simple updates for now, primarily related to text strings, which caused confusion with users during the test. 

> And if the UI changed, how about composer-cli? They have to be identical, otherwise, the user will be confused on them.

I am trying to bring the UI more inline with the cli, which uses the labels "customizations" and "packages". This is why the tabs on the Blueprint page are changing, and also why the ability to edit the description is moving to the Blueprint page header (because it isn't a "customization").

> Edit package, and choose component. What's the relationship between package and component? Does package include component? Before, blueprint includes component, that makes sense.

@KKoukiou asked a similar thing :-) I'm hesitant to remove or change "Component" at this time. This term was originally chosen so that we could list both packages and modules in the "Available Components" and "Selected Components" panels, which we still need to add. But usability participants were really confused by "Edit Blueprint" taking them to a place where they select packages. My thinking here is the "Edit Packages" page is where you specify what packages are included in your image by selecting packages and eventually modules. But we would definitely do more user testing in the future to verify that these updates make sense to users. 

> Component details tab, but there's not details, but customization? details === customization?

This was not intended. Thanks for catching this! I'll find that and change it back to "Details"

jgiardino:
@henrywang Thanks so much for updating the tests!!

@KKoukiou @martinpitt Are either of you around to do one more review of this?

PR #760, opened 2019-08-13
Convert to standard plugin

marusak:
Someone mentioned that this would make sense, so I am opening this PR to start some discussion.
The main reasoning is that the backend runs on a machine, so if you switch to other machine this plugin can disappear/appear/change.
What do you @martinpitt @larskarlitski think?

downthere

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -4,11 +4,9 @@
         "cockpit": "134"
     },
 
-    "dashboard": {
+    "tools": {
         "index": {
-            "label": "Image Builder",
-            "order": 30,
-            "icon": "pficon-build"
+            "label": "Image Builder"
         }
     },
     "content-security-policy": "default-src 'self' 'unsafe-eval'"


martinpitt:
In the screenshot it's currently a bit cluttered as you have the cockpit-tests playgrounds installed. At the sprint we discussed moving the "external" applications to the top. So "tools" is too far down IMHO, it should be a menu with priority 30, see https://cockpit-project.org/guide/latest/packages.html#package-manifest

marusak:
Menu with 30 seems too high for me, see:
menu30
I think it fits better on top of containers and vms: (order:41-49, if 50 then it ends up between Docker and Podman, if 40 then ends up between Stare and Networkign)
menu49

jgiardino:
I agree with what @martinpitt said about the navigation looking cluttered. Image Builder seems to get lost in that navigation. When we discussed an updated [navigation pattern based on PF](https://www.patternfly.org/v4/documentation/react/components/nav/#nav-mixed), my impression was that there would be a clear delineation between system tools and the additional applications that are installed, with the applications being at the top. 

My preference would be to make this change after we have adopted the updated navigation pattern that we discussed last week. 

But also, it looks like this implementation prevents me from adding secondary navigation. This is not something we have now, but would like to add in the future, and without knowing the timing of the PF navigation update, I wouldn't want to be blocked by that when we're ready to add secondary nav to image builder.

martinpitt:
I'm ok with 40 as well. :-)

But indeed that changes the game for composer quite a bit: before it had the full screen real estate, now there's always the cockpit menu. At one point [composer even did have its own menu](https://github.com/weldr/cockpit-composer/pull/150), this was the main reason why it started out as a dashboard.

So if there's any chance that a composer menu will come back, it needs to stay a dashboard for now, and we need to think about how to accomodate second-level menus.

> My preference would be to make this change after we have adopted the updated navigation pattern

Sounds reasonable to me!

larskarlitski:
Nice, thanks for starting the discussion here!

I think it definitely needs to move to the context of a single machine, but I'm with Jenn on this one. Especially that it blocks us from adding secondary navigation in Image Builder, for which we already have designs.

@garrett what do you think?

marusak:
I remembered https://github.com/cockpit-project/cockpit/issues/11310 which is kinda related. I think we should go with that one and come up with a new design. @garrett has some good points in the last comment of that issue.

It would be awesome if @andreasn @garrett and @jgiardino could design something.

garrett:
> My preference would be to make this change after we have adopted the updated navigation pattern that we discussed last week.

Yep, I was about to say the same as @jgiardino and fully agree :100:%.

Thanks for starting the conversation here though, so it's not just what we said last week. :+1:

I think this is probably something to work on together once we're all back from our Summer holidays.

marusak:
So there is design how menu should look like and it does not actually require manifest to change. Closing.

PR #767, opened 2019-08-23
 Add upload image functionality and a create image upload wizard

jkozol:
The create image modal has been replaced with a wizard that can create an image while also allowing the user to upload the image to a select group of cloud providers. This wizard guides the user through selecting the image type and where to upload the image, then has the user enter the settings needed to upload the image, and finally has the user review the upload settings before uploading the image. The redux store and sagas have also been updated to handle the upload data. 

This PR is dependent on https://github.com/weldr/lorax/pull/843

This PR includes only a basic uploads list attached to each image in the images list and no ability to create an upload for an already created image. There are also a few of console warnings which still need to be fixed:

- [x] Each child in a list should have a unique "key" prop error caused by the mapping of TextListItems

- [x] Changing an uncontrolled input of type text to be controlled caused by the use of TextInputs

- [x] Inline style errors caused by Patternfly 4 (fixed in the recent version of patternfly which requires additional configuration in our webpack config)


jgiardino:
🎉 🎉 🎉  This is awesome!

There are some console errors that prevented me from accessing some of the updates, but here are some comments and questions on what I could review...

- Should we provide a default value for image names? But also, is the user naming the file? Or what are they naming? And this question might be out of scope of this PR, but when we add support for listing the compose/manifest/osbuild from which an image file is created, would we name that for the user or let them provide a name? Maybe we can chat about all of this, and maybe at standup to include those working on osbuild.
- We can add styles to fix the issue with the  alignment of checkboxes in the form, and also the static text at the top:
  ```
  .pf-c-form__horizontal-group .pf-c-check,
  .cc-c-form__static-text {   padding-top: var(--pf-global--spacer--form-element);
    padding-bottom: var(--pf-global--spacer--form-element);
  }

  .pf-c-form__horizontal-group .pf-c-check__label {
    line-height: var(--pf-global--LineHeight--md);
  }
  ```
  - The class `cc-c-form__static-text` would be added to the `<p>` element that includes the blueprint name. I was thinking we could start namespacing all of the classes we add to PF4 components with `cc-` for "cockpit-composer".
- We should also use slightly different html to make the checkbox form group more accessible:
  ```
  <div class="pf-c-form__group">
    <span class="pf-c-form__label" id="horizontal-form-checkbox">
      <span class="pf-c-form__label-text">Upload Image</span>
    </span>
    <fieldset class="pf-c-form__horizontal-group" aria-labelledby="horizontal-form-checkbox">
      <div class="pf-c-check">
        <input id="azure-checkbox" class="pf-c-check__input" type="checkbox" aria-invalid="false" value="azure">
        <label class="pf-c-check__label" for="azure-checkbox">Upload to Azure</label>
      </div>
    </fieldset>
  </div>
  ```
  - The `<label>` that displays on the left become a `<span>` with an `id`.
  - The `div.pf-c-form__horizontal-group` becomes a `<fieldset>` element, and instead of a `<legend>` to label this fieldset, there's an `aria-labelledby` attribute that references the `id` of the label on the left.

jkozol:
> Should we provide a default value for image names? But also, is the user naming the file? Or what are they naming?

Okay so asking for Image Name may actually be misleading. The user is setting the image name for the upload. For example, if the user uploads an image to a service with the name "database-image" they will see an image with that name in their cloud provider. But, the image that we show in the UI will no have that user defined name. Only the uploads for a given image will have user defined names. Do you think we should change the terminology?

> And this question might be out of scope of this PR, but when we add support for listing the compose/manifest/osbuild from which an image file is created, would we name that for the user or let them provide a name? Maybe we can chat about all of this, and maybe at standup to include those working on

I agree that this is a good question to be be discussed in more detail at either the standup or engineering meeting. In my opinion I think we should allow users to name their compose, then have the image be named from the compose and image type such as "compose-name-tar", and finally allow the user to specify the name for their upload. 

I will update the styles per your recommendations.

package.json
```diff
@@ -8,6 +8,8 @@
   },
   "dependencies": {
     "@babel/polyfill": "7.4.4",
+    "@patternfly/react-core": "^3.75.2",
+    "@patternfly/react-styles": "^3.5.7",
```
KKoukiou: react-styles is brought as a dependency of react-core anyway. You should not explicitly set it unless some specific version is broken and you need to specify it.
See `npm info @patternfly/react-core` to verify that it's indeed dependency of that package.

package.json
```diff
@@ -52,11 +54,12 @@
     "eslint-plugin-react": "^7.11.1",
     "eslint-plugin-redux-saga": "^0.10.0",
     "faker": "^4.1.0",
-    "file-loader": "^0.9.0",
+    "file-loader": "^4.1.0",
     "glob": "^6.0.4",
     "html-webpack-plugin": "^3.2.0",
     "jed": "^1.1.1",
     "jsel": "^1.1.6",
+    "mini-css-extract-plugin": "^0.8.0",
```
KKoukiou: What do we need this for? I did start using PF4 in cockpit but i didn't need this.

webpack.config.js
```diff
@@ -95,7 +100,27 @@ module.exports = {
       },
       {
         test: /\.css$/,
-        use: ["style-loader", "css-loader"]
+        include: [
```
KKoukiou: It's better that each component from PF/react-core will import what css files it needs instead of importing everything by ourselves.
Or what are you actually trying to do here?

package.json
```diff
@@ -8,9 +8,9 @@
   },
   "dependencies": {
     "@babel/polyfill": "7.4.4",
```
KKoukiou: This commit should be squashed with the first in the series which introduces these new modules.

components/Modal/CreateImageUpload.js
```diff
@@ -123,7 +123,11 @@ class CreateImageUploadModal extends React.Component {
   }
 
   setUploadService(_, event) {
-    this.setState({ uploadService: event.target.value });
+    if (this.state.uploadService === event.target.value) {
```
KKoukiou: Please no fix-up commits. This should be merged with the previous commit where this functionality was introduced. In general, each commit in the PR should not introduce unfinished/problematic code waiting for followup commits to fix the issues.

components/ListView/ListItemUploads.js
```diff
@@ -0,0 +1,98 @@
...
+                      <div className="list-view-pf-additional-info-item">
+                        <span className="pficon pficon-error-circle-o" aria-hidden="true" />
+                        <FormattedMessage defaultMessage="Failed" />
+                      </div>
+                    )}
+                  </DataListCell>
+                ]}
+              />
+              <DataListAction id="upload-logs">
+                <Button variant="disabled">Logs</Button>
```
KKoukiou: This button is always disabled!?

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+  }
+
+  setImageType(imageType) {
+    this.setState({
+      imageType,
+      uploadService: ""
+    });
+  }
+
+  setImageName(imageName) {
```
KKoukiou: I am not fond of one liner functions. Just call setState directly instead of calling the function that calls setState.

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+
+  handleStartCompose(blueprintName, composeType, imageName, uploadSettings) {
+    const upload = {
+      image_name: imageName,
+      settings: uploadSettings
+    };
+    this.props.startCompose(blueprintName, composeType, upload);
+  }
+
+  handleNextStep(activeStep, callback) {
```
KKoukiou: Can you rename this from callback to something more intuitive so that someone can understand what this callback is actually doing?

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+    if (activeStep.name === "Image Type") {
+      if (this.state.uploadService.length === 0) {
+        if (this.isPendingChange()) {
```
KKoukiou: nitpick: (But good to be consistent with these) Curly brackets for onliners are not encouraged.

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+              name="horizontal-form-title"
+              onChange={this.setImageType}
+            >
+              <FormSelectOption isDisabled key="default" value="" label={formatMessage(messages.selectOne)} />
+              {imageTypes.map(type => (
+                <FormSelectOption isDisabled={!type.enabled} key={type.name} value={type.name} label={type.label} />
+              ))}
+            </FormSelect>
+          </FormGroup>
+          {imageType === "vhd" && (
```
KKoukiou: The providers code bellow has a lot of duplication. I would create a function returning the React Code for the checkboxes and call it with the different provider values.

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+      ...(showUploadVSphereStep ? [uploadVSphereStep] : []),
+      ...(showReviewStep ? [reviewStep] : [])
+    ];
+
+    const createImageUploadFooter = (
+      <WizardFooter>
+        <WizardContextConsumer>
+          {({ activeStep, onNext, onBack, onClose }) => {
+            return (
+              <>
```
KKoukiou: Please stick to either this syntax or <React.Fragment>, don't use them interchangeably

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+  warningUnsaved: {
+    defaultMessage:
+      "This blueprint has changes that are not committed. " +
+      "These changes will be committed before the image is created."
+  },
+  selectOne: {
+    defaultMessage: "Select one"
+  },
+  title: {
+    defaultMessage: "Create and Upload Image"
```
KKoukiou: The modal title is 'Create and Upload Image', however for many image types upload operation is not supported.

core/actions/uploads.js
```diff
@@ -0,0 +1,20 @@
...
+export const FETCHING_UPLOAD_PROVIDERS_SUCCEEDED = "FETCHING_UPLOAD_PROVIDERS_SUCCEEDED";
+export const fetchingUploadProvidersSucceeded = providerSettings => ({
+  type: FETCHING_UPLOAD_PROVIDERS_SUCCEEDED,
+  payload: {
+    providerSettings
+  }
+});
+
+export const UPLOAD_FAILURE = "UPLOAD_FAILURE";
+export const uploadFailure = error => ({
```
KKoukiou: Do you use it anywhere?
I tried to cause an error with upload, putting invalid URL/credentials for the provider and I never saw the error message on the UI.
In addition I don't see the 'error' state var being used anywhere.

components/Modal/CreateImageUpload.js
```diff
@@ -0,0 +1,751 @@
...
+              {Object.keys(providerSettings.openstack.auth).map(key => (
+                <FormGroup
+                  label={formatMessage({
+                    id: "openstack-auth",
+                    defaultMessage: providerSettings.openstack.auth[key]
+                  })}
+                  fieldId={key}
+                  key={key}
+                >
+                  <TextInput
```
KKoukiou: This gives the following warning
A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component

I don't see what's wrong there in the first place, but maybe because you are assigning the state to another variable and using that, react gets confused thinking that the component in uncontrolled.

package.json
jkozol: Fixed

components/Modal/CreateImageUpload.js
jkozol: Noted and fixed

components/Modal/CreateImageUpload.js
jkozol: Fixed

components/Modal/CreateImageUpload.js
jkozol: Fixed

components/Modal/CreateImageUpload.js
jkozol: Do you think we should start the title as Create Image and then if the user selects an upload provider change it to Create and Upload Image?

core/actions/uploads.js
jkozol: We do not so I removed it for now.

components/Modal/CreateImageUpload.js
jkozol: This was due to the value being initialized to undefined and then later updated to a string. It is now fixed.

components/ListView/ListItemUploads.js
jkozol: Removed

components/Modal/CreateImageUpload.js
jkozol: Fixed

components/Modal/CreateImageUpload.js
jkozol: Fixed

components/Modal/CreateImageUpload.js
jkozol: Fixed

jgiardino:
I missed this in previous reviews, but [PF4 recommends sentence-style capitalization](https://www.patternfly.org/v4/design-guidelines/content/grammar-and-terminology#capitalization) for all text strings (except proper nouns and product names).  We should start adopting the PF4 style guide for all new UI or UI updates.

jkozol:
@jgiardino I believe all of the capitalization should be fixed now by the Fix capitalization in CreateImageUpload Wizard commit. I will squish this commit after your review.

webpack.config.js
```diff
@@ -95,7 +100,26 @@ module.exports = {
...
+          path.resolve(__dirname, "node_modules/@patternfly/patternfly"),
+          path.resolve(__dirname, "node_modules/@patternfly/react-styles/css"),
+          path.resolve(__dirname, "node_modules/@patternfly/react-core/dist/styles/base.css"),
+          path.resolve(__dirname, "node_modules/@patternfly/react-core/dist/esm/@patternfly/patternfly"),
+          path.resolve(__dirname, "node_modules/@patternfly/react-core/node_modules/@patternfly/react-styles/css")
+        ],
+        use: [MiniCssExtractPlugin.loader, "css-loader"]
+      },
+      {
+        test: /\.(svg|ttf|eot|woff|woff2)$/,
```
jgiardino: Just as an fyi, here's an example of how these file types are handled in the patternfly-react-seed repo:
https://github.com/patternfly/patternfly-react-seed/blob/master/webpack.common.js#L32

jkozol:
This depends on #776 now which is included in this PR as well.

jgiardino:
There is an issue on the Edit page with the current css defined for the toolbars on that page and the new PF4 wizard. I noticed that if I changed `z-index: 1000;` to `z-index: 1;` for [this ruleset](https://github.com/weldr/cockpit-composer/blob/master/public/custom.css#L64), then the issue is fixed.

image

jkozol:
Currently, this PR only contains support for uploading to AWS since it is the only service I have successfully tested. Once I am able to test vsphere and openstack I will either add those to this PR or open another one.

jgiardino:
I tested the AWS flow. I don't think I was able to upload anything, unless it's taking a really long time. I'm assuming it failed, but I'm not sure what the point of failure was. Ultimately, this is something we should address (i.e. either eliminate or expose points of failure in the UI), but I'm not sure we can in the scope of this PR. 

Aside from that, here are some suggestions:

- I would use “Create image” instead of “Create and upload image” as the wizard title since image creation is the primary task. Image uploads are secondary and optional, and not available with all types
- Upload to AWS field labels - 
  - These should be translated, except for “bucket” which appears to be the only string not translated in AWS.
  - Here are some suggestions on alternate labels
    - AWS access key >>> AWS access key ID
    - AWS secret key >>> AWS secret access key
    - AWS bucket >>> Amazon S3 bucket name
      - In the AWS console, I see strings like “Amazon S3”, “S3 buckets” and “bucket name”. 
    - I think Image name is fine, until we get further feedback.
  - On the review screen, the order of fields is different. Was that intentional? I'd suggest keeping them in the same order they're presented in the previous screens.
  - I’m not able to enter any text in the Image name field. 
  - Are all fields required? We should include a statement about required fields. Based on the PF design documentation for [indicating required fields](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#indicating-required-fields), if they're all required, we can include a statement above the form saying so. In this case, we would still want `isRequired` on the `<TextInput>` but would remove it from `<FormGroup>` to just remove the red asterisk. Otherwise, if some are required and some are optional, we can use similar patterns that we have in other forms (like Create Blueprint)
    - How do you want to handle missing required fields? I’m thinking on the review screen we can block them from clicking Finish if any required fields are missing (and in the future we can extend this to also block if authentication is failing).
  - Is the AWS region required? I’m mostly confused about this field. 
    - Do you happen to know what it’s purpose is? When I review the documentation on weldr.io, there’s no mention of specifying a region when you create a bucket or uploading to a bucket. And if the bucket is totally unique, I’m curious why we need to ask for it here. 
    - If region is important to capture here, then do I enter the string as it’s displayed in the AWS console? Or is it possible to supply this as a set of options? (maybe as a future enhancement)

jkozol:
@jgiardino Thank you for the detailed review. 

> I would use “Create image” instead of “Create and upload image” as the wizard title since image creation is the primary task. Image uploads are secondary and optional, and not available with all types

That makes sense. I've updated the string.

> These should be translated, except for “bucket” which appears to be the only string not translated in AWS

Still working on fixing this.

>AWS access key >>> AWS access key ID
AWS secret key >>> AWS secret access key
AWS bucket >>> Amazon S3 bucket name 

All switched

> On the review screen, the order of fields is different. Was that intentional? I'd suggest keeping them in the same order they're presented in the previous screens.

The image name field has been moved into the correct position.

> Are all fields required? We should include a statement about required fields. Based on the PF design documentation for indicating required fields, if they're all required, we can include a statement above the form saying so. In this case, we would still want isRequired on the <TextInput> but would remove it from <FormGroup> to just remove the red asterisk. Otherwise, if some are required and some are optional, we can use similar patterns that we have in other forms (like Create Blueprint) 

I have updated this per the suggestions in the documentation

> How do you want to handle missing required fields? I’m thinking on the review screen we can block them from clicking Finish if any required fields are missing (and in the future we can extend this to also block if authentication is failing).

requiredInfoMissing

> Do you happen to know what it’s purpose is? When I review the documentation on weldr.io, there’s no mention of specifying a region when you create a bucket or uploading to a bucket. And if the bucket is totally unique, I’m curious why we need to ask for it here.

The region is used to [create the bucket](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/create-bucket.html).

> If region is important to capture here, then do I enter the string as it’s displayed in the AWS console? Or is it possible to supply this as a set of options? (maybe as a future enhancement)

Yes. You should enter the region as [listed here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html). This is good idea. Having a set of regions for users to select from would be useful.

jgiardino:
Your updates are great! Here are some comments...

1. Just capturing for the record... We chatted offline about removing Region, which is part of the flow for creating a bucket. I included a comment about that in #822. Once that field is removed, then the statement in the UI about all fields being required will be accurate.

2. We also chatted about obscuring the password-like inputs, which I don't see in the latest commits, but I assume you just haven't pushed yet.

3. And we chatted about AWS help text, but I think that should be added in a separate PR. 

4. Something I noticed in playing around with missing field info is that this message doesn't display on the review step if the bucket input is empty, but instead it should display if any required fields are missing (please ignore that it's in Spanish :-/ ).
    image

5. I had also mentioned in a previous comment that I can't enter anything in the Image name input. I tested this in different browsers, and this is still the case with FF on my macbook. I'm curious if you're able to reproduce this issue. If not, let me know and I can try to figure out what's causing that. I'm also fine moving this to a separate issue.

To summarize, I see 1, 2 & 4 as being items to address in the scope of this PR, and 3 & 5 can be separate issues. Let me know if you would suggest something different. 

BTW, I did kick off the upload process. This is exciting to see, even if it did fail. 

image

jgiardino:
Just took a look at the latest updates. They look great!! No suggestions from me. :-)

core/composer.js
```diff
@@ -168,49 +169,54 @@ export function diffBlueprintToWorkspace(blueprintId) {
 }
 
 export function getSourceInfo(sourceName) {
-  return get("/api/v0/projects/source/info/" + encodeURIComponent(sourceName));
+  return get("/api/v1/projects/source/info/" + encodeURIComponent(sourceName));
 }
 
 export function newSource(source) {
-  return post("/api/v0/projects/source/new", source);
+  return post("/api/v1/projects/source/new", source);
```
ondrejbudai: I think this won't work. v1 source API requires the id key and I didn't find any traces we're setting it in this PR. But maybe I overlooked something.

See: https://github.com/weldr/lorax/blob/master/src/pylorax/api/v1.py#L164

jkozol: Thanks for pointing this out. I've set it back to v0.

jkozol:
@jgiardino I fixed 1,2, and 4. I can add the help text in a separate PR. I am unable to replicate your issue with the Image Name field. Also, after talking with Tom, the AWS region is required.

jgiardino:
That's great! I'll review again later today. Do you happen to know why the region is required (i.e. what it's purpose is)? I can also check with Tom, but it would be helpful to include some statement about it in the field level help. 

Also, at the end of [this comment](https://github.com/weldr/cockpit-composer/pull/767#issuecomment-546314301), you mentioned making the region a select menu. I'm fine making that a separate PR. Do you know if these values would be defined as part of the api, or would they be defined in the UI?

jgiardino:
Following up on my previous comment, after chatting with Tom, this is our understanding and plan for the Region field:
- It's associated with the authentication process. The field should display with the authentication step and mention something about it's relationship to the other authentication fields, and maybe something like how the upload action is scoped to this region 🤷‍♀ 
- Values for region can change. For now, we will display the field as a regular text input. The field level help can include a link to the [list of supported regions](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html).

jgiardino:
> @jgiardino I fixed 1,2, and 4. I can add the help text in a separate PR. I am unable to replicate your issue with the Image Name field. Also, after talking with Tom, the AWS region is required.

I reviewed the UI, and see that 1, 2, and 4 were done. I'll investigate why 5 is happening (seems to be specific to a mac, maybe?). And for 3, I'll add a suggestion for help text for the Region field.

jkozol:
This PR is now dependent on #836. I will remove the eslint fixes if it turns out they are unnecessary. This PR will now fail the travis tests but it will fail the integration tests anyways since uploading is not supported by lorax.

larskarlitski:
Rebased on master to include #836.

martinpitt:
Sorry, this needs rebasing after #833, i. e. the tests need to be adjusted to the newer (and simpler) webdriver v5 style.

components/ListView/ListItemImages.js
```diff
@@ -290,6 +291,11 @@ class ListItemImages extends React.Component {
     }
     const size = Math.round(listItem.image_size / 10000000) / 100;
 
+    let uploadsSection;
```
ondrejbudai: The current code cannot be build due to redeclaring `uploadsSection` here.

jgiardino:
I'm testing this out and am having trouble getting a successful image build, but there are a couple of things I have noticed. These might not be related to this PR specifically, but wanted to note them here, since I'm currently looking at this PR. Let me know if you think these are actually separate issues.

In the example below showing RHEL:
- Both of these images have uploads to AWS, but the image type is displaying as Alibaba.

image

In the example below showing Fedora 32:
- The item with status "Image build in progress" seems like a bug. It has an older timestamp, and the Logs just say `Running...`. I really have no idea how to reproduce this issue though.
- When I originally implemented this updated list view for images and uploads, I think I was expecting all images to have an uploads array when uploads were supported. I included the uploads toggle at a time when uploads weren't supported, and used this [line](https://github.com/weldr/cockpit-composer/blob/52103cbbed5f1ba3b26985ba7b5ae082272b3f15/components/ListView/ListItemImages.js#L214) to not show the uploads toggle if there are no uploads defined. But, assuming all images would have an uploads array, even if empty, I implemented this so that the toggle would be included but with a class to visually hide it if no uploads. This is needed to create alignment within the list (e.g. the status messages should align whether uploads exist or not). This can definitely be a separate issue, since it's not critical, but wanted to mention it in case it's a super easy fix.

image

ondrejbudai:
> Both of these images have uploads to AWS, but the image type is displaying as Alibaba

This is a known issue on the osbuild-composer side: https://github.com/osbuild/osbuild-composer/issues/369

ondrejbudai:
I looked into the upload issues, afaik everything works correctly in `cockpit-composer`. However, I found this issue in `osbuild-composer`:

Uploads to `us-east-1` are failing, because this region seems to be currently very slow and the snapshot import part of the upload process is timing out. I created this PR to fix it: https://github.com/osbuild/osbuild-composer/pull/500.

@jgiardino

jgiardino:
Thanks for the update, @ondrejbudai!

I have tested this UI in both the RHEL and F32 auto-deploy instances, and was successfully able to build and upload an image. 🎉

PR #776, opened 2019-09-06
Add patternfly 4

jkozol:
Updates packages and webpack to use patternfly 4 components alongside the existing patternfly 3 elements. The webpack is updated to bundle our stylesheets and handle the additional fonts and images that patternfly 4's react components include.


jgiardino:
Also, I did take a look at Composer in Edge and didn't see any issues in the UI. But looking at the log file for the fedora-30/edge test, it isn't obvious to me what causes the test to fail.

@xiaofeng Do you have any insight into why that test would be failing?

martinpitt:
Sorry, tests are currently broken due to some "must not access npmjs.com any more" faff. I'm fixing it in PR #779, but that needs some work. I'll finish that tomorrow first thing in the morning.

martinpitt:
> as PF4 isn't imported anywhere

Ah sorry, it actually does get imported, but apparently not being used anywhere yet? Or is merely importing the CSS enough to change the UI?

jgiardino:
I tested this yesterday and didn't see any issues. 

@garrett I would love your thoughts on this, especially how we're handling fonts.

garrett:
It would be ideal if we're only referencing once set of fonts in the CSS for:
- the Cockpit login screen
- Cockpit PF3 overrides
- Cockpit PF4
- Composer

The most important part: Even if we do happen to ship font files in multiple places on disk, we should ensure the browser only ever loads from one location.

Unused assets on disk are just a few more megs on disk, but ones that are sent over the network impact performance, caching, and the user experience (by delaying the page more than needed and causing text pop-in).

So, from least ideal to most ideal:

1. Ship multiple copies, reference copies everywhere
2. Ship multiple copies, reference one copy everywhere
3. Ship one copy, reference one copy

I'm happy enough with option 2, but happier with 3. I don't really consider option 1 to be acceptable if we can help it. (But if we _need to_ as an interim… then so be it?)

For option 2 (or even maybe 3), if we have a global PF4 override, we could change the path to the fonts there (as an alternative to using `sed` at the command line).

garrett:
> This PR doesn't increase too much. ~16 mb to ~20mb but when I actually use pf4 components the size about doubles.

Is that when residing on disk, or is that difference being pulled in by browsers also?

If it's sitting on the disk, not doing much of anything, then that's okayish. If it's 4+ more megs being pulled in over the network, it's really quite bad. (We already need to fight bloat.)

jkozol:
@garrett Thanks for the review.  I believe I've updated the way we load fonts to be in line with how cockpit and cockpit's login page load them which is from a single copy under hostname:9090/cockpit/static/fonts. I don't use sed but instead redeclare in our custom css the font family pf4 uses. For example: 
```
@font-face {
  font-family: "RedHatDisplay";
  src: url("../../static/fonts/RedHatDisplay-Regular.woff2");
  src: url("../../static/fonts/RedHatDisplay-Regular.woff2")  format('woff2');
  font-style: normal;
  font-weight: 300;
  text-rendering: optimizeLegibility;
}
```
Also, the size now grows from 16.1 MB to 16.6 MB.

martinpitt:
This is a very nice, clean, and sensible approach now, thanks @jkozol ! But please squash the commits.

PR #779, opened 2019-09-11
test: Use same NPM registry in VM as on the test host

martinpitt:
Our internal CI now has npmjs.com access firewalled off, and sets an
internal NPM registry mirror. Make sure to use that in the nested VM as
well.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -93,7 +93,18 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
         composer.wait_boot()
         composer.set_address("10.111.113.1/20")
         composer.upload(["end-to-end"], "/root/")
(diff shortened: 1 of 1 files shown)


martinpitt:
Without this, the composer tests now [time out on npm install](https://209.132.184.41:8493/logs/pull-12768-20190911-112525-731a0f8d-weldr-cockpit-composer--fedora-30-chrome/log), so this is an urgent fix for restoring testing.

I needed to track down what still uses npmjs.com on our internal machines, and that's one of the cases.

martinpitt:
Argh, this is still falling over the "unknown cert" problem. I'll deal with that tomorrow.

PR #787, opened 2019-09-23
Remove @babel/polyfill, use core-js and regenerator-runtime instead

henrywang:
According to https://babeljs.io/docs/en/next/babel-polyfill, @babel/polyfill has been deprecated in Babel 7.4.0.


henrywang:
Thanks @martinpitt! All green now, merged!

PR #789, opened 2019-09-24
Replaces PF3 list view with PF4 data list on Blueprints page

jgiardino:
This PR replaces the PatternFly 3 list view with the PF4 Data list component on the Blueprints page.

Components inside the list (i.e. dropdown, buttons) were not updated to PF4 components.

Tests were updated where class names were updated.


jgiardino:
FYI - I'm still working on fixing tests.

jgiardino:
I fixed tests and squashed commits. This is ready for review.

test/end-to-end/components/Blueprint.component.js
```diff
@@ -26,7 +26,7 @@ class Blueprint {
   }
 
   get blueprintNameList() {
-    const selector = ".list-group-item-heading";
```
jgiardino: @henrywang I wanted to make sure you saw this change. I included data attributes `[data-description]` above and `[data-blueprint]` here to use instead of class names for the tests. The new class names in the PF4 component are more generic and not as useful for targeting specific contents on the page. I also like data attributes better than class names, because I know these attributes are applied for the purpose of testing. Whereas when we use class names for testing, it isn't obvious when updating the UI which selectors are used for tests until I break the tests :-)

henrywang: NICE!! Thanks @jgiardino!

components/ListView/BlueprintsDataList.js
```diff
@@ -0,0 +1,95 @@
...
+                <PficonTemplateIcon />
+              </div>
+              <DataListItemCells
+                dataListCells={[
+                  <DataListCell key="primary" width="2">
+                    <Link to={`/blueprint/${blueprint.name}`}>
+                      <strong>{blueprint.name}</strong>
+                    </Link>
+                  </DataListCell>,
+                  <DataListCell key="secondary" width="3" data-description>
```
jkozol: Can you change the width to an integer instead of a string i.e. width={2}

components/ListView/BlueprintsDataList.js
```diff
@@ -0,0 +1,95 @@
...
+    super();
+  }
+
+  render() {
+    const { blueprints, layout, ariaLabel } = this.props;
+    const { formatMessage } = this.props.intl;
+    return (
+      <DataList aria-label={ariaLabel} className="cc-m-nowrap-on-lg">
+        {blueprints.map(blueprint => (
+          <DataListItem key={blueprint.name} data-blueprint={blueprint.name}>
```
jkozol: Please add the aria-labelledby attribute to the DataListItem

jgiardino:
Thanks for the review, @jkozol! I fixed those issues, and squashed my commits.

PR #790, opened 2019-09-26
Expose Image Builder on /composer, not /welder

larskarlitski:
@martinpitt do we need to do something special in the spec file to make updates work correctly?


martinpitt:
@jkozol : https://github.com/cockpit-project/bots/pull/134 ought to fix that fedora-31 failure.

martinpitt:
@jkozol: Now fedora-31 [fails](https://209.132.184.41:8493/logs/pull-790-20191014-211313-57ead1ba-weldr-cockpit-composer--fedora-31-firefox/log.html) later on:

```
+ yum install -y nodejs
/var/tmp/vm.install: line 18: yum: command not found
```

That needs to be fixed in cockpit-composer properly.

Landing this PR now, no need to stall it further.

PR #795, opened 2019-10-01
test: Bump RAM of test VMs

martinpitt:
Move the default RAM size of the composer VM from the Makefile to
check-application, so that it applies also when running the latter
directly.

Bump memory size of the composer and the windows-10 VMs to avoid
running out of resources, causing "Couldn't connect to selenium server"
test failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ vm: $(VM_IMAGE)
 
 # run the end to end test
 check: $(VM_IMAGE)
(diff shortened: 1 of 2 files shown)


martinpitt:
This is an attempt to fix failures like https://209.132.184.41:8493/logs/pull-793-20191001-134848-a085ca7b-weldr-cockpit-composer--fedora-30-edge/log. (See https://github.com/cockpit-project/bots/pull/32 for some discussion around this).

henrywang:
@martinpitt, looks more memory resolved this issue. :+1:

PR #796, opened 2019-10-02
Support CI testing against a bots project PR

martinpitt:
* If `$COCKPIT_BOTS_REF` is set, check out that bots version instead of
   master.

 * Use git cache in $XDG_CACHE_HOME if available. Our CI uses that to
   save downloads, and it does not get in the way for local developers.

 * Stop making "bots" a phony target, and drop the now unnecessary
   `[ -d bots ]` check.

Ported from https://github.com/cockpit-project/starter-kit/pull/233

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,8 +114,11 @@ debug-check:
 
 # checkout Cockpit's bots for standard test VM images and API to launch them
 # must be from master, as only that has current and existing images; but testvm.py API is stable
+# support CI testing against a bots change
 bots:
-	[ -d bots ] || git clone --depth=1 https://github.com/cockpit-project/bots.git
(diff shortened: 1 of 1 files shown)


henrywang:
Thanks @martinpitt. One question, the CI running container has `$HOME/.cache/cockpit-project/bots` already?

marusak:
> One question, the CI running container has $HOME/.cache/cockpit-project/bots already?

Replying, since Martin is on PTO. It should have it, but the point of `--reference-if-able` is that if the cache is missing it still works but pulls it from github. (it just limits how many and how much we need to really pull from internet and how much we can just reuse from what we have locally. And we have been using this for some time already with cloning of repositories for testing).

henrywang:
Appreciate your reply, @marusak! Merged.

PR #797, opened 2019-10-02
Fix repeat api calls

jkozol:
Remove api calls from render functions so the same call isn't placed multiple times without waiting for a response.


jgiardino:
Thanks for making these updates! No suggestions from me.

PR #800, opened 2019-10-07
Update deprecated lifecycle methods

jkozol:
In React 17 componentWillMount, componentWillUpdate, and componentWillReceieveProps are deprecated. The functionality triggered by these hooks has been updated and moved to other methods.

### DIFF
--- a/components/ListView/DependencyListView.js
+++ b/components/ListView/DependencyListView.js
@@ -4,9 +4,7 @@ import PropTypes from "prop-types";
 import ListView from "./ListView";
 import ListItemComponents from "./ListItemComponents";
 
-class DependencyListView extends React.Component {
-  componentWillMount() {}
-
+class DependencyListView extends React.PureComponent {
   render() {
     return (
       <div>
@@ -21,7 +19,6 @@ class DependencyListView extends React.Component {
         <ListView className={this.props.className} stacked>
           {this.props.listItems.map(listItem => (
             <ListItemComponents
-              listItemParent={this.props.className}
               isDependency
               listItem={listItem}
               key={listItem.name}
--- a/components/ListView/ListItemComponents.js
+++ b/components/ListView/ListItemComponents.js
@@ -10,22 +10,9 @@ class ListItemComponents extends React.Component {
     this.state = { expanded: false };
   }
 
-  componentWillReceiveProps(newProps) {
-    // compare old value to new value, and if this component is getting new data,
-    // then get the current expand state of the new value as it is in the old dom
-    // and apply that expand state to this component
-    const olditem = this.props.listItem.name;
-    const newitem = newProps.listItem.name;
-    const parent = this.props.listItemParent;
-    if (olditem !== newitem) {
-      if ($(`.${parent} [data-component="${newitem.name}"]`).hasClass("active")) {
-        this.setState({ expanded: true });
-      } else {
-        this.setState({ expanded: false });
-      }
-    }
-    if (this.state.expanded === true && newProps.listItem.dependencies === undefined) {
-      this.props.fetchDetails(newProps.listItem);
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.expanded === true && this.props.listItem.dependencies === undefined) {
+      this.props.fetchDetails(this.props.listItem);
     }
   }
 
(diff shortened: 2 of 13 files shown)


components/ListView/ListItemComponents.js
```diff
@@ -10,22 +10,9 @@ class ListItemComponents extends React.Component {
...
-    const olditem = this.props.listItem.name;
-    const newitem = newProps.listItem.name;
-    const parent = this.props.listItemParent;
-    if (olditem !== newitem) {
-      if ($(`.${parent} [data-component="${newitem.name}"]`).hasClass("active")) {
-        this.setState({ expanded: true });
-      } else {
-        this.setState({ expanded: false });
-      }
-    }
```
jkozol: @jgiardino I believe when the props update (without a remount) the state won't be cleared. So, if the component has a new name and was already set as expanded it will remain expanded without this. Please let me know if I misconstrued the purpose of this method.

henrywang:
@martinpitt, only Travis CI got triggered on this PR, any issues about our CI? Thanks?

martinpitt:
@henrywang : Yes, the pods crash, I'll look at this now.

components/ListView/DependencyListView.js
```diff
@@ -4,9 +4,7 @@ import PropTypes from "prop-types";
 import ListView from "./ListView";
 import ListItemComponents from "./ListItemComponents";
 
-class DependencyListView extends React.Component {
-  componentWillMount() {}
-
+class DependencyListView extends React.PureComponent {
```
KKoukiou: You can also use the arrow functions for that.
I am not suggesting it, just so that it's clear.

components/Pagination/Pagination.js
```diff
@@ -19,26 +19,29 @@ const messages = defineMessages({
...
+  componentDidUpdate() {
     if (this.paginationPage === document.activeElement) {
       this.paginationPage.select();
     }
   }
 
+  static getDerivedStateFromProps(newProps, prevState) {
+    if (prevState.pageValue !== newProps.currentPage) {
+      return { pageValue: newProps.currentPage };
+    } else return null;
```
KKoukiou: Does this work correctly? 
I would expect something like:
  return {}
for zero updates.

components/Toolbar/FilterInput.js
```diff
@@ -3,8 +3,8 @@ import PropTypes from "prop-types";
 import { Filter } from "patternfly-react";
 
 class FilterInput extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
```
KKoukiou: How is this change related?

components/ListView/DependencyListView.js
jkozol: Ahhh so use `const DependencyListView = () => {...}` instead of a PureComponent?

KKoukiou: You can, just stick to some of the two patterns for the whole project basically :)

components/Pagination/Pagination.js
jkozol: I believe returning {} would not cause any updates to the state but per the docs:

> It should return an object to update the state, or null to update nothing

components/Toolbar/FilterInput.js
jkozol: I had been considering setting state from props in the constructor and needed this then but you're correct I don't any more. But, I was reading about why props need to be passed to the constructor and super and came across [this section of the docs](https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class) which says to always bind props to the constructor and super. It doesn't provide a reason but I think it is so this.props is initialized by the constructor and not after it. What are your thoughts?

KKoukiou: Yes, basically if you don't call super(props) in the contructor, this.props would still be undefined between the super call and the end of your constructor. But then react does this this.props = props anyway, so you can access the props from render and other methods later on.
So coming back to this specific change, I don't think it's anyhow adding some functionality. :)

components/Pagination/Pagination.js
KKoukiou: Ok thanks

components/Toolbar/FilterInput.js
jkozol: Okay I have removed it

components/ListView/ListItemComponents.js
jgiardino: This isn't needed. I was looking at this yesterday, and also ended up pulling this out as part of #802.

PR #802, opened 2019-10-08
[WIP] Converts PF3 list view to PF4 data list for components list views

jgiardino:
These updates migrate the PF3 list view to use the PF4 data list for anywhere components are listed.


jgiardino:
@henrywang I'm having trouble getting the tests to pass. I've tried a combination of different selectors in each commit in selectedComponents.component.js, but none of them are working. Do you have any suggestions?

jgiardino:
@henrywang I was able to fix 8 out of 10 failing tests with the last commit. I think I can figure out the last two tomorrow.

henrywang:
Nice work @jgiardino! Any help, just let me know. :+1:

henrywang:
@jgiardino, the following fix should work for you. :-)
```diff
diff --git i/test/end-to-end/components/Details.component.js w/test/end-to-end/components/Details.component.js
index 1b137bde..d47b28df 100644
--- i/test/end-to-end/components/Details.component.js
+++ w/test/end-to-end/components/Details.component.js
@@ -66,7 +66,7 @@ class DetailsComponent {
   }
 
   get depencenciesList() {
-    const selector = `${this.containerSelector} .list-pf-item`;
+    const selector = `${this.containerSelector} .pf-c-data-list__item`;
     browser.waitUntil(() => browser.isExisting(selector), timeout, "No dependencies loaded in component details");
     return $$(selector);
   }
diff --git i/test/end-to-end/components/dependencies.component.js w/test/end-to-end/components/dependencies.component.js
index b74fe361..51aa7719 100644
--- i/test/end-to-end/components/dependencies.component.js
+++ w/test/end-to-end/components/dependencies.component.js
@@ -1,7 +1,7 @@
 // "Dependencies" tab in Edit Blueprint page
 class Dependencies {
   constructor() {
-    this.containerSelector = ".cmpsr-blueprint__dependencies";
+    this.containerSelector = '[data-list="dependencies"]';
   }
 
   loading() {
@@ -11,7 +11,7 @@ class Dependencies {
   }
 
   get depencenciesList() {
-    const selector = `${this.containerSelector} .list-pf-item`;
+    const selector = `${this.containerSelector} .pf-c-data-list__item`;
     browser.waitUntil(() => browser.isExisting(selector), timeout, "No dependencies in Selected Component");
     return $$(selector);
   }
```

components/ListView/ComponentsDataListItem.js
```diff
@@ -0,0 +1,217 @@
...
+      this.props.fetchDetails(this.props.listItem);
+    }
+  }
+
+  render() {
+    const { listItem, noEditComponent, handleComponentDetails, handleRemoveComponent } = this.props;
+    const { expanded } = this.state;
+    const { formatMessage } = this.props.intl;
+    return (
+      <DataListItem key={listItem.name} data-component={listItem.name} isExpanded={expanded} className="cc-component">
```
jkozol: Can you add the prop `aria-labelledby` to the DataListItem element?

components/ListView/ComponentsDataListItem.js
```diff
@@ -0,0 +1,217 @@
...
+              <DataListCell
+                key="secondary"
+                className="cc-component__version pf-u-display-flex-on-xl pf-u-flex-direction-column"
+              >
+                <FormattedMessage
+                  defaultMessage="Version {version}"
+                  values={{
+                    version: <strong>{listItem.version}</strong>
+                  }}
+                  tagName={React.Fragment}
```
jkozol: This throws a prop type console error. What does the tagName prop do to the FormattedMessage?

components/ListView/ComponentsDataListItem.js
```diff
@@ -0,0 +1,217 @@
...
+});
+
+class ComponentsDataListItem extends React.Component {
+  constructor() {
+    super();
+    this.state = { expanded: false };
+    this.handleExpandComponent = this.handleExpandComponent.bind(this);
+  }
+
+  componentWillReceiveProps(newProps) {
```
jkozol: ComponentWIllReceiveProps is going to be deprecated. Can you change this to a different lifecycle method?

components/ListView/ComponentsDataListItem.js
jgiardino: I don't have the link where I found this solution, but I needed the `<FormattedMessage>` to not wrap the translated string in a `<span>`. The extra `<span>` tag was interferring with the flex css applied to the contents of `<DataListCell>`. The prop `tagName` lets you specify a different html element, like `<strong>` or something else, but the suggestion was to use `React.Fragment` to essentially wrap it in `<>`, hence no tag. 

To avoid the console error, I ended up breaking the translated string (i.e. "version") from the value being passed to the string (i.e. the version value).

jgiardino:
@jkozol, thanks for the review! I had totally rebased against your PR when you updated the deprecated lifecycles, and failed to catch that in the new file I created. 🤦‍♀ 

I made updates based on your suggestions. Let me know if there are any other updates needed.

components/ListView/ComponentsDataListItem.js
jkozol: Okay good to know. Thanks for the explanation.

jgiardino:
Thanks, @jkozol! Would you mind triggering the `rhel-7-8/firefox` test for me?

martinpitt:
@jgiardino : Nudged

jgiardino:
@henrywang Do you know why this last test is failing? I don't remember updating any components or tests related to the list of images. And I'm not sure why this would pass for the others and fail for this one.

henrywang:
@jgiardino, the failed case is because the `openssh-server` dependencies on RHEL 7.8 have `tar` included. The selector `strong=tar` will have two elements matched, and the first matched element is the dependence one, not the image type. But the `tar` is not one of dependencies of `openssh-server` on Fedora or RHEL 8. So this issue cannot be found on those two test scenarios.
I'm working on that to fix this issue.

jgiardino:
Thanks, @henrywang!

jgiardino:
@henrywang The rhel-7-8 test passes after your fix. Thank you for updating that! 

Now the rhel-8-1/chrome test is failing. Do you have any suggestions for this one? Should we just rerun the tests, or is there an update needed to make them pass?

jgiardino:
@jkozol Would you review this one more time?

PR #807, opened 2019-10-15
Use dnf instead of yum in Fedora to support Fedora 31 and use command -v instead of which

henrywang:
The `yum` has been removed since Fedora 31.
https://fedoraproject.org/wiki/Releases/31/ChangeSet#Retire_YUM_3
Solution is use `dnf` in Fedora and keep `yum` in RHEL

`which` is non-standard. Use builtin `command -v` instead

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -15,7 +15,7 @@ if type firewall-cmd >/dev/null 2>&1; then
 fi
 
 # Install nodejs if it does not exist
-which node > /dev/null 2>&1 || yum install -y nodejs
+command -v node > /dev/null 2>&1 || $(command -v dnf || command -v yum) install -y nodejs
 
 # disable https in cockpit and use http instead
 printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf


henrywang:
@martinpitt, we can Add `fedora-31` in to replace fedora-30 after this PR got merged.

test/vm.install
```diff
@@ -15,7 +15,11 @@ if type firewall-cmd >/dev/null 2>&1; then
 fi
 
 # Install nodejs if it does not exist
-which node > /dev/null 2>&1 || yum install -y nodejs
+if grep -isq fedora /etc/redhat-release; then
+    command -v node > /dev/null 2>&1 || dnf install -y nodejs
+else
+    command -v node > /dev/null 2>&1 || yum install -y nodejs
+fi
```
martinpitt: Could this rather test for `dnf` being available? RHEL 8 also has that. I. e.
```sh
command -v node > /dev/null 2>&1 || $(command -v dnf || command -v yum) install -y nodejs
```

henrywang: Better then `if`. :+1: Thanks @martinpitt.

martinpitt:
node-gyp needing python2 is an unrelated issue. Let's land this for now.

Issue #804, opened 2019-10-11
cockpit-composer PF4 migration - package lists


larskarlitski:
UI in Composer should be migrated to use PatternFly 4 UI components, to improve consistency with other products and take advantages of accessibility improvements.

As a primary focus for migration, all PF3 list views should be migrated to PF4 data list components. For this proposal, the focus will be on lists showing packages. This includes the following:

- Blueprint and Edit blueprint pages - list view of components / dependencies
- Blueprint and Edit blueprint pages - component details view in a panel - list of dependencies
- Blueprint and Edit blueprint pages - component details in an expanded list item - list of dependencies
- Edit blueprint page - list view of available components

The list of package updates in the Pending Changes modal will be included in a later proposal.

PR #814, opened 2019-10-27
Pf4 components list view secondary

jgiardino:
This PR addresses remaining items in #804 and is based on updates made in PR #802.

The list of available components and list of dependencies in the details of an expanded component are migrated from the PF3 to PF4 datalist component.

### DIFF
--- a/components/ListView/ComponentInputs.js
+++ b/components/ListView/ComponentInputs.js
@@ -28,7 +28,8 @@ const messages = defineMessages({
   }
 });
 
-class ComponentInputs extends React.PureComponent {
+// eslint-disable-next-line react/prefer-stateless-function
+class ComponentInputs extends React.Component {
   render() {
     const { components, label } = this.props;
     const { formatMessage } = this.props.intl;
--- a/components/ListView/ComponentInputs.js
+++ b/components/ListView/ComponentInputs.js
@@ -2,6 +2,15 @@ import React from "react";
 import { defineMessages, injectIntl, intlShape } from "react-intl";
 import PropTypes from "prop-types";
 import shortid from "shortid";
+import {
+  DataList,
+  DataListItem,
+  DataListItemRow,
+  DataListCell,
+  DataListItemCells,
+  Tooltip,
+  TooltipPosition
+} from "@patternfly/react-core";
 import ComponentTypeIcons from "./ComponentTypeIcons";
 
 const messages = defineMessages({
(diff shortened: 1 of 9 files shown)


jgiardino:
@henrywang It isn't obvious to me why the tests are failing. Do you have any suggestions? It looks like it can't find the Add or Remove button in the list of Available Components, but I can't figure out what's causing that failure.

BTW, I was thinking I would change this:
```
    if (n !== 0) {
      this.containerSelector = `[data-list='inputs'] [data-input]:nth-child(${n})`;
    } else {
      this.containerSelector = "[data-list='inputs'] [data-input]";
    }
```
to this:
```
    if (n !== 0) {
      this.containerSelector = `[data-list="inputs"] [data-input]:nth-child(${n})`;
    } else {
      this.containerSelector = '[data-list="inputs"] [data-input]';
    }
```
to be consistent with updates in #802, but I don't think that would cause the test failures.

henrywang:
@jgiardino, the case `Selected package icon should have blue border` failed is a bug.
The package in `Available Components` should have a blue border icon if it's added into blueprint. But it still keeps at `not added` icon.
cockpit-composer-bordered-icon

components/ListView/ComponentInputs.js
```diff
@@ -19,178 +28,91 @@ const messages = defineMessages({
...
-  }
-
-  handleEnterKey(event, component) {
-    if (event.which === 13 || event.keyCode === 13) {
-      const { handleComponentDetails } = this.props;
-      handleComponentDetails(event, component);
-    }
-  }
-
+class ComponentInputs extends React.PureComponent {
```
jgiardino: @jkozol The tests fail because changing this component to `React.PureComponent` results in the component not re-rendering when adding/removing components. 

I can change this line to the following to fix the re-render issue, but am wondering if you have any alternate suggestions.
```
// eslint-disable-next-line react/prefer-stateless-function
class ComponentInput extends React.Component {
```

jgiardino: I went ahead and pushed this change, just to see if the tests will pass.

jkozol: This fixed the issue for me and I think is good for now. I think not being able to use a PureComponent here might indicate we should reconsider how we mutate the list of available components.

henrywang:
@jgiardino, my fix works on this PR as well!

components/ListView/ComponentsDataListItem.js
```diff
@@ -0,0 +1,219 @@
...
+            dataListCells={[
+              <DataListCell key="primary" className="cc-component__name">
+                <div>
+                  <a href="#" onClick={e => handleComponentDetails(e, listItem)}>
+                    <strong id={listItem.name} data-component-name>
+                      {listItem.name}
+                    </strong>
+                  </a>
+                </div>
+                <div>{listItem.summary}</div>
```
jkozol: Again, should we consider translating the summaries?

jkozol: Ahh okay. I wasn't sure.

jgiardino:
@jkozol Whenever you get a chance, would you review this again, and could you also trigger the failed edge test?

jgiardino:
> Done (bots/tests-trigger 814). Are you not allowed to? (you should be, you are in the right GitHub team), or perhaps you are just missing the GitHub token setup?

Thanks, @martinpitt! I had tried this in the past, but don't remember what my issue was, and I didn't capture the error message I was getting. I thought it might be related to using a mac, and am slowly working on transitioning over to a thinkpad running fedora, and was going to try to get this working then.

PR #826, opened 2019-11-14
Adds new labels and updates one label for image output options

jgiardino:
New labels are added for alibaba, google, and hyper-v options.

The label for openstack is updated to match what displays in documentation.

This commit addresses issues noted in
https://bugzilla.redhat.com/show_bug.cgi?id=1769154


jgiardino:
Example of menu options in fedora:

image

Issue #718, opened 2019-07-05
Allow user to fix depsolve errors on Edit page


jgiardino:
### Description of problem
When a depsolve error occurs (i.e. a package version doesn't exist in one of the source repos), there is no way in the UI for the user to fix the issue. 

### Steps to Reproduce

1. In the CLI, edit the TOML file for one of the blueprints, and specify a package version that doesn't exist.
2. In the UI, navigate to the Edit page for that blueprint.

### Actual results

Instead of seeing a list of contents for the blueprint, an error message displays using the EmptyState pattern, and lists the packages that are failing.

### Expected results

1. The same error message should display, but it should use an [inline message style](https://www.patternfly.org/v4/documentation/core/components/alert#inline-alert-types), and not the EmptyState. This message should be somewhere above the list of selected packages (either before or after the tabs, depending on what seems to work best).
2. The tabs for "Selected Components" and "Dependencies" should display. We know what the selected components are, so we should be able to list them in the UI, even though depsolving failed. This will allow the user to easily see where the issues are and fix them.
    1.  In the list of selected components, an error icon could display next to the version number.
        - As part of this update, it might be worth considering how version is represented. E.g. do we display both what the user selected for version and the depsolved version?
          - Version 2.3 (2.x) 
            _displays when the package correctly depsolves; the "x" is used here to represent the wildcard asterisk_ 
          - Version ⚠️(2.x)  
            _displays when depsolve fails; the ⚠️is used here to represent an error icon_
    2. The dependencies tab should be disabled since we can't depsolve. Additionally, an error icon can display instead of the badge with number of dependencies.

### Additional info
The user should be able to then modify the version that is saved for a failing package, and see the error message and indications update to not list that package as failing.

The user should see a similar message on the Blueprint page (that is not the Edit page). In that case, the user should see an action in the inline Error message to navigate to the Edit page to fix the issues. 

**This might be a separate issue:** The user should not be able to create an image when depsolve errors exist. I _think_ the best way to handle this is by letting the user open the Create Image modal, and displaying an error message in the modal. This would work well in cases where the user isn't seeing the list of blueprint packages when they choose to open the Create Image modal. **This might be better as a separate issue/PR,** if we need to also add a depsolving action to occur when the user opens the modal from the list of Blueprints.

PR #830, opened 2019-11-21
Allow user to fix depsolve errors on Edit page

jgiardino:
Before this update, if a blueprint had a package with a version that resulted in a depsolve failure, then an error message would display in the Edit page, and the only workaround for fixing the error is by editing the TOML file through the cli.

This update still displays the error message, but also includes the list of selected components so that the user can find the component that is mentioned in the error and resolve the issue by either removing the component or selecting a different version.

~Still to be included is the last statement in issue #718 (i.e. preventing the user from creating an image when an error exists).~ Going to defer this to a future PR.

Some ways to test this:
1. Add the package "soletta" to a blueprint. In this case, dependencies required by soletta are missing and will return an error. But the list of packages are still displayed. In this case, you will also see the Undo action display in the alert.
2. Edit the toml file of a blueprint to have bad version numbers for packages (or navigate to the Edit page for the atlas blueprint, which has old version numbers defined).

### DIFF
--- a/components/ListView/BlueprintContents.js
+++ b/components/ListView/BlueprintContents.js
@@ -1,7 +1,7 @@
 import React from "react";
 import { defineMessages, injectIntl, intlShape, FormattedMessage } from "react-intl";
 import PropTypes from "prop-types";
-import { DataList } from "@patternfly/react-core";
+import { DataList, Alert } from "@patternfly/react-core";
 import { Tabs, Tab } from "patternfly-react";
 import ComponentsDataListItem from "./ComponentsDataListItem";
 import DependencyListView from "./DependencyListView";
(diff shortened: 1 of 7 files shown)


jgiardino:
This is ready for review. After chatting with @larskarlitski, I'm going to create a separate PR to add the ability to block image creation when a depsolve failure occurs.

Please let me know if any updates are needed.
Thanks!

jkozol:
This looks good to me. I have two things which I think are fine to add in separate PRs. 

1) I think we should display a similar error on the packages tab of the blueprint details page. Maybe it could direct the user to navigate to the edit blueprint page to fix the issue.

2) When I drill down to view the details of a component some of the information is missing and there is console error: `Error in fetchInputDeps TypeError: "response[responseIndex] is undefined"`

atlasDepError

larskarlitski:
Rebased to include Jacob's eslint fixes and fix conflicts.

henrywang:
@larskarlitski, feel free to merge it. All green now.

PR #836, opened 2019-11-29
Update formatting for eslint requirements

jkozol:
Fix the formatting across the project to prevent a build failure due to eslint. I am currently unsure what changed that makes these changes necessary but this PR will build locally for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "@patternfly/react-icons": "3.14.23",
     "bootstrap": "3.4.1",
     "core-js": "3.4.0",
+    "core-js-compat": "3.4.5",
     "fastclick": "1.0.6",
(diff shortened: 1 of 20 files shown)


larskarlitski:
Pushed an update which disabled `prettier` and pins `core-js-compat`.

package.json
```diff
@@ -12,6 +12,7 @@
     "@patternfly/react-icons": "3.14.23",
     "bootstrap": "3.4.1",
     "core-js": "3.4.0",
+    "core-js-compat": "3.4.5",
```
martinpitt: This is obsolete now, at least cockpit's tests run fine again. 3.4.7 now seems to be generally available. That said, don't let that block this PR, as that's a bit urgent.

PR #845, opened 2019-12-10
Use selenium containers on services image

martinpitt:
The old "selenium" image got deprecated.
See https://github.com/cockpit-project/bots/pull/281 for details.


martinpitt:
The real change here is that this will test against a much newer firefox (and also chromium). That got blocked for a while in https://github.com/cockpit-project/bots/pull/164 , but PR #833 landed now that should fix the tests with recent selenium node versions.

PR #846, opened 2019-12-12
[WIP]  Update images list to match design 

jgiardino:
- Updates text strings and moves most of them to the beginning of the file
- Adds image size to the image list item
- Migrates list of images to PF4 data list component and makes layout updates based on [design](https://docs.google.com/presentation/d/1LCvefUG_a3o4M2Cf4xbhtNHEeyJJ2uzfQpmycKEBJHY/edit?usp=sharing)
  - As part of this design, only either the Image logs section or Uploads section is visible at a time
- Moves Logs contents to a separate component for future use with Uploads list


henrywang:
@jgiardino, could you please rebase master or send PR on `rhel-8.2` branch? The test code in master has a lot of changes due to webdriver.io update to v5. The old test code will not work in master right now, but it keeps working on branch `rhel-8.2`. Sorry for inconvenience.

components/ListView/ListItemImages.js
```diff
@@ -63,173 +129,233 @@ class ListItemImages extends React.Component {
...
-      uploadsSection = <ListItemUploads uploads={this.props.listItem.uploads} />;
+      uploadsSection = (
+        <ImagesDataList ariaLabel={formatMessage(messages.imageUploads)}>
+          {listItem.uploads.map(upload => (
+            <ListItemUploads upload={upload} key={upload.uuid} />
+          ))}
+        </ImagesDataList>
+      );
     }
+    const size = Math.round(listItem.image_size / 10000000) / 100;
```
jkozol: I think you should use increments of 1024 here. Also this rounds to the nearest 10 MB correct?

components/ListView/ListItemImages.js
```diff
@@ -63,173 +129,233 @@ class ListItemImages extends React.Component {
...
+    let uploadsToggle;
+    if (listItem.uploads !== undefined) {
+      uploadsToggle = (
+        <DataListToggle
+            onClick={this.handleUploadsShow}
+            isExpanded={this.state.uploadsExpanded}
+            id="uploads-toggle"
+            aria-label={`${formatMessage(messages.imageUploads)} ${this.props.blueprint}-${listItem.version}-${listItem.compose_type}`}
+            aria-controls={`${listItem.id}-uploads`}
+            // ^ need to fix this attribute value
```
jkozol: What needs to be fixed here?

jkozol:
Hey this looks pretty good to me. I am getting a console error when opening the Logs of `Invalid prop `logContent` of type `object` supplied to `Logs`, expected `string`. Also, the in progress spinner is not aligned to the text for me

spinners

components/ListView/ListItemImages.js
jgiardino: Oh, I need to remove that comment. I fixed the attribute (i.e. it references a unique ID of the section in controls), but forgot to remove the comment.

components/ListView/ListItemImages.js
jgiardino: This rounds to the nearest hundredth decimal place. 

I originally thought that it should be increments of 1024, but when I downloaded the file, I noticed that it was the same value in GB as `const size = Math.round(listItem.image_size / 10000000) / 100;`. But this was lorax, and should be confirmed when using osbuild.

jgiardino:
Thanks for reviewing, @henrywang and @jkozol.

@henrywang - This is a work in progress, based on PR #767 which still needs to be rebased against the 8.2 branch. 

@jkozol - Thanks for catching those issues. I'll look into them.

jgiardino:
> Hey this looks pretty good to me. I am getting a console error when opening the Logs of Invalid prop logContentof typeobjectsupplied toLogs, expected string`. 

Thanks for catching this issue. I noticed that it's only caused when no logs are returned. I included a fix for this in my previous commit.

henrywang:
@jgiardino, I think we can close this PR because PR #849 got merged already?

jgiardino:
Good call ;-) Thanks, @henrywang!

PR #849, opened 2019-12-13
Migrates Images list view to the PF4 data list

jgiardino:
This is a duplicate of #846 except that it is based on master instead of #767.

- Updates text strings based on [latest design](https://docs.google.com/presentation/d/1LCvefUG_a3o4M2Cf4xbhtNHEeyJJ2uzfQpmycKEBJHY/edit?usp=sharing) and moves most of them to the beginning of the file
- Adds image size to the image list item
- Migrates list of images to PF4 data list component and makes layout updates based on [latest design](https://docs.google.com/presentation/d/1LCvefUG_a3o4M2Cf4xbhtNHEeyJJ2uzfQpmycKEBJHY/edit?usp=sharing)
- Moves Logs contents to a separate component for future use with Uploads list

### DIFF
--- a/test/end-to-end/specs/imageBuilding.test.js
+++ b/test/end-to-end/specs/imageBuilding.test.js
@@ -58,6 +58,9 @@ describe("View Blueprint Page", function() {
     toastNotificationPage.closeButton.click();
     $(toastNotificationPage.containerSelector).waitForExist(timeout, true);
     viewBlueprintPage.imagesTab.click();
+
+    // show actions menu
+    viewBlueprintPage.imageMoreButton.click();
   });
 
   after(function() {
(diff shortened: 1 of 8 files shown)


test/end-to-end/specs/imageBuilding.test.js
```diff
@@ -119,6 +122,7 @@ describe("View Blueprint Page", function() {
 
   describe("Delete Image Page", function() {
     before(function() {
+      viewBlueprintPage.imagesTab.click();
```
jgiardino: @henrywang I'm not sure whether this line was the right fix for this test or not, but now the `Delete Image Page` tests are passing with this update. 

As part of the PF4 migration, I moved the Download action to the actions menu. This required me to add `viewBlueprintPage.imageMoreButton.click();` above (line 63) so that the Download action could be found. But I think this also caused the same call on the next line (125) to result in dismissing the menu, making the Delete action not visible. Based on this assumption, I added `viewBlueprintPage.imagesTab.click();` here so that if the menu is visible, it is dismissed, so that when the `imageMoreButton` is clicked, it will open the menu again.

martinpitt:
[firefox failed again](https://209.132.184.41:8493/logs/pull-849-20191216-212626-ce3a2fd8-weldr-cockpit-composer--fedora-31-firefox/log.html), retrying

PR #851, opened 2019-12-16
Use selenium containers on services image

marusak:
The old "selenium" image got deprecated.
See https://github.com/cockpit-project/bots/pull/281 for details.


martinpitt:
Firefox fails with this `Error: HTTP method not allowed`. This requires another backport of commit f04103ce2f299d985350d575d1cd8b90b83456a2 . But more realistically, why not just reset the rhel-8.2 branch to master or the upcoming release? We are going to rebase either way, no?

martinpitt:
That said, this backport is already an improvement, so landing.

PR #854, opened 2019-12-17
test: Do not click on `Reuse my password` as it is by default on

marusak:


martinpitt:
We may actually need to conditionalize that for rhel-7 only, or do the test differently to always *check* (not toggle) the checkbox.

PR #855, opened 2019-12-17
test: Run tests by default as admin

marusak:


martinpitt:
Note: This *should* fail until #854 lands.

martinpitt:
Nice, so [this failure](https://209.132.184.41:8493/logs/pull-855-20191217-165454-19b8d587-weldr-cockpit-composer--rhel-8-2-firefox/log.html) is exactly what happens downstream. Setting blocked tag now, as this should land after working out a more general solution for #854.

PR #860, opened 2019-12-26
test: Add retry flaky tests feature

henrywang:


henrywang:
@martinpitt, thanks for review! I agree with you. The purpose of this PR is to save some re-run time, but I don't like the downside just like you said. I'll close this PR.
Anyway, the recent 3 PRs are all one time all green! That means some of flaky issues might be caused by resource I think!

PR #866, opened 2020-01-03
test: Lock webdriverio and all wdio relate version to 5.16

henrywang:
The latest webdriver.io release(v5.18.3) brings a bug https://github.com/webdriverio/webdriverio/issues/4934. This issue will cause 'non W3C properties' error. Solution is locking webdriver.io to 5.16 until issue got fixed.


henrywang:
Duplicated with PR #865.

PR #868, opened 2020-01-06
Switch to osbuild-composer backend for master

jkozol:
This is the same as PR #839 but to merged to master.

Per @larskarlitski 

>osbuild-composer is a drop-in replacement for lorax-composer, providing
the same API:
https://github.com/osbuild/osbuild-composer
Switch the dependency, documentation, and the name of the service for
the "Start Service" page.

PR #871, opened 2020-01-09
cockpituous: remove releases except for fedora-31

jkozol:
As part of the move to using osbuild-composer instead of lorax, we should only release on fedora-31 since the other distros will fail to install osbuild-composer. This PR is required for #868

### DIFF
--- a/utils/cockpituous-release
+++ b/utils/cockpituous-release
@@ -18,11 +18,9 @@ job release-srpm -V
 # Do fedora builds for the tag, using tarball
 job release-koji master
 job release-koji f31
-job release-koji f30
 
 job release-github
 job release-copr @weldr/cockpit-composer
 
 # Create a Bodhi update for stable Fedora releases
 job release-bodhi F31
-job release-bodhi F30


utils/cockpituous-release
```diff
@@ -16,13 +16,10 @@ job release-source
 job release-srpm -V
 
 # Do fedora builds for the tag, using tarball
-job release-koji master
```
marusak: I think you want master (that is rawhide)

martinpitt: Yes, definitively.

jkozol: Fixed now.

martinpitt:
If we do this, we also need to ensure that we don't push a newer lorax-composer into F30 that breaks cockpit-composer, as then we wouldn't have a chance any more to fix it. This is fine, but I wanted to point it out.

bcl:
> If we do this, we also need to ensure that we don't push a newer lorax-composer into F30 that breaks cockpit-composer, as then we wouldn't have a chance any more to fix it.

How could pushing a new lorax-composer break cockpit-composer?

martinpitt:
> How could pushing a new lorax-composer break cockpit-composer?

Not sure if that was a rhetorical question, but it could change API/behaviour and such. If that doesn't happen, then it's fine of course. I'm just interested in how we ensure that the current c-composer keeps working in F30.

bcl:
I see what you mean now. I was thinking that cockpit-composer for F30 was switching to osbuild-composer, and in that case there's no way lorax-composer could effect it.

FWIW there are no planned changes for F30 at this point, only bugfixes if needed.

martinpitt:
Thanks @bcl, that seems fine then.

PR #872, opened 2020-01-10
test: Use both "admin" and "root" as user name to run tests

henrywang:
To cover both root and non-root login scenario, only api related tests use "admin" as login user and the other tests use "root"

### DIFF
--- a/test/end-to-end/wdio.conf.js
+++ b/test/end-to-end/wdio.conf.js
@@ -250,11 +250,15 @@ exports.config = {
       browser.setWindowSize(1280, 1024);
     }
     // login cockpit and enter into composer page
-    browser.login();
-    browser.switchToComposerFrame();
-    // only the first suite needs start lorax-composer
+    // only api related test uses "admin" as user name
+    // the rest of suites will use "root"
     if (suite.title === "lorax-composer api sanity test") {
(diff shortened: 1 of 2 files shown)


henrywang:
Thanks @marusak, good idea. Updated.

test/end-to-end/wdio.conf.js
```diff
@@ -250,11 +250,15 @@ exports.config = {
       browser.setWindowSize(1280, 1024);
     }
     // login cockpit and enter into composer page
-    browser.login();
-    browser.switchToComposerFrame();
-    // only the first suite needs start lorax-composer
+    // only api related test uses "admin" as user name
+    // the rest of suites will use "root"
```
martinpitt: I'd argue for the other way around -- using composer as normal admin seems like the more realistic case, and also the one that more probably fails due to some permissions error (which we want the tests to detect). Does the "api sanity test" also involve the UI? I thought it just did some HTTP call. In that case that doesn't really test anything, as cockpit isn't involved?

henrywang: @martinpitt, it uses cockpit.http and cockpit.script API with `superuser: "required` param. So if login user does not have super user permisisson, test will be failed.

henrywang:
Let's use non-root user to run test. So close this PR.

PR #877, opened 2020-01-20
Set image size

jkozol:
A user may want to manually specify an image size. They may now enter a size into an input field. The size defaults to 2 GB which is also the minimum size allowed. If the size is over 2000 GB an alert is shown explaining that this image may not work on all systems.

This PR is dependent on #868, #767, and https://github.com/osbuild/osbuild-composer/pull/184

### DIFF
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -544,10 +544,6 @@ class CreateImageUploadModal extends React.Component {
             <Text>{formatMessage(messages.review)}</Text>
             {providerSettings[uploadService] !== undefined && (
               <TextList component={TextListVariants.dl}>
-                <TextListItem component={TextListItemVariants.dt}>
-                  <FormattedMessage defaultMessage="Image name" />
-                </TextListItem>
-                <TextListItem component={TextListItemVariants.dd}>{imageName}</TextListItem>
                 <dt>
                   <FormattedMessage defaultMessage="Image size" />
                 </dt>
@@ -590,6 +586,10 @@ class CreateImageUploadModal extends React.Component {
                     </TextListItem>
                   </React.Fragment>
                 ))}
+                <TextListItem component={TextListItemVariants.dt}>
+                  <FormattedMessage defaultMessage="Image name" />
+                </TextListItem>
+                <TextListItem component={TextListItemVariants.dd}>{imageName}</TextListItem>
                 {Object.keys(providerSettings[uploadService].settings).map(key => (
                   <React.Fragment key={key}>
                     <TextListItem component={TextListItemVariants.dt}>
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -416,9 +416,11 @@ class CreateImageUploadModal extends React.Component {
                     GB
                   </div>
                 </div>
-                {!this.isValidImageSize() && (
+                {imageSize < maxImageSize && imageType !== "" && (
                   <div
-                    className="pf-c-form__helper-text pf-m-error"
+                    className={
+                      !this.isValidImageSize() ? "pf-c-form__helper-text pf-m-error" : "pf-c-form__helper-text"
+                    }
                     id="help-text-simple-form-name-helper"
                     aria-live="polite"
                   >
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -20,7 +20,7 @@ import {
   WizardContextConsumer,
   WizardFooter
 } from "@patternfly/react-core";
-import { OutlinedQuestionCircleIcon } from "@patternfly/react-icons";
+import { OutlinedQuestionCircleIcon, ExclamationTriangleIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
 import { defineMessages, FormattedMessage, injectIntl, intlShape } from "react-intl";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
@@ -31,6 +31,11 @@ import { fetchingQueue, clearQueue, startCompose, fetchingComposeTypes } from ".
 import { fetchingUploadProviders } from "../../core/actions/uploads";
 
 const messages = defineMessages({
+  imageSizePopover: {
+    defaultMessage:
+      "Set the size that you want the image to be when instantiated. " +
+      "The total package size and target destination of your image should be considered when setting the image size."
+  },
   infotip: {
     defaultMessage: "This process can take a while. " + "Images are built in the order they are started."
   },
@@ -43,8 +48,8 @@ const messages = defineMessages({
     id: "empty-blueprint-alert",
     defaultMessage: "This blueprint is empty."
   },
-  warningMissingFields: {
-    defaultMessage: "Required information is missing."
+  warningReview: {
+    defaultMessage: "There are one or more fields that require your attention."
   },
   selectOne: {
     defaultMessage: "Select one"
@@ -55,6 +60,16 @@ const messages = defineMessages({
   review: {
     defaultMessage:
       "Review the information below and click Finish to create the image and complete the tasks that were selected. "
+  },
+  warningSizeSmall: {
+    defaultMessage: "Minimum size is {size} GB."
+  },
+  warningSizeLarge: {
+    defaultMessage:
+      "The size specified is large. We recommend that you check whether your target destination has any restrictions on image size."
+  },
+  warningSizeEmpty: {
+    defaultMessage: "A value is required."
   }
 });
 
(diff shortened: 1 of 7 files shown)


bcl:
This needs to make sure that the backend supports it, it looks like it is added to the POST json which means that if the backend is lorax-composer it will just be ignored, possibly surprising the user when their image isn't the size they asked for.

If the osbuild-composer API is going to be different it may be time to think about moving it to API v2 and leave lorax-composer at v1.

jgiardino:
Thanks for working on this! Overall this is great! There are just a couple of things that could be confusing. I provided some suggestions below...

1. When I enter a value that is less than the minimum size, I never see any information that tells me what that minimum value is. I include below the help text that I originally drafted, but this isn't ideal either, since we have a different minimum size per output type. To help with this, I think the popover help text should be included as rewritten below, and the minimum size should display below the field, as illustrated below.
    - original popover help text: "Set the size that you want the image to be when instantiated. Image Builder accepts any value larger than 2 GB. However, the total package size and target destination of your image should be considered when setting the image size. "
    - updated popover help text: "Set the size that you want the image to be when instantiated. The total package size and target destination of your image should be considered when setting the image size."
    - Updated default state:   
    image
    - Updated error state:
     image


2. The error state (i.e. red underline and error icon) displays on the input for the less-than-min size case, but not for the more-than-max size case. But these are also slightly different, if I understand the current implementation correctly. With the less-than-min case, we ignore the user input and use the minimum size. But with the more-than-max case, do we use the user specified value anyway? If that's true, then this is where this request for patternfly would come in handy: https://github.com/patternfly/patternfly/issues/2598, because an error means we don't use the user specified value, and warning means the user specified value is questionable and the user should know that, but we'll use it anyway if that's really what they want. If these cases are as I understand them, then they should be a little more distinct in the feedback they give the user. This is what I'm thinking, but would love some thoughts on this (cc @andreasn):
    - less-than-min case: Keep as is (including the additional text shown in the example above)
    - more-than-max case: Remove the class `pf-m-error` from the helper text so that the text is not red. Include a warning icon (`<ExclamationTriangleIcon />` with `color: var(--pf-global--warning-color--100);`) in front of the helper text (until that PF issue is closed). Here's an example:     
      image
      - NOTE that in this case, I don't think the statement "Minimum size is x GB." is needed and can be omitted from the helper text.

andreasn:
I have concerns about the behaviour where we override user specified values. If you set 5 GB, you probably did for a reason (keeping your AWS S3 bills down possibly?). But since 6 is the minimal in this case, I think we should probably block progress, so they have the possibility to do error fixing, and get a fuller understanding of minimal size.

andreasn:
If you do keep the error text, I feel the text itself is a little complicated to read (for a non-native English speaker at least.). I don't know if it's something about the structure, or maybe that it uses _minimal_ and _smallest_ interchangeably.

jkozol:
@jgiardino Thanks for the detailed feedback. I made your recommended changes and also disabled the next button if the image size is too small per the recommendation of @andreasn.

imageSizePopover
imageSizeErrorMin

jgiardino:
Thanks for the review, @andreasn! What you're suggesting with blocking progress makes sense. 

Currently, in simple modal forms (not wizards), we disable the submit button when error states exist, rather than display an alert above the form on submit. For now, I would suggest continuing with that pattern if the user is not uploading an image. But it would be worth discussing whether we should switch to the pattern described in the [PF docs for Validation on Submission](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#validation-on-submission) (i.e. the submit button is not disabled, but when clicked, displays the inline error described in # 1 above the form).

Additionally, in the Create Image modal/wizard, we only display a Review step if the user is uploading an image. I provided some ideas below on how to handle what you're suggesting for each flow (with or without upload).

1. If one or more fields has an **error** state:
    1. **In the flow with upload,** display an inline error message at the top of the Review step. The reasons for an error state could be that required information is missing or the provided value is bad and needs to be corrected. Currently, we have "Required information is missing" display as a Warning alert. I suggest changing this to the Error alert, and changing the message to something more generic like: "There are one or more fields that require your attention." This way, the message can display regardless of what the error is (i.e. required field is missing input, or user-specified value is not usable).
        image
    2. **In the flow without upload,** given my comments above, we just disable the "Create" button.
2. If the Image size value is too small or larger than recommended:
    1.  In the form field, display the messages described in the previous comment.
    2. **For the less-than-min case only,** follow the recommendations described for # 1 above (the more-than-max case is just a warning, not an error).
    3. **In the flow with upload,** display the value and the message as shown in this example on the Review step: 
        image
          - In the first example, I modified the less-than-min message in the example above. Is this better? 
3. **In the flow with upload,** we currently do not display an error icon and message for missing values in the Review step. For consistency with the updates described above, we should also display an error message if a value is missing (for Image size and any other required field):
    image

@jkozol I can share react code for the examples above. But I'm also curious what your thoughts are regarding scope of these additional updates relative to this PR. Let me know if you think any of these should be captured in a separate issue.

jgiardino:
I completely missed your comment before my last one. Great progress on this! 

- Regarding my comment 2.iii. above, given that you're handling this case in your latest commit, I'm fine with moving this to a separate issue. But eventually we should implement this to have consistency (i.e. Next is always enabled, and the Review step is where we stop the user from moving forward).
- There are some things I'm noticing related to help text that displays in a popover:
  - The help text icon should be wrapped in a button with an aria-label for accessibility:
    `<Button variant="plain" aria-label="[field label] help">[help icon]</Button>`
  - The button should precede the form input, also for accessibility. This is so that screen reader users discover the help text before the form input, and are aware that it's available before they reach the input. I have examples of how to implement this in the prototype.

andreasn:
_Minimum size is 6 GB_ is clear enough as a label by itself I feel.

andreasn:
> But it would be worth discussing whether we should switch to the pattern described in the PF docs for Validation on Submission (i.e. the submit button is not disabled, but when clicked, displays the inline error described in # 1 above the form).

For this simple case, I think either interaction model is fine, but depends on what's best for the other screens.

andreasn:
If I recall correctly, Cockpit tends to roll strongly with the Validation on Submission pattern. We ran into some really bad cases with Validation on Focus a couple of years ago, and was scared off by that :)

jgiardino:
These updates look great! Here are a few things I noticed:

- After an image type is specified, the "Minimum size is x GB" help text should always display (except for the case where the user-specified value is larger than recommended). It only displays with the pf-m-error class if the value is less than the minimum value. Currently, it displays for the less-than-min error case, but not for other value. 
- The order of fields in Review step should match order they're presented in the steps prior to review step
- In the Review step, any required field that is missing a value should display the message "A value is required." I see this display for the image size field, but not the upload fields.
- For the non-upload flow, we disable the Create button when the specify a value that's too small, but not when the value is missing.
  - I'm actually curious what everyone things we should do here. We currently block on the Review step, as stated above. We could simply state that the default value (i.e. the minimum size) will be used on the Review step to minimize user effort. But we can also refine this flow in a future PR and keep this behavior as it currently is.

jkozol:
> After an image type is specified, the "Minimum size is x GB" help text should always display (except for the case where the user-specified value is larger than recommended). It only displays with the pf-m-error class if the value is less than the minimum value. Currently, it displays for the less-than-min error case, but not for other value.

> The order of fields in Review step should match order they're presented in the steps prior to review step

These are fixed.

> In the Review step, any required field that is missing a value should display the message "A value is required." I see this display for the image size field, but not the upload fields.

I believe this is outside the scope of the PR. I agree that we should support these errors. I am in the process of refactoring this wizard and these alerts will be easier to add after/with the refactor. I created issue #939 to capture this.

> For the non-upload flow, we disable the Create button when the specify a value that's too small, but not when the value is missing.

The button is now disabled. I think it makes sense to disable it when there is no value specified. When the user selects an image type the image size is set to the default/min value. The only way them to have an empty value and a disabled `Create` button is if they delete the default value. I am happy to improve this flow in a future PR.

henrywang:
@jkozol, @jgiardino, please do not merge this PR because the test failed. It's caused by this PR. I'll push fix to @jkozol's branch. Thanks!

henrywang:
@jkozol, I have a question about image size. What does the image size mean for tar image? Form my understanding, the image size doe not make sense to tar. Any thoughts? Thanks!

components/Modal/CreateImageUpload.js
```diff
@@ -405,23 +530,46 @@ class CreateImageUploadModal extends React.Component {
...
               </Popover>
             </Title>
             <Text>{formatMessage(messages.review)}</Text>
             {providerSettings[uploadService] !== undefined && (
               <TextList component={TextListVariants.dl}>
+                <dt>
+                  <FormattedMessage defaultMessage="Image size" />
+                </dt>
+                <dd>
+                  {imageSize === "" && (
```
henrywang, line 551: @jkozol, I tried from UI, `A value is required.` warning message was never got displayed when I made image size input box empty.

jkozol: @henrywang Thanks for fixing this!

jkozol:
@henrywang 

>  I have a question about image size. What does the image size mean for tar image? Form my understanding, the image size doe not make sense to tar. Any thoughts? Thanks!

You are correct. It doesn't make sense to specify the image size for tar. I will create an issue and remove this soon.

henrywang:
@jkozol, I think we can land this PR fow now.

PR #890, opened 2020-02-15
tests: Disable session timeout

marusak:
Some tests take too long (like when image is being built) and cockpit
session times out.
Also it is very likely that selenium tests do not bump the timeout with
their actions.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -24,6 +24,9 @@ fi
 # disable https in cockpit and use http instead
 printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf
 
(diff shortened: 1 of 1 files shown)


martinpitt:
It seems the composer tests have  a hard time getting green even on the current images..

martinpitt:
I retried a few times, and it keeps failing. @henrywang , is this known? This smells like a regression in fedora 31 itself, perhaps a new osbuild version or so?

martinpitt:
PR #891 has the same error, edge test succeeded, and this does not functionally affect composer itself, so I think it's safe enough to land (and in fact should help test stability).

PR #896, opened 2020-02-19
test: Update source repo URL to sway-copr because the fish repo URL does not work

henrywang:


martinpitt:
@henrywang : Can you please send this PR against the rhel-8 branch as well?

PR #893, opened 2020-02-18
Dummy PR to test CI test scenario on branch rhel-8

henrywang:

PR #897, opened 2020-02-19
test: Update source repo URL to sway-copr because the fish repo URL does not work

henrywang:


martinpitt:
Tests look good, but I wonder why this only ran the Fedora tests.. That's not what [testmap](https://github.com/cockpit-project/bots/blob/master/task/testmap.py) says..

marusak:
> I wonder why this only ran the Fedora tests.. That's not what testmap says..

Interesting... It is possible to change base branch once PR is opened (`Edit` button next to the title) so I am wondering if the branch was not changed once PR was opened? If that is not the case, this needs some investigation

martinpitt:
I didn't know that it was possible to adjust the branch target after opening.. @henrywang , is that what  happened?

henrywang:
@martinpitt, but the test map works in PR #893. Interesting...

martinpitt:
I rebased rhel-8 branch to release 12.1, so this is obsolete now.

PR #905, opened 2020-03-05
test: Always explicitly install lorax-composer

martinpitt:
In Fedora, golang-github-osbuild-composer Provides: lorax-composer, but
does not currently fully replace it (e. g. it doesn't provide
lorax-composer.service). Also, we want to be explicit about which
backend we test against. So always explicitly install lorax-composer for
now.

In the future, we can add another scenario that tests against osbuild,
but that needs slightly different setup.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -3,10 +3,8 @@
 # The application RPM will be installed separately
 set -eux
 
-# Update lorax-composer using the updates-testing repository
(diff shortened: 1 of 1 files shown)


martinpitt:
This should fix the failures seen in https://github.com/weldr/cockpit-composer/pull/904 and https://github.com/cockpit-project/bots/pull/603

PR #909, opened 2020-03-14
package.json: Update regenerator-runtime package dependency

cockpituous:
Check for any outstanding update of point version of any outstanding NPM package dependency
 * [ ] npm-update


cockpituous:
Please manually check features related to these files:

```
main.js:// redux-saga uses generators, use regenerator-runtime/runtime to transform generators
main.js:import "regenerator-runtime/runtime";
package.json:    "regenerator-runtime": "0.13.5",
```

PR #913, opened 2020-03-19
test: Adjust report download strategy

henrywang:
1. Always download junit report because it's required by gating
2. Download timeline report for test failed only

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -147,7 +147,11 @@ def run_e2e(verbose, image, browser, cpus, memory, sit):
         # Download report and screenshot to environment $TEST_ATTACHMENTS
         report_dir = os.environ.get("TEST_ATTACHMENTS", os.getcwd())
         os.makedirs(report_dir, exist_ok=True)
-        composer.download_dir("/root/end-to-end/wdio_report", report_dir)
+        # Always download junit report required by gating
+        composer.download("/root/end-to-end/wdio_report/xunit_report.xml", report_dir)
+        # Download timeline report for failed only
+        if not success:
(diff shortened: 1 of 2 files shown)


martinpitt:
If that produces enough useful data on failure, that sounds great! However, it sounds like you would then only get a screenshot when the error happens, not  for the steps that led to it? So I was wondering if e. g. the screenshots could be removed once a test succeeds.

But I think you should also adjust test/check-application to only download and publish wdio_report on a failure? That's actually the more important part, and presumably also much easier to achieve?

henrywang:
@martinpitt, Here's a failed case, and [timeline report](https://logs.cockpit-project.org/logs/pull-913-20200319-142814-1d6e5257-fedora-31-edge/) is downloaded. Alll successful test will only have junit report downloaded due to required by gating.

martinpitt:
@henrywang : Structurally that's exactly what I had in mind, great! Now the timeline HTML is small. However, the screenshot links are now broken:

    src="/root/end-to-end/wdio_report/timeline/file-1584629208535.jpeg"

These need to be adjusted to just "./file-*" so that they actually work outside of the test VM.

If that's too hard to do, then I'd say just flip it around -- embed the images again and stop downloading the *.jpeg files. It's less flexible, but it's not worth wasting a lot of time on this.

henrywang:
Good catch! Thanks @martinpitt!
I sent a [PR](https://github.com/QualityOps/wdio-timeline-reporter/pull/41) to this plugin repo to fix this issue. For now I'll change back to embed enabled and copy HTML report only when test failed.

PR #918, opened 2020-03-27
Remove zanata configuration from .travis.yml

henrywang:
There's a deploy job to push to zanata in Travis. It caused Travis failed when when PR is merged into master. Since we do not use zanata any more, zanata setting should be removed.
Failed deploy job can be found from https://travis-ci.org/github/weldr/cockpit-composer/jobs/667298469?utm_medium=notification&utm_source=github_status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,28 +24,12 @@ install:
             npm install
         fi
 env:
-  global:
-    - CXX=g++-4.8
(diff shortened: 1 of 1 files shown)


henrywang:
@martinpitt, @marusak, The edge scenario failed is not caused by this PR and I'm working on PR #917 to fix flaky issue. Thanks!

PR #919, opened 2020-03-28
Rename the osbulid-composer dependency

teg:
As of F32, the osbuild composer package no longer has the
golang-github prefix in its name. Both the new and old package
names have worked interchangibly since v3 (using Provides) and
will continue to do so until F34, but let's drop this wart sooner
rather than later.

Signed-off-by: Tom Gundersen <teg@jklm.no>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ On Fedora or Red Hat Enterprise Linux:
 * First install Cockpit on your local machine as described in: https://cockpit-project.org/running.html.
 * Next install and start osbuild-composer:
 ```
-    $ sudo yum install golang-github-osbuild-composer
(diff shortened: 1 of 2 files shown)


henrywang:
@jkozol, @teg, this change will cause test fail on fedora-31, because no `osbuild-composer` on fedora-31. If we merge this PR, the CI will drop fedora-31 testing and only support fedora-32 scenario.

ondrejbudai:
On 31 we have `Provides: osbuild-composer` in the spec file, so it should work alright.

henrywang:
Thanks @ondrejbudai! Make sense to me. Let's land this PR because it passed both fedora-31 and fedora-32 testing.

Issue #921, opened 2020-04-01
Use cockpit's testing tools in cockpit-composer


jkozol:
**Problem**

The Selenium WebDriver based tests are difficult to maintain and will frequently fail for unknown reasons. This can be fixed by migrating our tests to cockpit's testing framework. This should provide increased stability and control over our test. Ideally, we will support both selenium and cdp tests for a period of time while the tests are gradually migrated over.

**Tasks**

- [x] Pull cdp and the files it is reliant on from cockpit into cockpit-composer
- [x] Use cockpit bots to configure a container to run cockpit-composer
- [x] Configure CI to run tests using both Selenium and cdp
- [x] Migrate tests to cdp
- [x] Remove Selenium from CI

PR #931, opened 2020-04-16
Move selenium to cdp

henrywang:
Follow issue #921


Makefile
```diff
@@ -3,8 +3,8 @@ VERSION=$(shell $(CURDIR)/rpmversion.sh | cut -d - -f 1)
 RELEASE=$(shell $(CURDIR)/rpmversion.sh | cut -d - -f 2)
 PACKAGE_NAME := $(shell awk '/"name":/ {gsub(/[",]/, "", $$2); print $$2}' package.json)
 TEST_OS ?= fedora-31
-BROWSER ?= firefox
-export TEST_OS BROWSER
+TEST_BROWSER ?= chromium
```
marusak, line 6: There is no need for this, tests already know what is the default

Makefile
```diff
@@ -102,25 +102,26 @@ local-clean:
 	rm -rf test/images tmp $(PACKAGE_NAME).spec $(PACKAGE_NAME)*.rpm $(PACKAGE_NAME)*.tar.gz test/end-to-end/wdio_report
 
 # build VMs
-$(VM_IMAGE): local-clean rpm bots
+$(VM_IMAGE): local-clean rpm bots test/common
 	# VM running cockpit, composer, and end to end test container
 	bots/image-customize -v \
-		-r "echo $(BROWSER) > /tmp/BROWSER" \
+		-r "echo $(TEST_BROWSER) > /tmp/BROWSER" \
```
marusak, line 108: I think this is obsolete. Or do you need this for some reason still?

Makefile
```diff
@@ -130,11 +131,18 @@ bots:
 	if [ -n "$$COCKPIT_BOTS_REF" ]; then git -C bots fetch --quiet --depth=1 origin "$$COCKPIT_BOTS_REF"; git -C bots checkout --quiet FETCH_HEAD; fi
 	@echo "checked out bots/ ref $$(git -C bots rev-parse HEAD)"
 
+# checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
+# when you start a new project, use the latest release, and update it from time to time
```
marusak, line 135: This second sentence can be removed :)

test/verify/check-service
```diff
@@ -0,0 +1,129 @@
...
+import string
+import random
+from hashlib import sha256
+from pathlib import Path
+
+import parent
+from testlib import *
+
+
+@nondestructive
```
marusak, line 18: You already made those tests nondestructive? Nice!!

test/verify/check-service
```diff
@@ -0,0 +1,129 @@
...
+        # create
+        b.click("#create-blueprint-modal-create-button")
+        b.wait_not_present("div[role=dialog] #cmpsr-modal-crt-blueprint")
+
+        # wait for all available compontents visible
+        with b.wait_timeout(300):
+            b.wait_present("ul[data-list=inputs]")
+        # search for openssh-server pacakge
+        b.set_input_text("#cmpsr-blueprint-input-filter", "openssh-server")
+        b.key_press("\r")
```
marusak, line 95: You should be able to merge this into previous command:
`b.set_input_text("#cmpsr-blueprint-input-filter", "openssh-server\n")`

test/verify/check-service
```diff
@@ -0,0 +1,129 @@
...
+        with b.wait_timeout(120):
+            b.wait_text("#openssh-server-input", "openssh-server")
+        # add it to blueprint
+        b.click("li[data-input=openssh-server] .fa-plus")
+        # wait for package added
+        b.wait_present("#blueprint-tabs")
+        b.wait_present("li[data-component=openssh-server]")
+        # click commit button
+        b.click(".cmpsr-header__actions .btn-primary")
+        b.wait_visible("#cmpsr-modal-pending-changes")
```
marusak, line 105: `click()` does `wait_visible`. So this command should not be needed as you do `click` as the next command.

test/verify/check-service
```diff
@@ -0,0 +1,129 @@
...
+        # click commit button
+        b.click("#cmpsr-modal-pending-changes .modal-footer .btn-primary")
+        b.wait_not_present("#cmpsr-modal-pending-changes")
+
+        # wait for toast notification dialog
+        b.wait_present("#cmpsr-toast-committed .pficon-ok")
```
marusak, line 111: as well, `wait_present` followed by `click` is not needed.

test/verify/run-tests
```diff
@@ -0,0 +1,234 @@
+#!/usr/bin/python3
```
marusak, line 1: So there is only one verify file, so this is not needed. But I guess you plan to extend another tests in the future, right?

test/verify/check-service
henrywang: The notification dialog has to change to OK status, not running. This wait has to be here. :)

Makefile
henrywang: ack. Thanks!

Makefile
henrywang: removed! Thanks!

test/verify/check-service
henrywang: Yes.

test/verify/check-service
henrywang: Got it. Nice cdp feature!

test/verify/check-service
henrywang: make sense to me. Thanks!

test/verify/run-tests
henrywang: Yes, more files are on the way. :)

test/verify/check-service
marusak: oh right, these two selectors are not the same :)

test/verify/check-service
henrywang: @marusak, looks `\n` not work here. https://logs.cockpit-project.org/logs/pull-931-20200416-140325-24c194ee-fedora-31-firefox/log.html

marusak: yeah, because it is not in the `keyMap` in `test/common/testlib.py`.
But you used `\r` before, why you try to use `\n` now?

marusak: oh because of my example. Sorry! That was just typo, `\r` is what you need.

henrywang: @marusak, `\r` does not work in this case...
https://logs.cockpit-project.org/logs/pull-931-20200417-051758-9ac4e8cc-fedora-31-firefox/log

marusak: please take a look at definition of `set_input_text`. You will notice that it is waiting by default for the value to be the same. Which of course is not the case here, so you just need to tell it not to check the value.

henrywang: @marusak, got it. Thanks!

test/verify/check-blueprint
```diff
@@ -0,0 +1,142 @@
...
+import testlib
+
+
+@testlib.nondestructive
+class TestBlueprint(testlib.MachineCase):
+    def setUp(self):
+
+        super().setUp()
+        for message in parent.allowed_journal_messages:
+            self.allow_journal_messages(message)
```
marusak, line 19: it takes n arguments for patterns, so you can just unpack the array and don't need to iterate over it :)

test/verify/check-image
```diff
@@ -0,0 +1,163 @@
...
+# checkpoint:
+# 1. create supported image
+
+
+import parent
+import testlib
+
+
+@testlib.nondestructive
+@testlib.timeout(2400)
```
marusak, line 12: always better to have this only for command which may actually take that long as when things get really broken (like some new changes and tests needs adjusting) then debugging test with such timeout is "fun" :)
You can do:
```
<command that should be quick>
with b.wait_timeout(2400):
   <command1 that takes long>
   <command2 that thakes long>
<command that should be also quick>
```

test/vm.install
```diff
@@ -22,22 +22,27 @@ printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf
 # disable session timeout as some tests wait for too long
 printf "[Session]\\nIdleTimeout=0\\n" >> /etc/cockpit/cockpit.conf
 
+cat /etc/cockpit/cockpit.conf
```
marusak, line 25: Is this here for some reason or just for debugging?

marusak:
Looks amazing @henrywang ! good job. Is this ready to land or is it still WIP?

test/vm.install
henrywang: oh, good catch! It's just for debugging, because I found an error screenshoot to cockpit login page, so I just suspect the configuration `cockpit.conf` does not work.

test/verify/check-blueprint
henrywang: Thanks. Updated!

test/verify/check-image
henrywang: From the [code](https://github.com/cockpit-project/cockpit/blob/9b786edbec00af2164f7e91df61329cc33044096/test/common/testlib.py#L1375), the `@testlib.timeout()` is different from `browser.wait_timeout()`. The first one is the whole test(method)/suite(class) scope. Since the image building needs take a long time, so I have to set the test scope timeout. And it should be bigger than `browser.wait_timeout()` which I configured in [test](https://github.com/osbuild/cockpit-composer/pull/931/files#diff-52613cf5f87fb672c28b7c6b27f4e655R126).
So I set `browser.wait_timeout()` to `30 minutes` and `@testlib.timeout()` to `40 minutes`.

henrywang:
Thanks @jkozol! Building tar image locally is really fast, say takes 590 seconds `# 2 TESTS PASSED [590s on toolbox]`. I'll give more cpu and memory resource to test vm and give it a try.

henrywang:
> Looks amazing @henrywang ! good job. Is this ready to land or is it still WIP?

I'm going to land this PR first because the code coverage has raised to 68% (75% for selenium). And more cases are under migrating and will be sent in the future PR.
Really appreciate your help! :+1:

test/verify/check-image
marusak: Oh, of course, you are right. Please ignore :)

henrywang:
@marusak, the fedora-31 test got "cockpit logout due to inactivity" screenshoot again. `IdleTimeout=0` has been configured already. Do you have any idea on that? Thanks.

marusak:
> @marusak, the fedora-31 test got "cockpit logout due to inactivity" screenshoot again. IdleTimeout=0 has been configured already. What happened? Do you have any idea on that? Thanks.

Hmm, not really. This should work and if it logs you out even with this configuration, than it seems to be a a bug in Cockpit. Can you please retry if this will happen again or it was one off? (Also can you reproduce locally?)

henrywang:
@marusak, Yes, it's reproduced again, and logout after 900 seconds.
https://logs.cockpit-project.org/logs/pull-931-20200423-044817-9aa4cb80-fedora-31/log.html

test/verify/check-blueprint
```diff
@@ -0,0 +1,140 @@
...
+import parent
+import testlib
+
+
+@testlib.nondestructive
+class TestBlueprint(testlib.MachineCase):
+    def setUp(self):
+
+        super().setUp()
+        self.allow_journal_messages(",".join(parent.allowed_journal_messages))
```
marusak, line 17: this is wrong. It will pass just one string. You want to do `self.allow_journal_messages(*parent.allowed_journal_messages)`

henrywang: ack! Thanks.

marusak:
> Yes, it's reproduced again, and logout after 900 seconds.

hmm, I did `make vm` and then just ran random test with `TEST_SHOW_BROWSER=1` and `teslit.sit()` inside it and it did not timeout.
Now doing the whole `make check`, lets see if there is any difference. (that is what CI does eventually)

marusak:
random idea - maybe nondestructive tests somehow affect this? (we surely do some cleanups of cockpit.conf in between non-d tests, let me see what kind of changes)

marusak:
```self.machine.write("/etc/cockpit/cockpit.conf", "[Log]\nFatal = criticals\n")``` replaces this file.
``` self.addCleanup(m.execute, "rm -f /etc/cockpit/cockpit.conf")``` removes this after each test.

**Comment by marusak** [comment]
> ```self.machine.write("/etc/cockpit/cockpit.conf", "[Log]\nFatal = criticals\n")``` replaces this file.
``` self.addCleanup(m.execute, "rm -f /etc/cockpit/cockpit.conf")``` removes this after each test.

Ha indeed, that is it! But this only happens with `test/common` version of 217 and newer, but Makefile requires 216 (thats why I could not reproduce it).
What does `make machine` do? What is that `rsycn` doing? What is it syncing with? Is it pulling newer versions of `test`?

So this is unlucky, I think our tests should count on edited `cockpit.conf`. Simple workaround is to edit this file in `SetUp`.

**Comment by henrywang** [comment]
@marusak. The `rsync` is to copy test required file/folder in bots into test folder. Test folder will be packaged into tar-ball. And SRPM will have test code and its dependence.
That'll avoid `git clone/fetch` in gating (tier-0 and tier-1) test environment. So test and its dependence code is able to `sync` with production code.

**Comment by marusak** [comment]
Ah thanks @henrywang for the explanation!
I took a second look why it could then fail and indeed the setting of `[Log]` comes only in 217, but the cleanup is already in 216. That means that first tests runs, when it finishes it removes `cockpit.conf`, then runs another tests but it does not have this file anymore.
I'll discuss this tommorow with Martin to see, if this is the right approach or we should be a bit more careful about his.
For now you should be able to just set it up in `SetUp` and possibly you just may need it in those tests, that might actually sit for 15 minutes around, no need to do it in tests, which should not do any action for such a long time.

Sorry for any inconveniences!

**Comment by henrywang** [comment]
@marusak, could you please review my new update if you have time? Thanks! I collaborate your suggestion with my thoughts. Do not remove anything, just add setup and clearup code for c-composer. So most of test file `check-*` do not need call `setUp()` any more.

File: README.md [hunk 13]
```diff
@@ -55,9 +52,23 @@ Cockpit Composer is built using [React](https://reactjs.org/). For inspecting th
...
+
+In the event you wish to visually see what the browser is doing you will want to run:
+
+    $ TEST_SHOW_BROWSER=true make check
+
+By default the cockpit-composer will be installed into Fedora 31 and test will be run on Chromium. To run it on Firefox, a prefix ```TEST_BROWSER=firefox``` needs to be added, like:
+
+    $ TEST_BROWSER=firefox make check
+
+To test cockpit-composer in different OS, set the `$TEST_OS` environment variable, for example:
```
marusak, line 69: A few comments on the readme:
1. `In the event you wish to visually` -> `In the case you wish to visually`
2. There is also `make debug-check` which does the same. Worth mentioning it here or dropping it?
3. `will be installed into Fedora 31` - that is true now, but not sure if you want to keep this up to date all the time. Maybe better to point to `TEST_OS` in Makefile.
4. `To run it on Firefox, a prefix` - it is not a prefix, it is an `environment variable`.

test/README.md
```diff
@@ -0,0 +1,104 @@
...
+debug for test development.
+
+## Requirement
+
+For testing, the following dependencies are required:
+
+    $ sudo dnf install curl expect xz rpm-build chromium-headless \
+        libvirt-daemon-kvm libvirt-client python3-libvirt
+
+And `chrome-remote-interface` and `sizzle` Javascript librarys need to be installed:
```
marusak, line 20: libraries

test/README.md
```diff
@@ -0,0 +1,104 @@
...
+instance.
+
+    $ ./test/verify/check-image --trace --sit
+
+Normally each test starts its own chromium headless browser process on a separate random port. To
+interactively follow what a test is doing, start the browser manually and tell the test which debug
+port it should attach to:
+
+    $ chromium-browser --remote-debugging-port=9222 about:blank
+    $ TEST_CDP_PORT=9222 ./test/verify/check-image --trace
```
marusak, line 53: This is correct, but I would advise to use `TEST_SHOW_BROWSER` as you need to make sure, you have correctly set up the browser (and also that HOME is clean enough).

test/verify/check-custom
```diff
@@ -0,0 +1,264 @@
...
+        b.wait_attr(".form-control-pf-save", "disabled", "")
+
+        # collect code coverage result
+        self.check_coverage()
+
+    def testUserAccount(self):
+        b = self.browser
+        m = self.machine
+
+        # work around "keystroke sent to browser" != "react component accepts it"
```
marusak, line 151: It would be great if you would open issue on c-composer and put here `#HACK: cockpit-composer may loose keystrokes, see #<issue_number>`

test/verify/check-image
```diff
@@ -0,0 +1,157 @@
...
+    def testBasic(self):
+        b = self.browser
+
+        self.login_and_go("/composer", authorized=True)
+        b.wait_present("#main")
+
+        # create image wizard (no upload support)
+        b.click("li[data-blueprint=httpd-server] #create-image-button")
+        b.wait_text("#create-image-upload-wizard #blueprint-name", "httpd-server")
+        # check popover, content, and x button (does not work)
```
marusak, line 23: Also `#HACK <issue>` would be great so it actually may get fixed one day :)

test/verify/check-image
```diff
@@ -0,0 +1,157 @@
...
+        test_selector = "#image-type option[value='tar']"
+        b.wait_present(test_selector)
+        value_id = b.attr(test_selector, "value")
+        b.set_val("#image-type", value_id)
+        b.wait_val("#image-type", value_id)
+        # group actions end
+        b.click("#continue-button")
+        b.wait_not_present("#create-image-upload-wizard")
+
+        # tosta notification
```
marusak, line 101: toast?

test/verify/composerlib.py
```diff
@@ -0,0 +1,72 @@
...
+allowed_journal_messages = [
+    "http:///run/weldr/api.socket/api/v0/.* couldn't connect: Connection refused",
+    "polkit-agent-helper-1: pam_authenticate failed: Authentication failure",
+    "We trust you have received the usual lecture from the local System",
+    "Administrator. It usually boils down to these three things:",
+    ".*Respect the privacy of others.*",
+    ".*Think before you type.*",
+    ".*With great power comes great responsibility.*",
+    ".*Sorry, try again.*",
+    ".*sudo: 3 incorrect password attempts.*",
```
marusak, line 18: this is a one message, isn't it?

test/verify/composerlib.py
```diff
@@ -0,0 +1,72 @@
...
+        done
+        composer-cli blueprints list
+        """)
+
+        # delete all blueprints
+        self.addCleanup(self.machine.execute,
+                        "for bp in $(composer-cli blueprints list); "
+                        "do composer-cli blueprints delete $bp; done")
+
+    def check_coverage(self):
```
marusak, line 55: So I guess we can close PR you sent into Cockpit with this function?

henrywang: PR closed! Thanks!

henrywang:
@marusak, Thanks for your review! I squash into one commit because it's all about test.

henrywang:
> Thanks!
> Please don't forget to send PR to bots to remove `fedora-31/chrome` and `fedora-31/edge` from testmap.

Thanks! I'll send a PR after landing. :+1:

Issue #829, opened 2019-11-21
upload image - embedded help for AWS images


jkozol:
The following are recomended help texts for the aws image upload wizard steps. These are still WIP.

**Help text to display when the user selects AWS**

> Image Builder can upload images you create to an S3 bucket in AWS and then import them into EC2. When the image build is complete and the upload action is successful, the image file is available in the AMI section of EC2. Most of the values required to upload the image can be found in the [AWS Management Console](https://console.aws.amazon.com/console/home). Also, you must have an **Identity and Access Management (IAM)** role named vmimport.

 **Access key ID**

> You can create and find existing Access key IDs on the **Identity and Access Management (IAM)** page in the AWS console.

 **Secret access key**

> You can find the Secret access key only when you create a new Access key ID on the **Identity and Access Management (IAM)** page in the AWS console.

**AWS Region**

> Provide the.AWS Region where you want to import your image.

**Amazon S3 bucket name**

> Provide the S3 bucket name to which the image file will be uploaded before being imported into EC2. The bucket must already exist in the Region where you want to import your image. You can find a list of buckets on the **S3 buckets** page in the **Amazon S3** storage service in the AWS console.

**Image name**

> Provide a file name to be used for the image file that will be uploaded.

jgiardino:
Thanks for creating this issue. These are great! Here are a couple of suggestions...

- Add "the" to this statement in the first text chunk
  > will be removed from **the** S3 bucket that you specified.

- All instances of "portal" should be replaced with either "console" or "management console". I think saying just "AWS console" in the field level help, and stating the full name in the opening help text (see comment below) would be fine.

- What are your thoughts on adding a statement like this to the end of the first chunk of help text:
  > Most of the values required to upload the image can be found in the [AWS Management Console](https://console.aws.amazon.com/console/home).
  - There were some comments in #781 about including links, but I don't think we should provide links to specific pages, and can just provide a link to the home page of each cloud provider instead.

teg:
> an EC2 instance

Are we sure this is the term we want, rather than just "EC2"?

> and will be removed from the S3 bucket that you specified

This does not currently happen, are we sure this is the behaviour we want?

> Amazon S3 bucket name

We should document that we will not create the bucket, but that it must already exist (in the right region), and must have the right permission set (we need to research exactly what we require here).

jgiardino:
Thanks for the feedback, @teg! 

[...]

PR #946, opened 2020-04-28
Add AWS help strings and improve review step

jkozol:
Popover help text is added providing an explanation for each field. General help text about uploading to AWS is also added. The structure of the upload and review steps of the CreateImageUpload wizard
have been restructured to enable the help text. The review step is also updated to improve how the fields are organized and presented.

The help text comes from #829

### DIFF
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -110,7 +110,7 @@ class CreateImageUploadModal extends React.Component {
       imageType: "",
       imageName: "",
       imageSize: "",
-      minImageSize: null,
+      minImageSize: 0,
       maxImageSize: 2000,
       showUploadAwsStep: false,
       showReviewStep: false,
@@ -123,6 +123,7 @@ class CreateImageUploadModal extends React.Component {
     this.missingRequiredFields = this.missingRequiredFields.bind(this);
     this.setNotifications = this.setNotifications.bind(this);
     this.setImageSize = this.setImageSize.bind(this);
+    this.setImageName = this.setImageName.bind(this);
     this.setImageType = this.setImageType.bind(this);
     this.setUploadSettings = this.setUploadSettings.bind(this);
     this.handleUploadService = this.handleUploadService.bind(this);
@@ -160,14 +161,21 @@ class CreateImageUploadModal extends React.Component {
     this.props.layout.setNotifications();
   }
 
-  setUploadSettings(value, event) {
-    const key = event.target.id;
+  setUploadSettings(_, event) {
+    const key = event.target.name;
+    const value = event.target.value;
     this.setState(prevState => ({ uploadSettings: Object.assign({}, prevState.uploadSettings, { [key]: value }) }));
   }
 
   setImageSize(value) {
     this.setState({
       imageSize: value
+    });
+  }
+
+  setImageName(value) {
+    this.setState({
+      imageName: value
     });
   }
 
@@ -226,6 +234,12 @@ class CreateImageUploadModal extends React.Component {
         case "aws":
           this.setState({
             uploadService: uploadService,
+            uploadSettings: {
+              accessKeyID: "",
+              secretAccessKey: "",
+              bucket: "",
+              region: ""
+            },
             showUploadAwsStep: true,
             showReviewStep: true
           });
@@ -311,23 +325,71 @@ class CreateImageUploadModal extends React.Component {
       uploadService
     } = this.state;
 
-    const providerCheckbox = (provider, displayName) => (
-      <FormGroup
-        label={this.props.intl.formatMessage({ id: `${provider}-form`, defaultMessage: "Upload image" })}
-        fieldId="provider-checkbox"
-      >
-        <Checkbox
-          value={provider}
-          isChecked={this.state.uploadService === provider}
-          onChange={this.handleUploadService}
-          label={this.props.intl.formatMessage({
-            id: `${provider}-checkbox`,
-            defaultMessage: `Upload to ${displayName}`
-          })}
-          id={`${provider}-checkbox`}
-          aria-labelledby="provider-checkbox"
-        />
-      </FormGroup>
+    const awsProviderCheckbox = (
+      <div className="pf-c-form__group">
+        <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-u-display-flex pf-m-justify-content-flex-start pf-m-nowrap">
+          <span className="pf-l-flex__item pf-c-form__label-text">
+            <FormattedMessage defaultMessage="Upload image" />
+          </span>
+          <Popover
+            id="popover-help"
+            bodyContent={
+              <TextContent>
+                <p>
+                  <FormattedMessage
+                    defaultMessage="
+                      Image Builder can upload images you create to an {bucket} in {aws} (AWS) and then import them into EC2. When the image build is complete
+                      and the upload action is successful, the image file is available in the AMI section of EC2. Most of the values required to upload
+                      the image can be found in the {console}.
+                    "
+                    values={{
+                      bucket: "S3 bucket",
+                      aws: "Amazon Web Services",
+                      console: <a href="https://console.aws.amazon.com/console/home">AWS Management Console</a>
+                    }}
+                  />
+                </p>
+                <p>
+                  <FormattedMessage
+                    defaultMessage="
+                      This upload process requires that you have an {iam} role named {vmimport} to ensure that the image can
+                      be imported from the S3 {bucket} into EC2. For more details, refer to the {role}.
+                    "
+                    values={{
+                      bucket: "bucket",
+                      vmimport: <code>vmimport</code>,
+                      iam: "Identity and Access Management (IAM)",
+                      role: (
+                        <a href="https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html#vmimport-role">
+                          AWS Required Service Role
+                        </a>
+                      )
+                    }}
+                  />
+                </p>
+              </TextContent>
+            }
+            aria-label="Upload image help"
+          >
+            <Button variant="plain" aria-label="Upload image help">
+              <OutlinedQuestionCircleIcon id="popover-icon" />
+            </Button>
+          </Popover>
+        </div>
+        <div className="pf-c-form__horizontal-group">
+          <Checkbox
+            value="aws"
+            isChecked={this.state.uploadService === "aws"}
+            onChange={this.handleUploadService}
+            label={this.props.intl.formatMessage({
+              id: `aws-checkbox`,
+              defaultMessage: `Upload to AWS`
+            })}
+            id="aws-checkbox"
+            aria-labelledby="provider-checkbox"
+          />
+        </div>
+      </div>
     );
 
     const imageStep = {
@@ -350,11 +412,9 @@ class CreateImageUploadModal extends React.Component {
               title={formatMessage(messages.warningEmptyBlueprint)}
             />
           )}
-          <Form isHorizontal>
-            <div className="pf-c-form__group">
-              <label className="pf-c-form__label" htmlFor="blueprint-name">
-                <span className="pf-c-form__label-text">Blueprint</span>
-              </label>
+          <Form isHorizontal className="cc-m-wide-label">
+            <div className="pf-c-form__group cc-c-form__horizontal-align">
+              <span className="pf-c-form__label-text">Blueprint</span>
               <div className="pf-c-form__horizontal-group cc-c-popover__horizontal-group">
                 <Text id="blueprint-name">{blueprint.name}</Text>
                 <Popover
@@ -376,7 +436,7 @@ class CreateImageUploadModal extends React.Component {
                 ))}
               </FormSelect>
             </FormGroup>
-            {imageType === "ami" && providerCheckbox("aws", "AWS")}
+            {imageType === "ami" && awsProviderCheckbox}
             <div className="pf-c-form__group">
               <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-m-justify-content-flex-start">
                 <label htmlFor="create-image-size" className="pf-l-flex__item">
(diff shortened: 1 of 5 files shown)


jgiardino:
Thanks for working on this!! I reviewed this implementation primarily focused on the alignment issues that you raised. These are all the things I noticed, but some of these comments might not be in the scope of this PR.

- I customized the length of form labels to be longer in the prototype to avoid wrapping in some cases. This is the css I used for that:
  ```
  .pf-c-form.pf-m-horizontal .pf-c-form__group {
    grid-template-columns: 200px 1fr;
  }
  ```

- `.cc-c-popover__horizontal-group` should have `align-item: baseline` to fix the alignment of the blueprint name with the label. Just note that this is a temporary fix, since this only works if the help button is present, and we have another issue to pull this button out.  And a side note, “Blueprint” isn’t really a label, so using the `<label for=“”>` in this case adds no value. But it also doesn’t hurt anything either :shrug:

- When looking at the Upload image input, these are some things I notice. Some of these are specific to this input, but some suggestions might address issues with other inputs:
  - aria-label for the help button could be “Upload image help”. This would work regardless of image type.
  - A checkbox input cannot have more than one label. In this case, the label is “Upload to AWS”. Since this is the only checkbox we have, you can make the “Upload image” element a `<span>`. But if we add more checkboxes, we would need to use this label as a group label. The prototype provides an example of this.
  - The classes you’re using on `pf-c-form__label` are correct. But the styles for `.pf-c-form__label` are overriding the styles for `.pf-l-flex`. It isn’t obvious to me how to fix this. An easy fix might be to pull in utility styles and add `.pf-u-display-flex` to force this element to have `display: flex`. The utility style import should be `import "@patternfly/react-styles/css/utilities/Display/display.css”;`

- There is an issue that occurs when the Esc key is pressed while help text is visible. If a popover is visible, then Esc should only close the popover. Currently, it closes the entire modal. Is this something you can address, or does it require an issue in PF-react?

- Regarding actual help text contents, the Access key ID appears to have the wrong help text.

jkozol:
@jgiardino Thanks for the review! Everything should be fixed now except for:

> I customized the length of form labels to be longer in the prototype to avoid wrapping in some cases 
>```
>.pf-c-form.pf-m-horizontal .pf-c-form__group {
>    grid-template-columns: 200px 1fr;
>}
>```

I am still unable to get this to work in cockpit-composer.

public/custom.css
```diff
@@ -851,4 +855,17 @@ body {
...
+/* Setting a fixed width similar to the values used in horizontal forms */
+/* var(--pf-c-form--m-horizontal--md__group--GridTemplateColumns)  */
+/* NOTE that 576px is defined as a sass variable in PF source code as min-width: $pf-global--breakpoint--sm */
+@media screen and (min-width: 576px) {
+  .pf-c-content dl.cc-m-column__fixed-width {
+    grid-template-columns: 200px 1fr;
+  }
+}
+
+.pf-c-form.pf-m-horizontal .pf-c-form__group {
```
jgiardino, line 869: Change this line to:
`.pf-c-form.pf-m-horizontal.cc-m-wide-label .pf-c-form__group {`

By doing this, we are creating a modifier class for the PF form component to create a wider field label.

And then this class `.cc-m-wide-label` would need to be applied to the element with class `pf-c-form`. 

This is a different approach that what I had originally suggested. But it's better than my previous suggestions. The modifier class lets us control which form has a wider label. And placing the modifier class on the form wrapper, rather than each input/label wrapper means we only need to define it once.

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +333,72 @@ class CreateImageUploadModal extends React.Component {
...
-            defaultMessage: `Upload to ${displayName}`
-          })}
-          id={`${provider}-checkbox`}
-          aria-labelledby="provider-checkbox"
-        />
-      </FormGroup>
+    const awsProviderCheckbox = (
+      <div className="pf-c-form__group">
+        <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-u-display-flex pf-m-justify-content-flex-start pf-m-nowrap">
+          <label htmlFor="aws-checkbox" className="pf-l-flex__item pf-c-form__label-text">
```
jgiardino, line 339: This element should be a `<span>` without the `htmlFor` but keeping the same className values. The checkbox input already has a label, "Upload to AWS"

jgiardino:
The updates look good! I provided a couple of inline comments for some easy fixes. The first is regarding your previous comment regarding css for the field label length.

public/custom.css
jkozol: Thanks. I updated this.

components/Modal/CreateImageUpload.js
jkozol: I fixed this

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                <label htmlFor="image-name-input" className="pf-l-flex__item">
+                  <span className="pf-c-form__label-text">
+                    <FormattedMessage defaultMessage="Image name" />
+                  </span>
+                  <span className="pf-c-form__label-required" aria-hidden="true">
+                    &#42;
+                  </span>
+                </label>
+                <Popover
+                  id="popover-help"
```
henrywang, line 603: Could you please give `id` a unique name, like `image-name popover`? That'll be very helpful to find this element when write test code. Thanks!

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                <label htmlFor="secret-access-key-input" className="pf-l-flex__item">
+                  <span className="pf-c-form__label-text">
+                    <FormattedMessage defaultMessage="Secret access key" />
+                  </span>
+                  <span className="pf-c-form__label-required" aria-hidden="true">
+                    &#42;
+                  </span>
+                </label>
+                <Popover
+                  id="popover-help"
```
henrywang, line 554: Same thing here. Thanks!

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                <label htmlFor="access-key-id-input" className="pf-l-flex__item">
+                  <span className="pf-c-form__label-text">
+                    <FormattedMessage defaultMessage="Access key ID" />
+                  </span>
+                  <span className="pf-c-form__label-required" aria-hidden="true">
+                    &#42;
+                  </span>
+                </label>
+                <Popover
+                  id="popover-help"
```
henrywang, line 518: Another one. Thanks.

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +320,70 @@ class CreateImageUploadModal extends React.Component {
...
-        />
-      </FormGroup>
+    const awsProviderCheckbox = (
+      <div className="pf-c-form__group">
+        <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-u-display-flex pf-m-justify-content-flex-start pf-m-nowrap">
+          <span className="pf-l-flex__item pf-c-form__label-text">
+            <FormattedMessage defaultMessage="Upload image" />
+          </span>
+          <Popover
+            id="popover-help"
```
henrywang, line 330: Another one.

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+            <div className="pf-c-form__group">
+              <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-u-display-flex pf-m-justify-content-flex-start pf-m-nowrap">
+                <label htmlFor="bucket-input" className="pf-l-flex__item">
+                  <span className="pf-c-form__label-text">Amazon S3 bucket</span>
+                  <span className="pf-c-form__label-required" aria-hidden="true">
+                    &#42;
+                  </span>
+                </label>
+                <Popover
+                  id="bucket-popover"
```
henrywang, line 633: This one is super awesome! :)

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+
+    const awsReviewStep = uploadService === "aws" && (
+      <TextContent>
+        <div className="pf-l-flex pf-u-display-flex">
+          <h3 className="pf-l-flex__item pf-u-mt-2xl pf-u-mb-md">
+            <FormattedMessage defaultMessage="Upload to Amazon" />
+          </h3>
+          <Popover
+            className="pf-l-flex__item"
+            id="popover-help"
```
henrywang, line 721: Mark here.

jkozol:
@henrywang I updated the popover ids

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +320,70 @@ class CreateImageUploadModal extends React.Component {
...
+            <FormattedMessage defaultMessage="Upload image" />
+          </span>
+          <Popover
+            id="popover-help"
+            bodyContent={
+              <TextContent>
+                <p>
+                  <FormattedMessage
+                    defaultMessage="
+                      Image Builder can upload images you create to an {bucket} in AWS and then import them into EC2. When the image build is complete
```
jgiardino, line 336: I didn't catch this in my first review. I'm wondering if the first time we mention AWS, we should use this:
> Amazon Web Services (AWS)

And then after that, just use "AWS".

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                  </span>
+                  <span className="pf-c-form__label-required" aria-hidden="true">
+                    &#42;
+                  </span>
+                </label>
+                <Popover
+                  id="region-popover"
+                  bodyContent={
+                    <FormattedMessage
+                      defaultMessage="Provide the AWS Region where you want to import your image. This must be the same region where the {bucket} exists."
```
jgiardino, line 681: In the original upload PR, you provided this link to AWS documentation that lists the regions: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html

We also discussed making this a select menu, which is out of scope of this PR. I'm also not sure how often region names are updated, which creates the risk that the region the user wants is not included as a select menu option. And we could always have an option to manually enter a region in that case. Which makes the control even larger in scope.

For now, I'm thinking we should at least provide the link above, and clarify that we want the "region" and not the "region name". For this latter point, we could use placeholder text "E.g. eu-central-1". And for the link, maybe this additional line in the popover would work:
> You can find a list of regions in the [AWS User Guide](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html).

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +320,70 @@ class CreateImageUploadModal extends React.Component {
...
+          <Checkbox
+            value="aws"
+            isChecked={this.state.uploadService === "aws"}
+            onChange={this.handleUploadService}
+            label={this.props.intl.formatMessage({
+              id: `aws-checkbox`,
+              defaultMessage: `Upload to AWS`
+            })}
+            id="aws-checkbox"
+            aria-labelledby="provider-checkbox"
```
jgiardino, line 383: `aria-labelledby` isn't needed here, and is potentially bad... The checkbox has a visible label that is already accessible to assistive technology.

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
-                    id={key}
-                    key={key}
-                    onChange={this.setUploadSettings}
-                    isRequired
-                  />
-                </FormGroup>
-              ))}
-            </Form>
-          )}
+                  <Button variant="plain" aria-label="access key id help">
```
jgiardino, line 529: Just to be save, can we make acronyms upper-case. I _think_ most assistive technology is pretty decent at identifying acronyms when all caps are used.

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                          buckets: "buckets",
+                          bucketsPage: <strong>S3 buckets</strong>,
+                          amazon: "Amazon"
+                        }}
+                      />
+                    </React.Fragment>
+                  }
+                  aria-label="bucket help"
+                >
+                  <Button variant="plain" aria-label="bucket help">
```
jgiardino, line 653: I usually recommend that the help text button follow this format: "[field label] help". In this case, I can see the case for shortening it when input labels are really long, but am wondering if we should at least include "S3" in front of "bucket help".

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                    <FormattedMessage
+                      defaultMessage="Provide the AWS Region where you want to import your image. This must be the same region where the {bucket} exists."
+                      values={{
+                        bucket: "S3 bucket"
+                      }}
+                    />
+                  }
+                  aria-label="region help"
+                >
+                  <Button variant="plain" aria-label="region help">
```
jgiardino, line 689: Similar to my previous comment, I would include "AWS" here: "AWS region help".

components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +496,297 @@ class CreateImageUploadModal extends React.Component {
...
+                        </a>
+                      )
+                    }}
+                  />
+                </p>
+              </TextContent>
+            }
+            aria-label="aws help"
+          >
+            <Button variant="plain" aria-label="aws help">
```
jgiardino, line 759: Similar to a previous comment, acronyms used in `aria-label` should be all caps.

jgiardino:
I provided a few comments about some of the `aria-label` attributes. In doing so, I realized that these are text strings, and therefore should be translated. The only way I'm aware of including translated strings in attributes is by defining at the beginning of the component instead of inline. For these, I'm also not sure if it's better to have separate strings for "[field label]" and "[field label] help" or to have just "help" defined once, and concatenated with "[field label]". I think the former might be more ideal for the translation, but I'm also not sure if there's much benefit compared to the cost of that variation.

jkozol:
@jgiardino All of the aria labels should be fixed now and the help text has been updated. The only things missing are translating the aria-labels and fixing the close wizard issue when a popover is open.

PR #947, opened 2020-04-29
test: Change image build type to openstack from tar

henrywang:
Because tar does not support image size setting, use openstack instead

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -101,7 +101,7 @@ class TestImage(composerlib.ComposerCase):
         # collect code coverage result
         self.check_coverage()
 
-    def testTar(self):
+    def testOpenStack(self):
         b = self.browser
         m = self.machine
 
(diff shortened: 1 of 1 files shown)


henrywang:
@marusak, @martinpitt, CDP is more stable and faster. Awesome! I love it. :1st_place_medal:

martinpitt:
@henrywang : I'm glad you like it! :+1:

henrywang:
> Does it also make sense to test creation of `tar` without setting of the size? Or is this the same from the point of c-composer? Or shouldn't there be then test, that it behaves expectedly when you set up size for tar? (warning that it is not possible, the size input is disabled/not shown...)

The image size does not make sense for `tar` image, and @jkozol filed an [issue](https://github.com/osbuild/cockpit-composer/issues/945) to remove image size setting in image creating wizard when `tar` has been selected.

PR #953, opened 2020-05-04
Fix missing image size in action/saga call

jkozol:
In the CreateImageUpload wizard a user can specify an image size. The image size was missing from the action dispatcher. The image size is now included when calling the start compose action/saga.


jkozol:
@henrywang it wasn't causing an issue in the console because it wouldn't cause an error. It would just cause the image size to be default.

PR #954, opened 2020-05-04
Add Azure Upload Support

jkozol:
Add ability to upload vhd images to Microsoft Azure in the CreateImageUpload wizard.

This is dependent on #946 and #953

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,10 @@
     "npm": ">=5.0"
   },
   "dependencies": {
-    "@patternfly/patternfly": "^2.31.0",
+    "@patternfly/patternfly": "2.31.0",
     "@patternfly/react-core": "3.153.13",
     "@patternfly/react-icons": "3.15.16",
-    "@patternfly/react-styles": "^3.7.13",
+    "@patternfly/react-styles": "3.7.13",
     "bootstrap": "3.4.1",
     "core-js": "3.6.4",
     "core-js-compat": "3.6.5",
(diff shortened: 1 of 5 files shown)


jgiardino:
Thanks for adding this! I'm still trying to get an image to successfully build. But here are some things I noticed. 

1. There is a difference in capitalization of "Azure" in the following example. Can we the Upload type also have an upper-case A for Azure?
    image
2. The first step includes a select menu for "Type" with options having both a descriptive label of the type along with a file format in parentheses. When "Type" is repeated on the review step, it displays the file format, and not the descriptive label. I would actually expect the opposite in this case. Would this be a separate issue (since the same is true for Amazon image types)?

jgiardino:
I was able to successfully build and upload in RHEL and F31. The F32 machine would not build an Azure image. :-(

jkozol:
@jgiardino Thanks for the review. I can create issues for your feedback and create separate PR's for them.

package.json
```diff
@@ -10,6 +10,7 @@
     "@patternfly/patternfly": "^2.31.0",
     "@patternfly/react-core": "3.153.13",
     "@patternfly/react-icons": "3.15.16",
+    "@patternfly/react-styles": "^3.7.13",
```
henrywang, line 13: @jkozol, could you please remove `^`? The cockpit-ci has `bots` to upgrade dependencies automatically, but it'll not update package with version start with `^`. Thanks!

jkozol: Thanks! I fixed this.

PR #956, opened 2020-05-05
Azure help strings

jkozol:
Add popover help strings for azure upload. This is dependent on #954

### DIFF
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -20,7 +20,12 @@ import {
   WizardContextConsumer,
   WizardFooter
 } from "@patternfly/react-core";
-import { OutlinedQuestionCircleIcon, ExclamationTriangleIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
+import {
+  OutlinedQuestionCircleIcon,
+  ExclamationTriangleIcon,
+  ExclamationCircleIcon,
+  ExternalLinkSquareAltIcon
+} from "@patternfly/react-icons";
 import { defineMessages, FormattedMessage, injectIntl, intlShape } from "react-intl";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
@@ -69,6 +74,9 @@ const messages = defineMessages({
   },
   warningSizeEmpty: {
     defaultMessage: "A value is required."
+  },
+  uploadAWS: {
+    defaultMessage: "Upload to AWS"
   }
 });
 
@@ -395,7 +403,20 @@ class CreateImageUploadModal extends React.Component {
                     "
                     values={{
                       bucket: "S3 bucket",
-                      console: <a href="https://console.aws.amazon.com/console/home">AWS Management Console</a>
+                      console: (
+                        <Button
+                          component="a"
+                          className="pf-icon"
+                          target="_blank"
+                          variant="link"
+                          icon={<ExternalLinkSquareAltIcon />}
+                          iconPosition="right"
+                          isInline
+                          href="https://console.aws.amazon.com/console/home"
+                        >
+                          AWS Management Console
+                        </Button>
+                      )
                     }}
                   />
                 </p>
@@ -410,9 +431,18 @@ class CreateImageUploadModal extends React.Component {
                       vmimport: <code>vmimport</code>,
                       iam: "Identity and Access Management (IAM)",
                       role: (
-                        <a href="https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html#vmimport-role">
+                        <Button
+                          component="a"
+                          className="pf-icon"
+                          target="_blank"
+                          variant="link"
+                          icon={<ExternalLinkSquareAltIcon />}
+                          iconPosition="right"
+                          isInline
+                          href="https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html#vmimport-role"
+                        >
                           AWS Required Service Role
-                        </a>
+                        </Button>
                       )
                     }}
                   />
@@ -431,10 +461,7 @@ class CreateImageUploadModal extends React.Component {
             value="aws"
             isChecked={this.state.uploadService === "aws"}
             onChange={this.handleUploadService}
-            label={this.props.intl.formatMessage({
-              id: `aws-checkbox`,
-              defaultMessage: `Upload to AWS`
-            })}
+            label={formatMessage(messages.uploadAWS)}
             id="aws-checkbox"
           />
         </div>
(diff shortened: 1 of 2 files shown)


components/Modal/CreateImageUpload.js
```diff
@@ -442,88 +555,515 @@ class CreateImageUploadModal extends React.Component {
...
+                </label>
+                <Popover
+                  id="bucket-popover"
+                  bodyContent={
+                    <TextContent>
+                      <FormattedMessage defaultMessage="Provide the Blob container to which the image file will be uploaded. You can find containers on the" />
+                      <strong>
+                        <FormattedMessage defaultMessage=" Blobs " />
+                      </strong>
+                      <FormattedMessage defaultMessage="page of a storage account. You can find storage accounts on the" />
```
jgiardino, line 991: Would this statement make sense? I'm not seeing a specific "Blob" page. Open to suggestions on this one.
> You can find containers under the **Blob service** section of a storage account.

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +337,111 @@ class CreateImageUploadModal extends React.Component {
...
+                <p>
+                  <FormattedMessage
+                    defaultMessage="
+                      Image Builder can upload images you create to an {bucket} in AWS and then import them into EC2. When the image build is complete
+                      and the upload action is successful, the image file is available in the AMI section of EC2. Most of the values required to upload
+                      the image can be found in the {console}.
+                    "
+                    values={{
+                      bucket: "S3 bucket",
+                      console: <a href="https://console.aws.amazon.com/console/home">AWS Management Console</a>
```
jgiardino, line 359: When I click this link, nothing happens. However if I right-click and select to open the link in a new tab, it does work. I would expect in this context that external links to automatically open in a new tab. 

When including external links, we should include the icon `<ExternalLinkSquareAltIcon />` at the end of the text. The [PF react button component](https://www.patternfly.org/v4/documentation/react/components/button#variations) includes an example of how this looks, but if you were going to use the react component, I think we would also need the `isInline` prop so that the link displays proper spacing when inline with text.

components/Modal/CreateImageUpload.js
```diff
@@ -311,23 +337,111 @@ class CreateImageUploadModal extends React.Component {
...
+          <Popover
+            id="popover-help"
+            bodyContent={
+              <FormattedMessage
+                defaultMessage="
+                    Image Builder can upload images you create to a Blob container in Microsoft {azure}. When the image build is complete 
+                    and the upload action is successful, the image file is available in the Storage account and Blob container that you specified.
+                    "
+                values={{
+                  azure: "Azure"
```
jgiardino, line 421: Can we associate a url with "Microsoft Azure"? And which one would make more sense? https://azure.microsoft.com/ or https://portal.azure.com/ ?

Also see my previous comment about the `<ExternalLinkSquareAltIcon />` icon.

components/Modal/CreateImageUpload.js
jkozol: Yes this makes sense.

components/Modal/CreateImageUpload.js
jkozol: I decided to go with https://portal.azure.com/ since we send users to the aws console.

jkozol: I decided to go with https://portal.azure.com/ since we send users to the aws console.

components/Modal/CreateImageUpload.js
jkozol: This icon has been added. Is this the desired alignment?

linkIconAlignment

jgiardino: I see the same alignment when I build locally. When I compare the runtime html to the example on [PF.org](https://www.patternfly.org/v4/documentation/react/components/button), I see that the `<svg>` in the PF example includes `vertical-align` which results in a lower position:
`<svg style="vertical-align: -0.125em;" fill="currentColor" height="1em" width="1em" viewBox="0 0 448 512" aria-hidden="true" role="img">`

But it isn't obvious to me why our `<svg>` is missing that `vertical-align` property. As a short term fix, we could add the same styles to our css file, for example:
```
.pf-icon svg {
  vertical-align: -0.125em;
}
```
and apply `.pf-icon` to the this link.

Also, I noticed that this is a `<button>` element, but it should be an `<a>` element since it has an `href` attribute.

components/Modal/CreateImageUpload.js
```diff
@@ -447,16 +487,42 @@ class CreateImageUploadModal extends React.Component {
...
+                  azure: (
+                    <Button
+                      variant="link"
+                      icon={<ExternalLinkSquareAltIcon />}
+                      iconPosition="right"
+                      isInline
+                      href="https://portal.azure.com/"
+                    >
+                      Microsoft Azure
+                    </Button>
```
jgiardino, line 508: Per my previous comment, this component should render as an `<a>` html element. If you add the following prop here, that should do it:
`component="a"`

Also per my previous comment, this component is also where you can add `className="pf-icon"` to apply the vertical-align styles to the svg that it includes.

jkozol: I have added your suggested pf-icon css and I updated the buttons to render as an `<a>` element by using `component="a"`

components/Modal/CreateImageUpload.js
```diff
@@ -942,18 +1090,41 @@ class CreateImageUploadModal extends React.Component {
...
+                      <FormattedMessage defaultMessage="Provide the Blob container to which the image file will be uploaded. You can find containers under the" />
+                      <strong>
+                        <FormattedMessage defaultMessage=" Blob service " />
+                      </strong>
+                      <FormattedMessage defaultMessage="section of a storage account. You can find storage accounts on the" />
+                      <strong>
+                        <FormattedMessage defaultMessage=" Storage accounts " />
+                      </strong>
+                      <FormattedMessage defaultMessage="page in the {azure} portal." values={{ azure: "Azure" }} />
+                    </TextContent>
```
jgiardino, line 1114: When I review the .pot file for this string, the strings are broken up. For example, one of the strings is presented to translators like this:
```
#: build/messages/components/Modal/CreateImageUpload.json
#. [components.Modal.2862073206]
#. defaultMessage is:
#. Provide the Blob container to which the image file will be uploaded. You can find containers under the
msgid "Provide the Blob container to which the image file will be uploaded. You can find containers under the"
msgstr ""
```

When strings are provided to translators, they should be complete phrases or sentences. The best solution I can think of for handling this is how it's handled [here](https://github.com/osbuild/cockpit-composer/blob/master/components/ListView/SourcesListItem.js#L52).

There are a couple of other strings in this PR that require similar fixes.

PR #969, opened 2020-05-25
compose: add ostree commit image types

teg:
These are a bit different than our other image types, but this is the
closest fit I could find without majorly reworking the API/UI.

The image types are not yet part of osbuild-composer, but I think we can
merge this eagerly as they will be ignored in case the backend doesn't
support them.

Two new image types are introduced "Fedora IoT" and "RHEL for Edge",
these are going to be very similar, but obviously available for
Fedora/RHEL respectively.

Signed-off-by: Tom Gundersen <teg@jklm.no>

### DIFF
--- a/core/composer.js
+++ b/core/composer.js
@@ -133,10 +133,12 @@ export function getComposeTypes() {
   const imageTypeLabels = {
     ami: "Amazon Machine Image Disk (.ami)",
     "ext4-filesystem": "Ext4 File System Image (.img)",
+    "fedora-iot-commit": "Fedora IoT Commit (.tar)",
     "live-iso": "Live Bootable ISO (.iso)",
     "partitioned-disk": "Raw Partitioned Disk Image (.img)",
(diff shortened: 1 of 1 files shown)


henrywang:
@teg, I have an question about this change. The rhel one will be available only if it's on rhel, right? Otherwise that'll cause issue if user build rhel iot image on fedora.

teg:
> @teg, I have an question about this change. The rhel one will be available only if it's on rhel, right? 

Correct, osbuild-composer will only expose the Fedora variant on a Fedora machine and the RHEL variant on a RHEL machine. So as long as cockpit-composer only shows the image types osbuild-composer tells it about, we should all be good :)

henrywang:
@teg, make sense to me. Thanks!
I tried three times, the only failed case is due to image building timeout(30 minutes). I think most of time spend on package downloading. I'll retry the failed case tomorrow morning (my timezone), and all traffic in/out will become `silence`.

teg:
Thanks!

PR #972, opened 2020-05-27
Fix translations

jkozol:
Fix translations for blueprint components and filtering.

### DIFF
--- a/components/Toolbar/FilterInput.js
+++ b/components/Toolbar/FilterInput.js
@@ -1,6 +1,28 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Filter } from "patternfly-react";
+import { defineMessages, injectIntl, intlShape } from "react-intl";
+
+const messages = defineMessages({
+  filterNameTitle: {
+    defaultMessage: "Name"
+  },
+  filterNamePlaceholder: {
+    defaultMessage: "Filter by Name"
+  },
+  filterReleaseTitle: {
+    defaultMessage: "Release"
+  },
+  filterReleasePlaceholder: {
+    defaultMessage: "Filter by Release"
+  },
+  filterVersionTitle: {
+    defaultMessage: "Version"
+  },
+  filterVersionPlaceholder: {
+    defaultMessage: "Filter by Version"
+  }
+});
 
 class FilterInput extends React.Component {
   constructor() {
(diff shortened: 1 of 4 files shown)


components/Toolbar/FilterInput.js
```diff
@@ -49,7 +52,10 @@ class FilterInput extends React.Component {
               {currentFilterType !== undefined && filters.filterTypes.length > 1 && (
                 <Filter.TypeSelector
                   filterTypes={filters.filterTypes}
-                  currentFilterType={currentFilterType}
+                  currentFilterType={formatMessage({
+                    id: "current filter type",
+                    defaultMessage: currentFilterType.title
```
jgiardino, line 57: Is this possible? I thought that `defaultMessage` required explicitly defined text strings. 
Looking at our [documentation](https://github.com/osbuild/cockpit-composer#i18n) for this, I'm not seeing a command I can run to test that this does or doesn't work.

The documentation also states that an `id` isn't needed, but I'm also not sure if this would break anything.

jgiardino: @marusak Do you know what command I can run, and where I can look to see if the strings in this file are getting pulled correctly?

jgiardino: Looking at this some more, there are a couple of things I noticed that might be issues with how these strings are defined.

Anywhere we use `formatMessage`, as in `aria-label={formatMessage(messages.ariaLabel)}`, the text string is defined at the beginning of the same file in something like the following:
```
    details: {
      defaultMessage: "details"
    }
  });
```
But in this file, the `defineMessages()` is not there, and the text strings themselves are also not actually in this file. Instead, they are defined in [store.js](https://github.com/osbuild/cockpit-composer/blob/master/core/store.js#L123).

To test whether having strings defined in store.js is a problem, I updated the file to include string definitions using `defineMessages({ ... })` at the beginning of the file. After doing that, the following error was returned when trying to build:
`[React Intl] Messages must be statically evaluate-able for extraction.`

So I think this means that the strings—`filterType.title` and `filterType.placeholder`—should be moved from store.js to FilterInput.js. 

Also, regarding my previous question with testing strings, anytime you run a command like `npm run build`, it appears to extract the strings and dump them into .json files under build/messages. When I look in these files, I'm not seeing any strings where the following pattern is used. I see these strings display in the runtime DOM, but I think for strings to be extracted for translation, `defineMessages()` is needed when using {formatMessage(...)}.
```
{formatMessage({
  id: "filter key",
  defaultMessage: item.key
})}
```

components/Toolbar/ToolbarLayout.js
```diff
@@ -15,18 +15,26 @@ class ToolbarLayout extends React.Component {
...
+                <FormattedMessage defaultMessage="Active Filters" />:
+              </Filter.ActiveLabel>
               <Filter.List>
                 {filters.filterValues.map(item => {
                   return (
                     <Filter.Item key={`${item.key}-${item.value}`} onRemove={this.handleRemoveFilter} filterData={item}>
-                      {item.key}: {item.value}
+                      {formatMessage({
+                        id: "filter key",
+                        defaultMessage: item.key
```
jgiardino, line 35: My comments [above](https://github.com/osbuild/cockpit-composer/pull/972/files#r432127303) also apply to this case.

components/Toolbar/FilterInput.js
marusak: @jgiardino I think you mean to see how it would be generated for translators, right?

If so, then you can just run `make po/cockpit-composer.pot` and then look into `po/cockpit-composer.pot`.

jgiardino: Thanks, @marusak! I have one more question for you :-)
When do you usually run steps 2 and 4 that are listed [here](https://github.com/osbuild/cockpit-composer#i18n)? Or, when do you recommend that we run them?

marusak: That doc is out of date.
`translations:push` and `translations:pull` don't exist anymore. This [doc](https://github.com/cockpit-project/cockpit/blob/master/doc/i18n.md#translation-process) describes it correctly. I'll send PR to fix it.

And you don't want to run this manually normally, it is automated and every 7 days bots do that for us, see PRs like [this](https://github.com/osbuild/cockpit-composer/pull/971) one for example.

jgiardino: 👍 Thanks!!

components/Toolbar/ToolbarLayout.js
jkozol: Fixed

components/Toolbar/ToolbarLayout.js
```diff
@@ -1,8 +1,20 @@
...
 
+const messages = defineMessages({
+  filterNameID: {
+    defaultMessage: "name"
+  },
+  filterReleaseID: {
+    defaultMessage: "release"
+  },
+  filterVersionID: {
+    defaultMessage: "version"
```
jgiardino, line 14: One minor nit here regarding capitalization. Could we use upper-case for the beginning of each, i.e. "Name" instead of "name"?

jkozol: In master these are displayed as lower case. This is updated now.

jgiardino: Thanks!!

PR #982, opened 2020-06-03
test: Add more complicated password in test

henrywang:


henrywang:
Thanks @jkozol. All green! Let's land it.

PR #984, opened 2020-06-03
Add messaging about empty blueprints

### LINKED ISSUE #973: Unsure if I must add packages to Empty Blueprint: "This blueprint is empty"
### Description of problem

I created a new blueprint, and gave it a name. I didn't add anything to the blueprint, and clicked "Create Image". It gave me an error "This blueprint is empty"

I then cancelled my image creation, and tried to figure out what to add, like the kernel, systemd, bash and so on ... to make a bootable system.

It is unclear to me if this is my responsibility as a user of cockpit-composer, or whether the default set of packages is handled for me. I became confused when I saw "This blueprint is empty" 

@ondrejbudai Helped explain things to me on IRC, and I now understand that I don't need to add any packages to create a default image.

### Component (web, API, etc)

cockpit-composer

cockpit-composer-18-1.fc32.noarch

### How often reproducible

Every time.

### Steps to Reproduce

1. Create new blueprint
2. Click 'Create Image'

### Actual results

Screenshot from 2020-05-29 11-42-21

### Expected results

Some indication that an empty blueprint is acceptable. Whether in the main screen or in the Create Image dialog. Or at a minimum, suggest not displaying an error. This is confusing to me.

jgiardino:
Includes messaging to communicate that empty blueprints result in minimal images, and packages needed for selected image types are automatically included.

fixes #973

Added the second statement in the empty state message below:
Screen Shot 2020-06-03 at 4 43 02 PM

Changed the `warning` to an `info` for the Alert, and included an additional description about the minimal image.
Screen Shot 2020-06-03 at 4 43 20 PM

### DIFF
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -47,6 +47,9 @@ const messages = defineMessages({
     id: "empty-blueprint-alert",
     defaultMessage: "This blueprint is empty."
   },
+  warningEmptyBlueprintDesc: {
+    defaultMessage: "A minimal image will be created with only the packages needed to support the selected image type."
+  },
   warningReview: {
     defaultMessage: "There are one or more fields that require your attention."
   },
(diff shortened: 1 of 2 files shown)


jgiardino:
@andreasn Would you take a look at the text strings, and let me know if you have any suggestions?

jgiardino:
After review with others, I modifed the second statement based on suggestions. I did not however split the last statement off into a separate paragraph, since that would require refactoring the component :-/ but I did go ahead and split it as a separate text string to support that future work (without requiring new translations for essentially the same text). I also modified the last statement only slightly to be more specific about when the packages are included.

image

PR #986, opened 2020-06-03
Password verification

### LINKED ISSUE #985: Remove password verification
### Description of problem
A [test in cockpit-composer](https://github.com/osbuild/cockpit-composer/blob/6e735255bb114be4ee5687d354aaec6fe3d67bd1/test/verify/check-custom#L173) regarding password verification is currently failing in Fedora but passing in RHEL. But considering the use cases that cockpit-composer supports, requiring an acceptable password based on the host machine isn't a valid use case. The proposed solution is to drop the password verification, to avoid blocking the user from defining the password that they want to use.

### Steps to Reproduce

1. Navigate to the Blueprint page.
2. Click the "Create User Account" button on the Customizations tab.
3. Specify a weak password, like "foobar", in both password fields.

### Actual results
A message displays, saying "Password quality check failed: The password is shorter than 8 characters" and you are prevented from creating the user account.

### Expected results
The user should not be blocked when specifying a weak password. Additionally, the password verification is checking the password policy on the host machine, which may or may not be the password policy of the target machine for the image you're building. Therefore, I'd propose we remove the password verification altogether, and just check that the passwords match.

cc @jkozol @henrywang

jgiardino:
This removes the password quality check from the Create User Account modal.

_I forgot to include in the original commit:_
fixes #985
🤦

### DIFF
--- a/components/Form/Password.js
+++ b/components/Form/Password.js
@@ -39,12 +39,17 @@ class Password extends React.Component {
     }
   }
 
-  handleWarnings() {
+  // warnings display:
+  // if password two has a value and the values for one and two do not match, OR
+  // if password two does NOT have a value but one does, and the user is exiting password two
+  // warnings do not display: 
+  // if password one has a value and two does not, and the user is exiting password one,
+  // because we assume the user is intending to provide a password in this case
+  handleWarnings(input) {
     setTimeout(() => {
       if (
-        this.state.passwordOne !== this.state.passwordTwo &&
-        this.state.passwordOne !== "" &&
-        this.state.passwordTwo !== ""
+        (this.state.passwordOne !== this.state.passwordTwo && this.state.passwordTwo !== "") || 
+        (input === "two" && this.state.passwordOne !== "" && this.state.passwordTwo === "")
       ) {
         this.setState({ displayWarningMatch: true });
       } else {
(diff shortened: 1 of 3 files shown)


jgiardino:
@henrywang @jkozol Do we also want to handle the password verification test as part of this PR? Please let me know if you have any suggestions on updates for tests that are needed.

components/Form/Password.js
```diff
@@ -1,56 +1,28 @@
...
-      // if both values match, and will pass the password to the parent component
-    } else {
-      this.setState({
-        passwordQuality: "",
-        warningQuality: ""
-      });
-    }
     if (event.target.value === this.state.passwordTwo) {
       this.setState({ displayWarningMatch: false });
       if (event.target.value === "") {
```
jkozol, line 28: The Create button is not disabled when only the first password is entered. Adding the else statement will invalidate the password unless both passwords are empty.

createButtonEnabled

jkozol:
```suggestion
      if (event.target.value === "") {
        this.props.setValidPassword(event.target.value);
      }
    } else {
      this.props.setValidPassword();
    }
```

components/Form/Password.js
```diff
@@ -64,21 +36,15 @@ class Password extends React.Component {
     if (this.state.passwordOne === event.target.value) {
       this.setState({ displayWarningMatch: false });
     }
-    // if the passwords match, and there's no quality warning message defined
-    // then pass the password to the parent component, otherwise, the password values
-    // are invalid and the submit button will be disabled
-    if (this.state.passwordOne === event.target.value && this.state.warningQuality === "") {
+    if (this.state.passwordOne === event.target.value) {
       this.props.setValidPassword(event.target.value);
```
jkozol, line 40: These if statements can be combined.

jkozol:
@jgiardino Can you review the changes I added per my comments?

PR #988, opened 2020-06-04
test: Fix preferences for new Firefox

martinpitt:
Cockpit's CI moved from an old Firefox 69 to the current version and
nightly builds. This requires some additional user preferences which are
fixed in cockpit master, but not in a release yet.

So check out the corresponding SHA from cockpit's test/common. Once
Cockpit 221 gets released, it can go back to cloning the tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,9 +139,12 @@ machine: bots
 	rsync -avR --exclude="bots/machine/machine_core/__pycache__/" bots/machine/testvm.py bots/machine/identity bots/machine/cloud-init.iso bots/machine/machine_core bots/task/testmap.py test
 
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
+# this needs a recent adjustment for firefox 77, so checkout a SHA until cockpit 221 releases
 test/common:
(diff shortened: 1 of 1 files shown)


martinpitt:
With current master, tests [fail like this](https://logs.cockpit-project.org/logs/pull-943-20200604-113520-281416b3-fedora-31-firefox-osbuild-cockpit-composer/log.html). I didn't test this locally yet, but (ab)use our CI to tell me whether this is sufficient.

martinpitt:
FTR, the net difference in the firefox config is https://github.com/cockpit-project/cockpit/commit/bd3086070cc6 . This is fully backwards compatible -- we landed that change in Cockpit with Firefox 69.

martinpitt:
So this mostly fixes it, but there are [two](https://logs.cockpit-project.org/logs/pull-988-20200604-133340-8a485cb5-fedora-31-firefox/log.html) [failures](https://logs.cockpit-project.org/logs/pull-988-20200604-133340-8a485cb5-fedora-32-firefox/log.html) still. In particular, TestSources.testBasic failed twice. I'll retry once to compare.

PR #991, opened 2020-06-05
test: Add rhel-8-3 in CI and use internal rhel repo

henrywang:


test/verify/check-image
```diff
@@ -142,9 +143,14 @@ class TestImage(composerlib.ComposerCase):
...
         selector = "{}-compose-name".format(uuid)
-        b.wait_text("#{}".format(selector), "httpd-server-0.0.1-OpenStack")
-        b.wait_text("li[aria-labelledby={}] [data-image-type=OpenStack]".format(selector),
+        if "fedora" in os.environ["TEST_OS"]:
+            b.wait_text("#{}".format(selector), "httpd-server-0.0.1-OpenStack")
+            b.wait_text("li[aria-labelledby={}] [data-image-type=OpenStack]".format(selector),
                     "OpenStack")
+        else:
+            b.wait_text("#{}".format(selector), "httpd-server-0.0.1-openstack")
+            b.wait_text("li[aria-labelledby={}] [data-image-type=openstack]".format(selector),
```
martinpitt, line 152: Is that the added value that we get from RHEL? different word capitalization? :-P

henrywang: @martinpitt, Yes. same word, different upper/lower case. Really wired.

larskarlitski: This was a bug in osbuild-composer 11, which we didn't catch because we don't running cockpit-composer's tests on it yet. I've filed https://github.com/osbuild/cockpit-composer/pull/993 as a temporary workaround, so that it doesn't break on the next image refresh.

PR #993, opened 2020-06-08
test/verify/check-image: don't special-case based on OS

larskarlitski:
375ccf07c introduced a special-case for Fedora, expecting different
capitalization for an image type (`openstack` vs `OpenStack`). This was
a bug in osbuild-composer-11 which has since been fixed.

Allow both capitalizations for now, so that this test doesn't fail
against the new version of osbuild-composer. This can be removed in the
future.

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -3,7 +3,6 @@
 # checkpoint:
 # 1. create supported image
 
-import os
 import composerlib
 import testlib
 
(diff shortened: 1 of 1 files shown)


larskarlitski:
Update: fix `flake8` errors.

martinpitt:
[testExport](https://logs.cockpit-project.org/logs/pull-993-20200608-072724-37159a73-fedora-32-firefox/log.html) failure is unrelated, but [testOpenstack](https://logs.cockpit-project.org/logs/pull-993-20200608-072724-37159a73-fedora-32-firefox/log.html#20) is the very test case that is touched here, so retrying.

larskarlitski:
> [testExport](https://logs.cockpit-project.org/logs/pull-993-20200608-072724-37159a73-fedora-32-firefox/log.html) failure is unrelated, but [testOpenstack](https://logs.cockpit-project.org/logs/pull-993-20200608-072724-37159a73-fedora-32-firefox/log.html#20) is the very test case that is touched here, so retrying.

I'm looking into those separately, because they seem flaky. The failure was not related to my change though and running again worked.

PR #994, opened 2020-06-08
Modals: add ability to create ostree image

jkozol:
A user can create a fedora-iot-commit or a rhel-edge-commit through the CreateImageUpload Wizard. If a user selects either of these image types they are required to enter an ostree parent and ref.

ostreeFields

@jgiardino I disable the create button if the user does not enter the required Parent and Branch fields but I do not display an explicit error. Do you think I should use an alert on the input box like we do for an invalid image size?

invalidImageSize

### DIFF
--- a/components/Modal/CreateImageUpload.js
+++ b/components/Modal/CreateImageUpload.js
@@ -147,6 +147,14 @@ const ariaLabels = defineMessages({
   azure: {
     id: "azure-help",
     defaultMessage: "Azure help"
+  },
+  ostreeParent: {
+    id: "ostree-parent-help",
+    defaultMessage: "OSTree parent help"
+  },
+  ostreeRef: {
+    id: "ostree-ref-help",
+    defaultMessage: "OSTree ref help"
   }
 });
 
@@ -189,20 +197,28 @@ class CreateImageUploadModal extends React.Component {
       imageSize: "",
       minImageSize: 0,
       maxImageSize: 2000,
+      ostreeSettings: {
+        parent: undefined,
+        ref: undefined
+      },
       showUploadAwsStep: false,
       showUploadAzureStep: false,
       showReviewStep: false,
       uploadService: "",
       uploadSettings: {}
     };
+    this.disableCreateButton = this.disableCreateButton.bind(this);
     this.getDefaultImageSize = this.getDefaultImageSize.bind(this);
     this.isPendingChange = this.isPendingChange.bind(this);
     this.isValidImageSize = this.isValidImageSize.bind(this);
+    this.requiresImageSize = this.requiresImageSize.bind(this);
     this.missingRequiredFields = this.missingRequiredFields.bind(this);
     this.setNotifications = this.setNotifications.bind(this);
     this.setImageSize = this.setImageSize.bind(this);
     this.setImageName = this.setImageName.bind(this);
     this.setImageType = this.setImageType.bind(this);
+    this.setOstreeParent = this.setOstreeParent.bind(this);
+    this.setOstreeRef = this.setOstreeRef.bind(this);
     this.setUploadSettings = this.setUploadSettings.bind(this);
     this.handleUploadService = this.handleUploadService.bind(this);
     this.handleCreateImage = this.handleCreateImage.bind(this);
@@ -256,6 +272,14 @@ class CreateImageUploadModal extends React.Component {
     });
   }
 
+  setOstreeParent(value) {
+    this.setState(prevState => ({ ostreeSettings: Object.assign({}, prevState.ostreeSettings, { parent: value }) }));
+  }
+
+  setOstreeRef(value) {
+    this.setState(prevState => ({ ostreeSettings: Object.assign({}, prevState.ostreeSettings, { ref: value }) }));
+  }
+
   setImageType(imageType) {
     const defaultImageSize = this.getDefaultImageSize(imageType);
     this.setState({
@@ -263,6 +287,10 @@ class CreateImageUploadModal extends React.Component {
       imageName: "",
       imageSize: defaultImageSize,
       minImageSize: defaultImageSize,
+      ostreeSettings: {
+        parent: undefined,
+        ref: undefined
+      },
       uploadService: "",
       uploadSettings: {},
       showUploadAwsStep: false,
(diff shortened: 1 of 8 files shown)


henrywang:
@jkozol, I have a question about this PR. What if there's no parent commit?

jgiardino:
> @jgiardino I disable the create button if the user does not enter the required Parent and Branch fields but I do not display an explicit error. Do you think I should use an alert on the input box like we do for an invalid image size?

Ultimately we should refactor the interaction pattern we follow for form completion, so that we leave the submit button enabled always, and flag errors when clicked. But for now, what you have is consistent with how we're handling required fields.

However, I'm cc'ing @teg here, because I'm not sure parent comment and branch should be required.

teg:
Loving this work!

`Parent` should not be required, and if 'Branch' is left out we do have a default value we fall back to (`fedora/32/{arch}/iot` and `rhel/8/{arch}/edge`, respectively), so I think that could be left optional too.

As to the nomenclature, are we sure `Branch` is the right term? Internally, this is always referred to as `Ref`, so I just wanted to make sure. @gicmo, @nullr0ute, @mrguitar?

jgiardino:
Can we include help text for these fields? I'm good with adding that to a later PR, but while we have 👀 , I'd love suggestions.

**For parent commit:** Provide the ID of the latest commit in the updates repository for which this commit provides an update.

**For the other one, whose label is to be determined:** (I'm a little clueless on what both the label and help text should be. Any suggestions would be appreciated. Including defaults would also be helpful, if possible.)

Also, just a reminder that all strings should be sentence-style capitalization. "Parent Commit" should be "Parent commit".

larskarlitski:
Nice work! Needs a test :wink: 

> As to the nomenclature, are we sure `Branch` is the right term? Internally, this is always referred to as `Ref`, so I just wanted to make sure. @gicmo, @nullr0ute, @mrguitar?

Also, @gicmo mentioned we might need a URL as well.

jkozol:
@jgiardino  

> Ultimately we should refactor the interaction pattern we follow for form completion, so that we leave the submit button enabled always, and flag errors when clicked. But for now, what you have is consistent with how we're handling required fields.

Okay. I agree that this will need a refactor.

> For parent commit: Provide the ID of the latest commit in the updates repository for which this commit provides an update.

This sounds good to me. For the ref I am thinking of using:

`Ref: The name of the branch for the content`

test/verify/check-image
```diff
@@ -192,6 +192,84 @@ class TestImage(composerlib.ComposerCase):
...
 
+    def testOSTree(self):
+        b = self.browser
+        m = self.machine
+        if (m.image == "fedora-32"):
+            image_type_ostree = "fedora-iot-commit"
+        elif (m.image == "rhel-83"):
+            image_type_ostree = "rhel-edge-commit"
+        else:
+            return
```
jkozol, line 203: @henrywang I would appreciate your advice on this test. We should only build an ostree image on compatible systems. Do you have any recommendations?

components/Modal/CreateImageUpload.js
```diff
@@ -374,11 +429,75 @@ class CreateImageUploadModal extends React.Component {
...
+            type="text"
+            id="ostree-parent-input"
+            onChange={this.setOstreeParent}
+          />
+        </div>
+        <div className="pf-c-form__group">
+          <div className="pf-c-form__label pf-m-no-padding-top pf-l-flex pf-u-display-flex pf-m-justify-content-flex-start pf-m-nowrap">
+            <label htmlFor="ostree-ref-input" className="pf-l-flex__item">
+              <span className="pf-c-form__label-text">
+                <FormattedMessage defaultMessage="Branch" />
```
jgiardino, line 473: Did you mean to change this to "Ref"? I thought when chatting with @gicmo we discussed displaying "Ref" as the input label, and explaining that it is a branch in the help text.

jkozol: Thanks for the catch. I updated this to `Ref` now.

henrywang:
@jkozol, I got the same building error. Your test code looks good.
```shell
Assembler org.osbuild.ostree.commit:
{
  "ref": "fedora/32/x86_64/iot",
  "tar": {
    "filename": "commit.tar"
  }
}

Output:
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
Committing...done
error: While writing rootfs to mtree: Writing content object: min-free-space-percent '3%' would be exceeded, at least 2.7 MB requested
Initializing root filesystem
Copying data
  File "/run/osbuild/lib/assemblers/org.osbuild.ostree.commit", line 167, in <module>
    r = main(args["tree"], args["output_dir"], args["options"])
  File "/run/osbuild/lib/assemblers/org.osbuild.ostree.commit", line 145, in main
    subprocess.run(argv,
  File "/usr/lib64/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['rpm-ostree', 'compose', 'commit', '--repo=/run/osbuild/output/repo', '--write-composejson-to=/run/osbuild/output/compose.json', '/tmp/tmppxc370y2.json', '/run/osbuild/output/tmps5w_5v1s']' returned non-zero exit status 1.
```

test/verify/check-image
```diff
@@ -192,6 +192,85 @@ class TestImage(composerlib.ComposerCase):
         # collect code coverage result
         self.check_coverage()
 
+    def testOSTree(self):
```
henrywang, line 195: Can we skip this test on rhel8 and fedora-31? It's more reasonable and efficient. Thanks
```python
@unittest.skipUnless(os.environ.get("TEST_OS") == "fedora-32", "Do not support ostree image")
```
And we can changed to `skipIf` when ostree image lands on rhel8
```python
@unittest.skipIf(os.environ.get("TEST_OS") == "fedora-31", "Do not support ostree image")
```
Please `import os` and `import unittest`.

jkozol: I added this. Thanks!

larskarlitski:
This looks really cool and seems to work as well. I'll give it another go later.

I wonder why the rhel-8-3 tests didn't run though. These are kind of important for this one.

larskarlitski:
> I wonder why the rhel-8-3 tests didn't run though. These are kind of important for this one.

Ah, they wouldn't pas, because we don't have the right osbuild-composer in there.

larskarlitski:
I've added two commits: 

1. skip the ostree test on RHEL. We cannot run those tests yet, because osbuild-composer is not yet in RHEL.
2. Bump the osbuild-composer version to the one it actually depends on.

PR #995, opened 2020-06-08
tests / build: move closer to what other cockpit project are doing

larskarlitski:
More consistency means that it's easier for others to help with problems.

The main commit moves this project to use all of cockpit's upstream test framework, dropping `test/verify/run-tests` (sorry @henrywang!). This also actually includes the source code and build tools in the released tarballs.


Makefile
```diff
@@ -30,31 +37,38 @@ po-push: po/cockpit-composer.pot $(WEBLATE_REPO)
...
 
-dist-gzip: NODE_ENV=production
-# package test dependence to avoid clone or fetch
-dist-gzip: all $(PACKAGE_NAME).spec machine test/common
-	mkdir -p $(PACKAGE_NAME)-$(VERSION)
-	cp -r public/ test/ LICENSE.txt README.md $(PACKAGE_NAME).spec io.weldr.cockpit-composer.metainfo.xml $(PACKAGE_NAME)-$(VERSION)
-	tar -czf $(PACKAGE_NAME)-$(VERSION).tar.gz $(PACKAGE_NAME)-$(VERSION)
-	rm -rf $(PACKAGE_NAME)-$(VERSION)
+
+dist-gzip: $(TARFILE)
```
martinpitt, line 56: Nice, as a side issue this should also fix `make` to not always rebuild everything even if nothing changed.

PR #998, opened 2020-06-09
test: Update test to support aarch64

henrywang:


larskarlitski:
Oh, is it really that much slower? Maybe we can increase some global timeout instead of adding one to most calls?

martinpitt:
I mentioned something similar on IRC, like setting `self.browser.cdp.timeout = 300` only(!) on aarch64 (and possibly other slow architectures, but not x86). This would be a much less intrusive patch, but at the same time also make failing tests unnecessarily slow.

But I think a nice compromise would be if the `ComposerTest` class overrides `login_and_go()` with that code, so that the hack doesn't need to be repeated. IMHO that should be done instead of landing it as-is, but I don't want to be the final arbiter here. :-)

henrywang:
Thanks @martinpitt. Overriding is a great suggestion. Updated.

test/verify/composerlib.py
```diff
@@ -52,6 +52,11 @@ def setUp(self):
                         "for bp in $(composer-cli blueprints list); "
                         "do composer-cli blueprints delete $bp; done")
 
+    # Hack to add more wait time to work with aarch64 platform test
+    def login_and_go(self, *args, **kwargs):
+        with self.browser.wait_timeout(300):
```
martinpitt, line 57: This could use 300 only on aarch64, and otherwise keep the default, so that a failing test case doesn't regularly sit around for 5 minutes until it ends. But that's not important at all.

PR #1001, opened 2020-06-11
 Stop importing cockpit's base1/patternfly.css

KKoukiou:


larskarlitski:
Thank you :heart: 

> This is deprecated API and will be dropped at some point, in favor of projects shipping their own CSS.

Do you already know when this "some point" will be?

KKoukiou:
> Thank you heart
> 
> > This is deprecated API and will be dropped at some point, in favor of projects shipping their own CSS.
> 
> Do you already know when this "some point" will be?

We discussed about fedora-33 and rhel-8-4 but there are not yet any changes in the spec files in cockpit. We are trying to keep track of all projects using this API, and I am trying to do the same change to all of them. So the above seem reasonable to me assuming that I am taking over the responsibility to port to the new PF consumption pattern.

jgiardino:
Thanks for making these changes! I poked around and most things look good, but there are a couple of issues with font styling and spacing. I'm happy to work on the updates needed for those issues that I found. Let me know if you have suggestions on how to move forward with changes (i.e. push additional commits on top of this PR, or some other approach?).

larskarlitski:
> Let me know if you have suggestions on how to move forward with changes

Doesn't matter where we fix those, but I wouldn't land this PR before we do. I think I'd work in a new PR based on @KKoukiou's commits.

KKoukiou:
> Thanks for making these changes! I poked around and most things look good, but there are a couple of issues with font styling and spacing.

Can you point the exact problems? This solution should work for all cockpit projects without workarounds, so if something is broken here, it will probably be problematic in other projects  as well and I would like a unified solution.

> I'm happy to work on the updates needed for those issues that I found. Let me know if you have suggestions on how to move forward with changes (i.e. push additional commits on top of this PR, or some other approach?).

You can push seperate commit in this PR, or just send the screenshots where the problems appear and I can see what I can do.

KKoukiou:
@jgiardino I checked your comments regarding users table header fonts, and this is not a regression caused by this PR - it's the same on master.
But since we saw that issue I sent a PR which converts that table to PF4 so the fonts will be picked up correctly now.

See the relevant PR in https://github.com/osbuild/cockpit-composer/pull/1013

martinpitt:
There's a trivial conflict. Force-pushed to fix that, as Katerina is on holidays now.

@jgiardino : As #1013 landed now, this seems ready?

jgiardino:
Thanks, Martin!

I took another look and the only big difference I see now is with the font size in the empty state. I included a before/after of one of the empty states below. I would like to wait until we have updates to the Empty State available before merging this. Is there any concern with waiting to merge this until a later cycle? 

Screen Shot 2020-06-26 at 3 28 51 PM
Screen Shot 2020-06-26 at 3 21 19 PM

martinpitt:
@jgiardino : Not a big concern, it's just a relatively large PR, so it tends to bitrot quickly. But of course it shouldn't be merged with regressions. Thanks for thorough testing!

KKoukiou:
> Thanks, Martin!
> 
> I took another look and the only big difference I see now is with the font size in the empty state. I included a before/after of one of the empty states below. I would like to wait until we have updates to the Empty State available before merging this. Is there any concern with waiting to merge this until a later cycle?
> 
> Screen Shot 2020-06-26 at 3 28 51 PM
> 
> Screen Shot 2020-06-26 at 3 21 19 PM

@jgiardino this in unfortunately not reproducible on my side. I see the 'before' screenshot even with my PR. Can you please recheck? And if you still see it show me the screenshot from your dev tools on the title element that has different font weight and size for you.

jgiardino:
@KKoukiou, here is a screenshot:

image

KKoukiou:
@jgiardino I saw the issue you noticed, I could either override the font-size for that h1 element, but instead I thought it would be nicer to fix the EmptyState component to be actually PF4 and user smaller font size for the header. 
Let me know if you like it now.

Screen Shot 2020-07-08 at 12 43 43

jgiardino:
👍 I think your suggestion makes sense. I also was thinking that  updating the EmptyState should happen before/with these changes. 

Looking across different empty states, the only remaining issues I see are the following: 
- The spacing for the actions seem off. The styles for spacing  seem dependent on the presence of the EmptyStatePrimary component, which is being pulled into the Blueprints page, but not the examples shown below. 
  - What are your thoughts on this? Should this update be included here or a separate PR? 
  - As a side note, the version we're pulling in is different than the latest examples on pf.org, which no longer include the use of the EmptyStatePrimary component, but does require the Button component to be used to get the expected spacing. So if we did stick with the current version of pf-react, and made this change, we'd need to pull it out once we update to the latest version. 
  - 🤷 
- There's some inconsistency with which variant is being used for the action that displays (i.e. as a link or secondary instead of primary). I think this can definitely be handled in a separate PR, but wanted to mention it here since it is somewhat related to the previous item.

image

image

KKoukiou:
> +1 I think your suggestion makes sense. I also was thinking that updating the EmptyState should happen before/with these changes.
> 
> Looking across different empty states, the only remaining issues I see are the following:
> 
>     * The spacing for the actions seem off. The styles for spacing  seem dependent on the presence of the EmptyStatePrimary component, which is being pulled into the Blueprints page, but not the examples shown below.
>       
>       * What are your thoughts on this? Should this update be included here or a separate PR?
>       * As a side note, the version we're pulling in is different than the latest examples on pf.org, which no longer include the use of the EmptyStatePrimary component, but does require the Button component to be used to get the expected spacing. So if we did stick with the current version of pf-react, and made this change, we'd need to pull it out once we update to the latest version.
>       * shrug'

For that you will need to start using PF4 buttons, now you use PF3 only. And of course the spacing rule does not apply with these.
Let's keep it like this, add a CSS rule for the spacing and change your PF3 buttons to PF4 in another PR and then we can delete the WA and remove also the EmptyStatePrimary wrapper component.

> 
>     * There's some inconsistency with which variant is being used for the action that displays (i.e. as a link or secondary instead of primary). I think this can definitely be handled in a separate PR, but wanted to mention it here since it is somewhat related to the previous item.
> 
> 
> image
> 
> image

KKoukiou:
@jgiardino here are the screenshots with the updated code:

Screen Shot 2020-07-09 at 10 42 32
Screen Shot 2020-07-09 at 10 41 48
Screen Shot 2020-07-09 at 10 41 38

jgiardino:
This works for me. Thanks for making these updates!

@jkozol are you good with this?

PR #1004, opened 2020-06-15
test: add rhel ostree test

jkozol:
With the added support for both fedora-iot-commits and rhel-edge-commits the ostree test now runs on both fedora-32 and rhel-8.

This test won't pass until osbuild-composer v15 is available.


larskarlitski:
Nice! Note: I also made the resizing conditional on Fedora (in a bit of a hacky way): https://github.com/osbuild/cockpit-composer/blob/master/test/vm.install#L7

jkozol:
This depends on #1023 and #1024

PR #1007, opened 2020-06-17
test: Stop hardcoding the VG name

martinpitt:
It is different in recent Fedora 31/32 images.


martinpitt:
Untested locally; I triggered a run against https://github.com/cockpit-project/bots/pull/987 and https://github.com/cockpit-project/bots/pull/989 as well to make sure that it works with the old and new  images.

martinpitt:
It's past vm.install everywhere, so this works. I can haz review, please?

PR #1008, opened 2020-06-18
test: increase testDiscard timeouts

### LINKED ISSUE #1005: check-package.TestPackage.testDiscard is flaky
The `check-package.TestPackage.testDiscard` test sometimes fails in RHEL gating. I couldn't reproduce this locally yet.

```
# 1 TEST FAILED [59s on ibm-p8-kvm-03-guest-02]
Started /usr/lib64/chromium-browser/headless_shell (pid 63546) on port 10103
  File "test/verify/check-package", line 219, in testDiscard
    b.click("li[data-input=cockpit-composer] .fa-plus")
  File "/var/str/source/test/common/testlib.py", line 215, in click
    self.mouse(selector + ":not([disabled])", "click", 0, 0, 0)
  File "/var/str/source/test/common/testlib.py", line 212, in mouse
    self.call_js_func('ph_mouse', selector, type, x, y, btn, ctrlKey, shiftKey, altKey, metaKey)
  File "/var/str/source/test/common/testlib.py", line 196, in call_js_func
    return self.eval_js("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/var/str/source/test/common/testlib.py", line 180, in eval_js
    silent=False, awaitPromise=True, returnByValue=True, no_trace=no_trace)
  File "/var/str/source/test/common/cdp.py", line 109, in invoke
    res = self.command(cmd)
  File "/var/str/source/test/common/cdp.py", line 136, in command
    raise RuntimeError(res["error"])
RuntimeError: TypeError: Cannot read property 'concat' of undefined
```

jkozol:
testDiscard is flaky when run on rhel gating tests. I have been unable to replicate a test failure so I assume the cause of the test failure is a timeout. The timeout is doubled and will hopefully improve the
reliability of the test.

This will hopefully resolve #1005

### DIFF
--- a/test/verify/check-package
+++ b/test/verify/check-package
@@ -207,13 +207,13 @@ class TestPackage(composerlib.ComposerCase):
 
         # go to edit package page
         b.click("li[data-blueprint=openssh-server] a:contains('Edit Packages')")
-        with b.wait_timeout(120):
+        with b.wait_timeout(240):
             b.wait_present("ul[aria-label='Available Components'] li:nth-child(1)")
 
         # search for openssh-server pacakge
         b.set_input_text("#cmpsr-blueprint-input-filter", "cockpit-composer")
         b.key_press("\r")
-        with b.wait_timeout(120):
+        with b.wait_timeout(240):
             b.wait_text("#cockpit-composer-input", "cockpit-composer")
         # add it to blueprint
         b.click("li[data-input=cockpit-composer] .fa-plus")


marusak:
> This will hopefully resolve #1005

I don't think that is related. #1005 is about the fact, then when you click on `li[data-input=cockpit-composer] .fa-plus` somewhere in the code you try to call `.concat` on `undefined`.
(`git grep concat` can give you list of possible places).

The command you wrap around does not timeout, so that won't be it. Maybe you need to wait for something else as well? (It is then questionable if user would not hit it as well...)

jkozol:
@marusak Okay I will close this then and keep looking

Issue #1009, opened 2020-06-18
coverity: implicit_this_used


larskarlitski:
Red Hat's coverity reported this issue: 

```
cockpit- composer/components/ListView/ComponentDetailsView.js:210: implicit_this_used: Calling function "EditBlueprintPage.handleUpdateComponent" with implicit 'this' argument when explicit 'this' is expected.
```

PR #1010, opened 2020-06-18
components: fix update component arrow function

jkozol:
In the component details view the handleUpdateComponent is called in an arrow function. This function needs to be wrapped by curly brackets.

This should help with #1009

### DIFF
--- a/components/ListView/ComponentDetailsView.js
+++ b/components/ListView/ComponentDetailsView.js
@@ -206,9 +206,9 @@ class ComponentDetailsView extends React.Component {
                     <button
                       className="btn btn-primary"
                       type="button"
-                      onClick={e =>
-                        handleUpdateComponent(e, component.name, component.builds[selectedBuildIndex].version)
-                      }
+                      onClick={e => {
+                        return handleUpdateComponent(e, component.name, component.builds[selectedBuildIndex].version)
+                      }}
                     >
                       <FormattedMessage defaultMessage="Apply Change" />
                     </button>


jkozol:
@larskarlitski I agree that both the `handleAddComponent` and `handleRemoveComponent` are similar calls. The only difference I can find between these three calls is that `handleAddComponent` and `handleRemoveComponent` are inline whereas `handleUpdateComponent` requires a new line due to line length limits and in the es6 arrow function docs all non inline functions were wrapped in curly brackets. I am not entirely sure this will fix anything and am looking into testing against coverity to actually know.

gicmo:
In 14.2 from [ECMA-262](http://www.ecma-international.org/ecma-262/6.0/ECMA-262.pdf):
```
ArrowFunction[In, Yield]:
   ArrowParameters[?Yield][no LineTerminatorhere]=>ConciseBody[?In]

ConciseBody[In]:
   [lookahead ≠ {]AssignmentExpression[?In]
   {FunctionBody }
```
`AssignmentExpression` (12.14) is quite complex:
```
AssignmentExpression[In, Yield]:
  ConditionalExpression[?In, ?Yield]
  [+Yield]YieldExpression[?In]
  ArrowFunction[?In, ?Yield]
  LeftHandSideExpression[?Yield] =AssignmentExpression[?In, ?Yield]
  LeftHandSideExpression[?Yield]AssignmentOperator AssignmentExpression[?In, ?Yield]
```
with `Expression` (12.15):
```
Expression[In, Yield]:
  AssignmentExpression[?In, ?Yield]
  Expression[?In, ?Yield],AssignmentExpression[?In, ?Yield]
```
we end up with `Statement` (13):

```
Statement[Yield, Return]:
  BlockStatement[?Yield, ?Return]
  VariableStatement[?Yield]
  EmptyStatement
  ExpressionStatement[?Yield]
  [...]
```
Glancing at this I don't think a newline between ` => ` and the rest is ok, so that patch makes sense to me. But I only spend like 5 mins looking now. How this triggers the coverity warning or if that is a false positive of a side effect of this maybe malformed js code is beyond my knowledge ;D

jkozol:
@larskarlitski Are you okay merging this? This is my best guess to fix the coverity issue and doesn't have any impact on functionality.

jgiardino:
Regarding the checks: RHEL tests passed, but Fedora tests will fail until dependency issues with osbuild-composer and osbuild are resolved. I went ahead and merged so that we can get these changes in our next release for RHEL.

PR #1013, opened 2020-06-19
Port blueprint customizations users table to PF4

KKoukiou:
Before:
Screen Shot 2020-06-19 at 17 23 14

After:
Screen Shot 2020-06-19 at 17 22 43

### DIFF
--- a/pages/blueprint/index.css
+++ b/pages/blueprint/index.css
@@ -1,0 +1,9 @@
+.user-list .pf-c-table {
+  margin-bottom: 2rem;
+}
+.user-list .control-label {
+  padding-top: .375rem;
+}
+.user-list th {
+  padding-top: .375rem;
+}
--- a/pages/blueprint/index.css
+++ b/pages/blueprint/index.css
@@ -1,9 +1,11 @@
 .user-list .pf-c-table {
   margin-bottom: 2rem;
 }
+
 .user-list .control-label {
-  padding-top: .375rem;
+  padding-top: 0.375rem;
 }
+
 .user-list th {
-  padding-top: .375rem;
+  padding-top: 0.375rem;
 }
--- a/pages/blueprint/index.js
+++ b/pages/blueprint/index.js
@@ -5,6 +5,12 @@ import { FormattedMessage, defineMessages, injectIntl, intlShape } from "react-i
 import cockpit from "cockpit"; // eslint-disable-line import/no-unresolved
 import PropTypes from "prop-types";
 import { Tab, Tabs } from "patternfly-react";
+import {
+  Table,
+  TableHeader,
+  TableBody,
+  TableVariant,
+} from '@patternfly/react-table';
 import { connect } from "react-redux";
 import Link from "../../components/Link/Link";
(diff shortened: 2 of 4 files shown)


pages/blueprint/index.js
```diff
@@ -407,75 +464,14 @@ class BlueprintPage extends React.Component {
...
-                                          </a>
-                                        </li>
-                                      </ul>
-                                    </div>
-                                  </td>
-                                </tr>
-                              ))}
-                            </tbody>
-                          </table>
+                        <div className="user-list">
```
jgiardino, line 467: Can `.user-list` be moved further up the DOM to the parent `div.form-group`? This would allow us to add a few more styles to create better vertical alignment among the label "Users" and the table headers. For example:
```
.user-list .control-label {
  padding-top: .375rem;
}
.user-list th {
  padding-top: .375rem;
} /* this one might need more specificity 🤷 */
```

The `.375rem` is based on the padding-top currently applied to the "Create User Account" button. Also note, I would consider these styles as a temporary fix, until the rest of the page is refactored.

pages/blueprint/index.js
```diff
@@ -310,6 +318,55 @@ class BlueprintPage extends React.Component {
...
+          { title: (
+            user.key !== undefined && (
+              <span>
+                <span className="fa fa-check" />
+                {` `}
+                {user.key.split(" ")[2] || ""}
+              </span>
+            )
+          ) },
+        { title: (<>
```
jgiardino, line 335: Can you change this `<>` to `<div>`? In the mobile layout, each child is placed in a grid column, spanning the full width, which results in the buttons being stacked. Having a parent `<div>` as the child instead of the separate buttons yields a better layout for this view.

jgiardino:
Here are before/after examples relative to my previous comments.

_Regarding my first comment about additional styles:_

Before:
image


After:
image


Before: 
image

After: 
image

_Regarding my second comment about when the table switches to the mobile layout:_

Before:
image

After:
image

KKoukiou:
@jgiardino thanks for the css fixes - I included them and reposted.

pages/blueprint/index.css
```diff
@@ -0,0 +1,9 @@
+.user-list {
```
jgiardino, line 1: I didn't mention this yesterday, but now that `.user-list` is higher in the DOM, can you add `.pf-c-table` to this one (e.g. `.user-list .pf-c-table`)?

jgiardino:
Thanks for making those updates! 🎉  I checked the latest and everything looks great, except for one minor thing that I just now commented on.

pages/blueprint/index.css
KKoukiou: I adjusted but I had to change the padding to margin, for some reason padding on the table element was not applied.

jgiardino:
Regarding the checks: RHEL tests passed, but Fedora tests will fail until dependency issues with osbuild-composer and osbuild are resolved. I went ahead and merged so that we can get these changes in our next release for RHEL.

PR #1018, opened 2020-06-25
pages: don't render input list until blueprint is loaded

### LINKED ISSUE #1005: check-package.TestPackage.testDiscard is flaky

jkozol:
The input list allows users to add and remove components from a blueprint. These actions can not occur if the blueprint is still being fetched. By disabling the input list until the blueprint is loaded the
user is prevented from interacting with the list when it may reflect inaccurate data. 

This will hopefully fix #1005. This test fails when adding a component to the blueprint due to an issue concatenating. The handleAddComponent function concats twice with the blueprint components and blueprint packages. Conditionally rendering the input list should fix this issue.

### DIFF
--- a/pages/blueprintEdit/index.js
+++ b/pages/blueprintEdit/index.js
@@ -571,7 +571,7 @@ class EditBlueprintPage extends React.Component {
         <h3 className="cmpsr-panel__title cmpsr-panel__title--sidebar">
           {formatMessage(messages.listTitleAvailableComps)}
         </h3>
-        {(inputComponents !== undefined && (
+        {((inputComponents !== undefined && blueprint.components !== undefined && blueprint.packages !== undefined) && (
           <div className="cmpsr-panel__body cmpsr-panel__body--sidebar">
             <div className="toolbar-pf">
               <form className="toolbar-pf-actions">
@@ -618,7 +618,7 @@ class EditBlueprintPage extends React.Component {
                 />
               </div>
             </div>
-            {(blueprint.components === undefined || blueprint.components.length === 0) &&
+            {blueprint.components.length === 0 &&
               Object.keys(this.props.blueprintContentsError).length === 0 &&
               componentsFilters.filterValues.length === 0 && (
                 <div className="alert alert-info alert-dismissable">
@@ -644,13 +644,13 @@ class EditBlueprintPage extends React.Component {
                   />
                 </div>
               )}
-            <ComponentInputs
-              label={formatMessage(messages.listTitleAvailableComps)}
-              components={inputComponents[inputs.selectedInputPage]}
-              handleComponentDetails={this.handleComponentDetails}
-              handleAddComponent={this.handleAddComponent}
-              handleRemoveComponent={this.handleRemoveComponent}
-            />
+              <ComponentInputs
+                label={formatMessage(messages.listTitleAvailableComps)}
+                components={inputComponents[inputs.selectedInputPage]}
+                handleComponentDetails={this.handleComponentDetails}
+                handleAddComponent={this.handleAddComponent}
+                handleRemoveComponent={this.handleRemoveComponent}
+              />
           </div>
         )) || (
           <div className="cmpsr-panel__body cmpsr-panel__body--sidebar">


pages/blueprintEdit/index.js
```diff
@@ -644,13 +644,15 @@ class EditBlueprintPage extends React.Component {
...
                 </div>
               )}
-            <ComponentInputs
-              label={formatMessage(messages.listTitleAvailableComps)}
-              components={inputComponents[inputs.selectedInputPage]}
-              handleComponentDetails={this.handleComponentDetails}
-              handleAddComponent={this.handleAddComponent}
-              handleRemoveComponent={this.handleRemoveComponent}
-            />
+            {(blueprint.components !== undefined && blueprint.packages !== undefined) &&
```
jgiardino, line 647: I'm wondering if this condition should be moved [higher](https://github.com/osbuild/cockpit-composer/pull/1018/files#diff-268ff6cc6477c8bc85a815bc0b236684R574) and combined with the following condition:
`{(inputComponents !== undefined && (`

Because including this condition here results in the following result:
image

But if we move it higher, I think we'd continue to see the loading state while we wait for the blueprint to finish loading, which seems like a better behavior:
image

BTW, I can't actually test this condition, because the blueprint always loads faster than the inputs, and I'm not sure how to reproduce the reverse.

jkozol: Thanks for the suggestion! I moved the conditional up. I can also not reproduce it. This is just my best attempt to fix #1005 without being able to replicate it.

henrywang:
@larskarlitski, @gicmo: There're two selinux error on osbuild:
```
audit: type=1400 audit(1593198435.846:2947): avc:  denied  { nnp_transition nosuid_transition } for  pid=36979 comm="semodule" scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:setfiles_mac_t:s0 tclass=process2 permissive=0
audit: type=1401 audit(1593198435.846:2947): op=security_bounded_transition seresult=denied oldcontext=system_u:system_r:osbuild_t:s0 newcontext=system_u:system_r:setfiles_mac_t:s0
audit: type=1400 audit(1593198492.978:2948): avc:  denied  { nnp_transition nosuid_transition } for  pid=47083 comm="dracut" scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:setfiles_mac_t:s0 tclass=process2 permissive=0
audit: type=1401 audit(1593198492.978:2948): op=security_bounded_transition seresult=denied oldcontext=system_u:system_r:osbuild_t:s0 newcontext=system_u:system_r:setfiles_mac_t:s0
```
Is this bug or just ignore them? Thanks!

jkozol:
@henrywang I moved this discussion to #1022. Since this PR passes the travis and rhel-8-3 tests I am going to merge it to master and include it in the next rhel release.

Issue #1022, opened 2020-06-28
testOSTree failing due to avc denial


jkozol:
The OSTree test is failing on fedora due to the following journal message:

```
audit: type=1400 audit(1593198435.846:2947): avc:  denied  { nnp_transition nosuid_transition } for  pid=36979 comm="semodule" scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:setfiles_mac_t:s0 tclass=process2 permissive=0
```

jkozol:
@larskarlitski, @gicmo: There're two selinux error on osbuild:
```
audit: type=1400 audit(1593198435.846:2947): avc:  denied  { nnp_transition nosuid_transition } for  pid=36979 comm="semodule" scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:setfiles_mac_t:s0 tclass=process2 permissive=0
audit: type=1401 audit(1593198435.846:2947): op=security_bounded_transition seresult=denied oldcontext=system_u:system_r:osbuild_t:s0 newcontext=system_u:system_r:setfiles_mac_t:s0
audit: type=1400 audit(1593198492.978:2948): avc:  denied  { nnp_transition nosuid_transition } for  pid=47083 comm="dracut" scontext=system_u:system_r:osbuild_t:s0 tcontext=system_u:system_r:setfiles_mac_t:s0 tclass=process2 permissive=0
audit: type=1401 audit(1593198492.978:2948): op=security_bounded_transition seresult=denied oldcontext=system_u:system_r:osbuild_t:s0 newcontext=system_u:system_r:setfiles_mac_t:s0
```
Is this bug or just ignore them? Thanks!

_Originally posted by @henrywang in https://github.com/osbuild/cockpit-composer/pull/1018#issuecomment-650674334_

Moving this discussion here because I am closing that PR due to it passing the rhel tests.

jkozol:
@henrywang this should be fixed on osbuild-composer's side by https://github.com/osbuild/osbuild-composer/pull/799

jkozol:
This was temporarily fixed by #1023

PR #1023, opened 2020-06-29
test: allow selinux journal messages

jkozol:
The ostree test was failing due to an avc and selinux denial. The test has been updated to ignore these journal messages. This is a temporary fix for #1022 until these messages are no longer printed.


jkozol:
This depends on #1024

PR #1024, opened 2020-06-29
package.json: update eslint loader

jkozol:
This is an attempt to fix the error:

`error  Prop type `object` is forbidden  react/forbid-prop-types`


martinpitt:
That was a pretty bad workaround - you don't want to pin an old eslint version forever. Why not fix the actual errors instead? Or, if you don't like a particular check, you can disable it in .eslintrc.

PR #1028, opened 2020-07-01
Update eslint to 7.2.0 and update eslint-* to latest version

henrywang:
The update found a lot of eslint errors, most of them have been fixed in this commit. Some of rules which found errors are disabled in this commit. They will be re-enabled and errors will be fixed in other commit.


henrywang:
Oh, osbuild-* are still not `stable` on fedora-31.

jkozol:
@henrywang The osbuild dependency issue for fedora-31 will be resolved tomorrow since the new version of osbuild is on its way into stable.

PR #1034, opened 2020-07-06
Fix pagination

### LINKED ISSUE #842: Pagination text input isn't editable
### Description of problem
On the Edit Packages page, there is a pagination control for the list of available components on the left. The pagination originally allowed the user to specify an exact page number in the text input. This text input is no longer editable.

### Steps to Reproduce

1. Navigate to the Edit Packages page
2. Set focus in the input in the pagination UI for Available Components
3. Type a valid value

### Actual results
The value is updated in the input.

### Expected results
The value does not change.

### Additional info
Originally, this input prevented the user from specifying an invalid page value. That interaction is expected as part of a fix for this issue.

jkozol:
I have migrated the pagination to patternfly 4. I have also simplified the store. This closes #842.


jgiardino:
Thanks for making these updates! This is definitely an improvement to what was there before!! 

I'll provide feedback about layout inline, but here are is some feedback about the interaction with the control: 

1. Hit the Next page button. The loading message displays while fetching components. If I hit the Previous page button while the loading message is visible, I continue to see the loading message.
    - Is it possible to keep the previous pages that we fetched? So in this case, when I click the Previous button, I immediately see results?
2. Then when the loading disappears, first I see the contents from page 2, then a few moments later the contents from page 1.
    - Can we dismiss the previous fetch for page 2 if the user requests to go to another page?

jkozol:
@jgiardino Thanks for the feedback!

> Then when the loading disappears, first I see the contents from page 2, then a few moments later the contents from page 1.
    Can we dismiss the previous fetch for page 2 if the user requests to go to another page?

This is fixed.

> Hit the Next page button. The loading message displays while fetching components. If I hit the Previous page button while the loading message is visible, I continue to see the loading message.
>
> Is it possible to keep the previous pages that we fetched? So in this case, when I click the Previous button, I immediately see results?

This is possible but I would prefer to create a separate issue for this.

pages/blueprintEdit/index.js
```diff
@@ -569,17 +563,24 @@ class EditBlueprintPage extends React.Component {
...
+                <TextInput
+                  className="pf-m-flex-1"
+                  type="text"
+                  id="cmpsr-blueprint-input-filter"
+                  aria-label={formatMessage(messages.filterByLabel)}
+                  placeholder={formatMessage(messages.filterByPlaceholder)}
+                  onKeyPress={(e) => this.getFilteredInputs(e)}
+                />
+                <Pagination
+                  className="pf-m-flex-2"
```
jgiardino: There are a few issues with the layout as shown in the examples below. The input should fill the space available to it, but it gets really crunched. One of the reasons it gets crunched is the pagination control stays on the same row and wraps within it self. 

Screen Shot 2020-07-10 at 5 13 35 PM
Screen Shot 2020-07-10 at 5 13 16 PM

Here are some suggestions for modifying this:

- A ratio of 1:2 isn't always accurate for the `<TextInput>`:`<Pagination>` controls. The pagination is more or less a fixed size (i.e. it doesn't grow or shrink when it's container grows). Whereas the input is elastic and can shrink or grow to any width. I'd suggest that the input and pagination should be more of a 1:1 radio. I'd also suggest that the input needs a minimum width that is ~50% of the container (i.e. ~equal in length to the pagination control), so that it isn't forced to be shorter than it should at smaller viewports.
- Looking at how the beta Toolbar component is built, they apply `flex-wrap: nowrap` to the pagination control. In the examples above, the input and pagination remain split on the same row, but the elements inside the pagination component start to wrap. Instead, it would be better to keep the pagination elements all in a single line, and move them down to the second row, and allow the input to take up the full width. Instead of adding the beta Toolbar component now, we can use the utility class to apply `nowrap` and also set this on a breakpoint so that at the smallest viewport sizes, it can start to wrap. The class for this is `pf-u-flex-nowrap-on-[breakpoint].

The combination of these changes results in the filter refusing to shrink any smaller when it's half the width, but also not making the pagination component wrap within itself to accommodate the width of the input, therefore forcing the whole pagination component to a new row. 

Here are some examples after making these changes:
Screen Shot 2020-07-10 at 5 08 30 PM
Screen Shot 2020-07-10 at 5 08 45 PM
Screen Shot 2020-07-10 at 5 08 30 PM

jgiardino:
I posted a comment inline about updates needed for the layout. I also pushed a commit with specific changes related to these suggestions, but I somehow included another commit unintentionally (and would fix it if I knew how).

jkozol:
@jgiardino I fixed up the commit history and applied your changes. I believe it is all good now.

PR #1039, opened 2020-07-09
Bump rhel-8 branch to rhel-8-3

jkozol:


martinpitt:
https://github.com/cockpit-project/bots/pull/1057 landed, so triggering rhel-8-3 tests now.

jkozol:
@martinpitt Okay sounds good. We will be adding a new release Wednesday so the tests will pass then.

PR #1052, opened 2020-07-20
enable filtering with multiple values or wildcards

### LINKED ISSUE #1026: api: filtering with multiple values misses packages.
This has been moved from #249 

When a filter is entered that includes a space or comma-separated values, the UI should handle these separate values in the api call.

- For example, "emacs,vim" will currently get submitted as `/api/v0/modules/list/*emacs,vim*?limit=50&offset=0` which will return only packages that end with "emacs" or start with "vim". But instead, the call should be `/api/v0/modules/list/*emacs*,*vim*?...`
- The following filter values specified by the user should return the same results:
      - "x, y"
      - "x,y"
      - "x y"

jgiardino:
The API currently supports the following use cases in the string it accepts:
- User specifies multiple comma-separated values
- User specifies a wildcard

Currently, the UI takes whatever the user input is and wraps it with a wildcard. This results in a couple of issues in the case where the user intends to use a wildcard (e.g. "httpd*" for only packages starting with "httpd") or where the user wants to specify multiple filter values (as noted in #1026). 

This PR addresses both use cases:
- If multiple filter values are provided separated by comma and/or space, then these separators (i.e. " " or ", ") are converted to commas (i.e. ",").
- If the filter value includes at least one wildcard, then the filter value provided to the api uses the wildcards as entered, otherwise, the filter value is wrapped with wildcards.

fixes #1026

### DIFF
--- a/core/sagas/inputs.js
+++ b/core/sagas/inputs.js
@@ -41,7 +41,12 @@ function flattenInputs(response) {
 function* fetchInputs(action) {
   try {
     const { filter, selectedInputPage, pageSize } = action.payload;
-    const filterValue = `*${filter.value}*`.replace("**", "*");
+    const wildcardsUsed = filter.value.includes("*");
+    const regex = / +|, +/g;
+    let filterValue = filter.value.replace(regex, ",");
+    const regexStrip = /(^,+)|(,+$)/g;
+    filterValue = filterValue.replace(regexStrip, "");
+    filterValue = wildcardsUsed ? filterValue : `*${filterValue}*`.replace(/,/g, "*,*");
     // page is displayed in UI starting from 1 but api starts from 0
     const pageIndex = selectedInputPage - 1;
     const response = yield call(composer.listModules, filterValue, pageIndex, pageSize);


core/sagas/inputs.js
```diff
@@ -41,7 +41,9 @@ function flattenInputs(response) {
 function* fetchInputs(action) {
   try {
     const { filter, selectedInputPage, pageSize } = action.payload;
-    const filterValue = `*${filter.value}*`.replace("**", "*");
+    const wildcardsUsed = filter.value.includes("*");
+    let filterValue = filter.value.replace(/, /g, ",").replace(/ /g, ",");
```
jkozol, line 45: I believe you can use the regex or operator `|` here to create only one regex pattern.

```const regex = / |, /g```

We might also want to specify that these replacements should come between words with `\b`. So we don't add a comma at the beginning.

```const regex = /\b \b|\b, \b/g```

Also it might be a good idea to replace multiple strings using so `a,  b` doesn't get converted to `*a*,**,*b*`.

```const regex = /\b +\b|\b, +\b/g```

Comment (jgiardino) [reply core/sagas/inputs.js:45]: Thanks for the suggestions! 

If I use `\b` and the filter value includes wildcards, like `yell* pink *purple*`, then these separate values do not meet the criteria for `\b` and therefore the spaces are not replaced.

For the case where the user starts or ends with a space, the only alternative I can think of is to look for and remove any comma we inserted from the beginning or end of the string. Do you have any alternate suggestions?

Comment (jgiardino) [reply core/sagas/inputs.js]: I included updates using the first `const regex = / |, /g` example you suggested along with ` +` in one commit, and then an example of what I suggested above for removing commas from the start or end in the next commit. Let me know if you have any other suggestions I should try instead.

File: core/sagas/inputs.js [hunk 2]
```diff
@@ -41,7 +41,16 @@ function flattenInputs(response) {
...
-    const filterValue = `*${filter.value}*`.replace("**", "*");
+    const wildcardsUsed = filter.value.includes("*");
+    const regex = / +|, +/g;
+    let filterValue = filter.value.replace(regex, ",");
+    if (filterValue.charAt(0) === ",") {
+      filterValue = filterValue.slice(1);
+    }
+    if (filterValue.charAt(filterValue.length - 1) === ",") {
+      filterValue = filterValue.slice(0, filterValue.length - 1);
+    }
```
jkozol, line 52: I would replace this with a regex to remove multiple leading and following commas. A user could enter ` ,dog` which would be replaced to `,,dog` and then stripped to `,dog` and then replaced to  `**,*dog*`. Maybe the input box should be restricted here to alert a user they can't enter leading commas. The regex I would use to strip the filter is 

`stripRegex = /(^,+)|(,+$)/g`

jkozol, line 52:
```suggestion
   const regexStrip = /(^,+)|(,+$)/g;
   filterValue = filterValue.replace(regexStrip, "");
```

jgiardino: 👍 This suggestion is amazing! Thanks for your help.

jgiardino: I committed your suggestion, tested it, and squashed all commits.

PR #1059, opened 2020-07-22
Fix upload type capitalization

### LINKED ISSUE #959: Upload type should be capitalized
image

"Upload type **azure**" should instead be "Upload type **Azure**"

jkozol:
The image and upload list now map from image/provider type to a display label. This conversion mimics the labels used in the CreateImageUpload wizard.

Closes #959

### DIFF
--- a/components/ListView/ListItemUploads.js
+++ b/components/ListView/ListItemUploads.js
@@ -46,6 +46,15 @@ const messages = defineMessages({
 class ListItemUploads extends React.PureComponent {
   constructor(props) {
     super(props);
+    this.uploadProviderToLabel = this.uploadProviderToLabel.bind(this);
+  }
+
+  uploadProviderToLabel(uploadType) {
+    const uploadTypeLabels = {
+      aws: "AWS",
+      azure: "Azure",
+    };
+    return uploadTypeLabels[uploadType] !== undefined ? uploadTypeLabels[uploadType] : uploadType;
   }
 
   render() {
@@ -122,7 +131,7 @@ class ListItemUploads extends React.PureComponent {
                 )}
                 <div className="pf-l-flex__item">
                   <span>{formatMessage(messages.uploadType)}</span>{" "}
-                  <strong id={`${upload.uuid}-type`}>{upload.provider_name}</strong>
+                  <strong id={`${upload.uuid}-type`}>{this.uploadProviderToLabel(upload.provider_name)}</strong>
                 </div>
                 <div className="pf-l-flex__item">
                   <span>{formatMessage(messages.timeStarted)}</span> <strong>{formattedTime}</strong>
--- a/core/sagas/composes.js
+++ b/core/sagas/composes.js
@@ -115,8 +115,25 @@ function* fetchQueue() {
 
 function* fetchComposeTypes() {
   try {
-    const response = yield call(composer.getComposeTypes);
-    yield put(fetchingComposeTypesSucceeded(response));
+    const imageTypes = yield call(composer.getComposeTypes);
+    const imageTypeLabels = {
+      alibaba: "Alibaba Cloud (.qcow2)",
+      ami: "Amazon Web Services (.raw)",
+      "fedora-iot-commit": "Fedora IoT Commit (.tar)",
+      google: "Google Cloud Platform (.vhd)",
+      "hyper-v": "Hyper-V (.vhd)",
+      "live-iso": "Installer, suitable for USB and DVD (.iso)",
+      openstack: "OpenStack (.qcow2)",
+      "partitioned-disk": "Disk Image (.img)",
+      qcow2: "QEMU Image (.qcow2)",
+      "rhel-edge-commit": "RHEL for Edge Commit (.tar)",
+      vhd: "Microsoft Azure (.vhd)",
+      vmdk: "VMWare VSphere (.vmdk)",
+    };
+    const imageTypesLabelled = imageTypes.map((type) => {
+      return { ...type, label: imageTypeLabels[type.name] || type.name };
+    });
+    yield put(fetchingComposeTypesSucceeded(imageTypesLabelled));
   } catch (error) {
     console.log("Error in loadImageTypesSaga");
   }
--- a/pages/blueprint/index.js
+++ b/pages/blueprint/index.js
@@ -36,7 +36,7 @@ import {
   setSelectedInputParent,
   fetchingDepDetails,
 } from "../../core/actions/inputs";
-import { fetchingComposes } from "../../core/actions/composes";
+import { fetchingComposes, fetchingComposeTypes } from "../../core/actions/composes";
 import {
   setModalUserAccountVisible,
   setModalUserAccountData,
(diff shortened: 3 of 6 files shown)


pages/blueprint/index.js
```diff
@@ -278,6 +280,34 @@ class BlueprintPage extends React.Component {
...
+      "rhel-edge-commit": "RHEL for Edge Commit (.tar)",
+      tar: "TAR Archive (.tar)",
+      vhd: "Azure Disk Image (.vhd)",
+      vmdk: "VMware Virtual Machine Disk (.vmdk)",
+      alibaba: "Alibaba Machine Image (.qcow2)",
+      google: "Google Compute Engine Image (.tar.gz)",
+      "hyper-v": "Hyper-V Virtual Machine Disk (.vhdx)",
+    };
+    return imageTypeLabels[imageType];
+  }
```
jgiardino, line 301: We currently have a list of image types and their labels in [composer.js](https://github.com/osbuild/cockpit-composer/blob/master/core/composer.js#L132). Would it be possible to have this list of labels defined in one place that is referenced by every page/component that needs it?

Also, in composer.js, there is a conditional to handle the case where a new type has been added to osbuild-composer but hasn't been added to this list yet. If we do keep separate lists for defining labels, it might be good to add something like that here.

pages/blueprint/index.js
```diff
@@ -278,6 +280,34 @@ class BlueprintPage extends React.Component {
     return `/cockpit/channel/${cockpit.transport.csrf_token}?${query}`;
   }
 
+  imageTypeToLabel(imageType) {
+    const imageTypeLabels = {
+      ami: "Amazon Machine Image Disk (.ami)",
```
ondrejbudai, line 285: The issue of `.ami` is that the file extension was actually made up by us. AFAIK it isn't mentioned anywhere in AWS docs and also the produced image doesn't have this extension - we use an extension that represents the actual file's format. In the past, it was `.vhdx`, now it's `.raw` and I hope we will switch it to `.vmdk` soon.

components/ListView/ListItemUploads.js
```diff
@@ -46,6 +46,15 @@ const messages = defineMessages({
...
     super(props);
+    this.uploadProviderToLabel = this.uploadProviderToLabel.bind(this);
+  }
+
+  uploadProviderToLabel(uploadType) {
+    const uploadTypeLabels = {
+      aws: "AWS",
+      azure: "Azure",
+    };
+    return uploadTypeLabels[uploadType];
```
jgiardino, line 57: This will throw an error if a new upload type is ever introduced in the API that isn't listed here. Could we future-proof this a bit, and just return the raw `uploadType` value if a label isn't defined for that upload type? We have something similar currently for the compose types.

pages/blueprint/index.js
jgiardino: @jkozol What are your thoughts on updating this in this PR?

jkozol: @jgiardino I am fine including this in this PR. Do you have any thoughts on what this image type should be named? "Amazon Raw Image"?

components/ListView/ListItemUploads.js
jkozol: This is fixed. If uploadTypeLabels[uploadType] is undefined I will return the provided uploadType.

pages/blueprint/index.js
jgiardino: Oooh, thanks for asking :-D

Below is a screencap of what we used in the usability prototype. This was something I reviewed with @andreasn, and the consensus was that the labels should reflect the name of the platform or service provider. So for AWS, we replaced "Amazon Machine Image Disk" with "Amazon Web Services". It would be awesome if these labels could be updated based on what's below.

image

Providing the correct file type is tricky though. Ideally, this is provided by the API, since the API is defining what file type to generate. And since it sounds like it is likely to change, it would likely be incorrect at some point in time (i.e. now) if defined in the UI. I expect many users would want to know what file type they would get when selecting an image type before kicking off a process that could take some time (and some users would probably want this info  when researching the tool to determine if it's something they would want to use). I understand the argument that the file type isn't important in the context of the complete image upload workflow for cloud images, but that's only a portion of the workflows supported. So I'm on the fence about this one. Does anyone have thoughts on whether we should pull them or keep and maintain them as they change?

jkozol: I updated the image labels to these without the section headers. I can do that in a seperate PR if you would like. I also think we should maybe pass more info from the api's /compose/types route in order to get the both the image type and image label (i.e. Azure vs. .vhd) since there are multiple providers for the same image type (i.e. qcow2).

jgiardino: 👍 

> I updated the image labels to these without the section headers. I can do that in a seperate PR if you would like.

I think this is another one of those things we need from the API (cc'ing @teg on this), so I wouldn't add any groupings until that's provided.

jkozol: @jgiardino ok. I believe that the groupings are outside the scope of this PR since the goal here was just to standard the image list and upload list with the labels used in the wizard. Is it okay if we move this to an RFE issue?

jgiardino: Yep, that's what I was thinking, and what I meant to suggest. 🙃

core/sagas/composes.js
```diff
@@ -115,8 +115,25 @@ function* fetchQueue() {
...
+      ami: "Amazon Web Services (.raw)",
+      "fedora-iot-commit": "Fedora IoT Commit (.tar)",
+      google: "Google Cloud Platform (.vhd)",
+      "hyper-v": "Hyper-V (.vhd)",
+      "live-iso": "Installer, suitable for USB and DVD (.iso)",
+      openstack: "OpenStack (.qcow2)",
+      "partitioned-disk": "Disk Image (.img)",
+      qcow2: "QEMU Image (.qcow2)",
+      "rhel-edge-commit": "RHEL for Edge Commit (.tar)",
+      vhd: "Azure Disk Image (.vhd)",
```
jgiardino, line 130: Can you change "Azure Disk Image" to "Microsoft Azure". The others look good!

jgiardino:
These updates look great! I noted one minor thing inline, and then there's another item I noticed which may be out of scope of this PR but wanted to go ahead and mention here...

In following image, the step on the left is "Upload to AWS" and then in the Review screen it's labeled "Upload to Amazon". With this PR, we now have a long description and short description for AWS: "Amazon Web Services" vs "AWS", and I think the way we are currently using those is good. There are cases where we use just "Amazon", but that's when referring to "S3" and is consistent with [how Amazon refers to S3 ](https://aws.amazon.com/s3/).  For the example below, I'd suggest replacing "Upload to Amazon" with "Upload to AWS", to be consistent with the label on the left, and also consistent to how we're handling labels in both places for Azure. 

But, I'm thinking this should be a separate issue, since the string we send to translation is "Upload to Amazon" vs "Upload to {CSP}", and we're past translation freeze.

image

jkozol:
@jgiardino Thank you for pointing this out. I added this fix in a separate commit. So it is easier to leave it out of any bug fixes we bring to rhel-8-3

core/sagas/composes.js
jkozol: This is fixed.

PR #1067, opened 2020-07-29
wizard: add help text to rhel edge ref

### LINKED ISSUE #1065: RHEL for Edge default Ref hint
When no reference is specified in the UI `rhel/8/$ARCH/edge` will be used, it might be nice to hint at that value in the UI (input box?) so that it is clear to the user what happens if nothing is filled in there.

jkozol:
The default ref for a rhel edge image is rhel/8/$ARCH/edge. The ref field now states this.

Fixes #1065

### DIFF
--- a/components/Wizard/ImageStep.js
+++ b/components/Wizard/ImageStep.js
@@ -146,13 +146,25 @@ class ImageStep extends React.PureComponent {
               </Button>
             </Popover>
           </div>
-          <TextInput
-            className="pf-c-form-control"
-            value={ostreeSettings.ref !== undefined ? ostreeSettings.ref : ""}
-            type="text"
-            id="ostree-ref-input"
-            onChange={setOstreeRef}
-          />
+          <div className="pf-c-form__horizontal-group">
+            <TextInput
+              className="pf-c-form-control"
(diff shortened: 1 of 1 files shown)


jgiardino:
These updates look great! 

I took a look at why the `<em>` text is bold, and I noticed that when applying `font-weight: bold;` to the first instance of an `<em>` element (shown below), that the text becomes less bold. Otherwise, it has the normal styles for bold applied to it that's applied to most everything else (i.e. `font-weight: 400` via the PF variable), but it's displaying as more bold.

image

I'm cc'ing @KKoukiou and @garrett on this one, in case this issue has come up in cockpit or if they have suggestions on how to fix this. I'm not able to reproduce this in the patternfly.org examples, so I don't think it's an issue there.

jkozol:
@jgiardino This is now rebased on top of the font fixes. Can you give it another review? Thanks!

gicmo:
Very cool! I guess it is not trivial to just fill in $arch?

jkozol:
> Very cool! I guess it is not trivial to just fill in $arch?

@gicmo There is no nice way to do this through cockpit composer currently.

PR #1070, opened 2020-07-31
lib/patternfly: fix red hat font weights

jkozol:
Rewritting prior message:

Font weights are now fetched from https://github.com/RedHatOfficial/RedHatFont/blob/master/webfonts/red-hat-font.css

Information on the Red Hat fonts is available in multiple locations such as google fonts. We now only follow the font weights and weighted names listed in the official repository.

This PR requires changes to cockpit first since we fetch our fonts from cockpit. This will remain a draft until then.


KKoukiou:
@jkozol @jgiardino please don't merge this as is - we need first to make the correct fixes in cockpit repo and the sync from there.

jkozol:
@jgiardino This is ready for review now :)

PR #1093, opened 2020-08-25
package.json: Update @patternfly/react-core package dependency

jkozol:
This commit updates the patternfly react dependency. Also, due to changes in @patternfly/react-core, the links in our wizard sidebar are now buttons. The tests are updated for these changes. This is a continuation of #1091 but with test fixes.


jkozol:
@henrywang Thanks for the fix. It's weird that this only started recently since we've been passing the ostree test for a few weeks. Do you know what changed?

henrywang:
@jkozol, I don't think my patch is the root cause, just make things clear and remove unstable `variables`.

I've tried `check-image` tests a lot of times. Still can't reproduce this issue. I got failed caused by building error (download error), but it's not related with this PR. I still think it's not a bug IMHO.

jkozol:
@henrywang I have not been able to replicate the final test failure locally. Any thoughts?

henrywang (approved):
I think we can land this PR and send another PR to make a double check.

PR #1099, opened 2020-09-03
Update translations from Fedora Weblate

cockpituous:
Update translations from Fedora Weblate
 * [x] po-refresh


jkozol (approved):
This looks good to me. I can't verify the actual translations but I can affirm that the strings marked for translation are correct.

PR #1100, opened 2020-09-03
wizard: refactor image step

### LINKED ISSUE #1011: Validate the `Ref` input for ostree based commits
Not all values are accepted for the `Ref` field. If something wrong is put in there `osbuild` will fail very late in the build with something like:
```
error: Invalid refspec /rhel8/gateway
```
ostree uses a regex to validate the ref in [`ostree_parse_refspec`](https://github.com/ostreedev/ostree/blob/73742252e286e8b53677555dc1b0d52d55fb7012/src/libostree/ostree-core.c#L172):

```
#define OSTREE_REF_FRAGMENT_REGEXP "[\\w\\d][-._\\w\\d]*"
#define OSTREE_REF_REGEXP "(?:" OSTREE_REF_FRAGMENT_REGEXP "/)*" OSTREE_REF_FRAGMENT_REGEXP
#define OSTREE_REMOTE_NAME_REGEXP OSTREE_REF_FRAGMENT_REGEXP

g_regex_new ("^(" OSTREE_REMOTE_NAME_REGEXP ":)?(" OSTREE_REF_REGEXP ")$", 0, 0, NULL);
```

Discovered by @mrguitar

jkozol:
The image step now primarily uses pf4 react components instead of html elements with pf4 styles applied. This includes several minor improvements.

Field validation is improved. The image size field's validation style has been improved to include success and error colors and icons. The ostree ref is now validated to make sure the format of the ref won't error during the build process.

Image type is now marked as a required field. The help icons are formatted to be in line with text. The image size field is not required for all image types so it is now only displayed after selecting image type.

All changes follow patternfly's [form style guide](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms)

Fixes #1011

### DIFF
--- a/components/Wizard/ImageStep.js
+++ b/components/Wizard/ImageStep.js
@@ -7,6 +7,8 @@ import {
   FormGroup,
   FormSelect,
   FormSelectOption,
+  InputGroup,
+  InputGroupText,
   Popover,
   Text,
   TextInput,
@@ -46,6 +48,14 @@ const messages = defineMessages({
   infotip: {
     defaultMessage: "This process can take a while. " + "Images are built in the order they are started.",
   },
+  ostreeRefHelperText: {
+    defaultMessage:
+      "Valid characters for ref are letters from a to z, the digits from 0 to 9, the hyphen (-), the underscore (_), the period (.), " +
+      "and the forward slash (/). A ref must start with a letter, a number, or an underscore. Slashes must also be followed by a letter or number.",
+  },
+  requiredField: {
+    defaultMessage: "This is a required field",
+  },
   type: {
     defaultMessage: "Type",
   },
@@ -65,7 +75,7 @@ const messages = defineMessages({
     defaultMessage: "Select one",
   },
   warningSizeSmall: {
-    defaultMessage: "Minimum size is {size} GB.",
+    defaultMessage: "Minimum size is {size} GB",
   },
   warningSizeLarge: {
     defaultMessage:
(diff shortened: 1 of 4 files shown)


jkozol:
@jgiardino Thank you for the review!

I added the proper ids to the inputs' aria labels so these should be fixed now.

In regards to the help text, it is a bit difficult to get multi line help text to play nicely in the patternfly react components. Also, the [forms design document](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms) recommends against multi line help text. I did make some updates so the text now shows as follows:

doubleHelp
singleHelp
errorHelp

components/Wizard/ImageStep.js
```diff
@@ -292,20 +295,141 @@ class ImageStep extends React.PureComponent {
...
+            type="text"
+            id="ostree-ref-input"
+            aria-describedby="ostree-ref-input-helper"
+            onChange={this.handleOSTreeRef}
+            validated={ostreeRefValidated}
           />
-        </div>
-      </div>
+          {ostreeRefValidated === "default" && imageType === "rhel-edge-commit" && (
+            <p className="pf-c-form__helper-text" id="ostree-ref-input-helper" aria-live="polite">
```
jgiardino, line 422: These updates look great regarding how it behaves and looks! 

There's just one minor issue regarding the `id` value defined for this `<p>` element. I included a screencap of what gets rendered, where you can see that both the `<div>` and `<p>` have the same `id` value, which isn't valid html. 

I'm guessing that the `<FormGroup>` component automatically generates the `id` for the `<div>` based on the value provided for `fieldID` a few lines above, so the `<p>` element will need a new id. And you can still reference this text in `aria-describedby` because this attribute can list more than one `id`, separated by spaces. 

So you can fix this by making the following updates:

- Provide a different value for this `<p>` element, e.g. `id="ostree-ref-input-helper-default"`
- For the `<TextInput>` component, reference both elements with help text, e.g. `aria-describedby="ostree-ref-input-helper-default ostree-ref-input-helper"`. 

image

jkozol: Thanks for pointing this out. Should be fixed now.

components/Wizard/ImageStep.js
```diff
@@ -292,20 +291,141 @@ class ImageStep extends React.PureComponent {
...
+          helperTextInvalid={formatMessage(messages.ostreeRefHelperText)}
+          validated={ostreeRefValidated}
+          hasNoPaddingTop
+        >
+          <TextInput
+            className="pf-c-form-control"
+            value={ostreeSettings.ref !== undefined ? ostreeSettings.ref : ""}
+            type="text"
+            id="ostree-ref-input"
+            aria-describedby="ostree-ref-input-helper ostree-ref-input-helper-default"
```
jgiardino, line 413: Soooo close! The `id`s look good. But ideally, they would be listed in the value of `aria-describedby` in the same order that they display in the UI/DOM. Below is a screencap of the order they appear in the DOM. So here you should have this instead:
  `aria-describedby="ostree-ref-input-helper-default ostree-ref-input-helper"`

The reason for this is because screen reader users can be sighted users. Probably the majority of screen reader users are blind, but there can be other reasons why a user uses a screen reader for assistance, such as someone with dyslexia or low vision. So if the screen reader is announcing contents that don't match what displays on-screen, it can be very confusing. 

image

jkozol: ahhh I see. I set it to `ostree-ref-input-helper-default ostree-ref-input-helper` now.

jgiardino (dismissed):
LGTM! Thanks!

jkozol:
@henrywang could you take a look at the fedora-32/firefox test? It consistently refuses to pass but I can't figure out why.

henrywang:
> @henrywang could you take a look at the fedora-32/firefox test? It consistently refuses to pass but I can't figure out why.

Working on that.

jkozol:
I have rebased this on top of master

components/Wizard/ImageStep.js
```diff
@@ -292,20 +291,141 @@ class ImageStep extends React.PureComponent {
...
+              aria-label={formatMessage(ariaLabels.ostreeRef)}
+            >
+              <Button variant="plain" aria-label={formatMessage(ariaLabels.ostreeRef)}>
+                <OutlinedQuestionCircleIcon className="cc-c-text__align-icon" id="popover-icon" />
+              </Button>
+            </Popover>
+          }
+          fieldId="ostree-ref-input"
+          helperText={formatMessage(messages.ostreeRefHelperText)}
+          helperTextInvalid={formatMessage(messages.ostreeRefHelperText)}
```
jgiardino, line 404: I'm not seeing this helper text display, and I'm not sure why. Are you seeing it in your build?

image

jgiardino: I'm deleting this comment, because I think I realized what the issue was on my end. 🙄

jgiardino (approved):
Everything still looks good! 😄

jkozol:
Merging with the failed f33 test because the screenshot shows that the image built correctly.

PR #1103, opened 2020-09-09
utils: remove fedora 33 cockpituous release

jkozol:
Fedora 33 is not currently supported by osbuild-composer so cockpit-composer should not release into fedora 33 until support is
added.


marusak (approved):
Right, "don't release if you don't test", agree with that :)

PR #1110, opened 2020-09-21
pages: fix blueprints page error state

jkozol:
When a user enters the blueprints page and osbuild-composer is not running they are prompted to start the socket. This same error state displays if the user is not root but the button to start the socket is disabled and alerts the user they don't have permissions. The check-service test is updated to accept the log messages associated with trying to access the api without root privileges.

### DIFF
--- a/pages/blueprints/index.js
+++ b/pages/blueprints/index.js
@@ -95,8 +95,11 @@ class BlueprintsPage extends React.Component {
           manageSources={manageSources}
         />
         {(blueprintsLoading === true && <Loading />) ||
+          (blueprintsError === undefined && (
+            <EmptyStateInactive fetchingBlueprints={this.props.fetchingBlueprints} />
+          )) ||
           (blueprintsError !== null &&
-            ((blueprintsError.message === "not-found" && (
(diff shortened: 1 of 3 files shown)


pages/blueprints/index.js
```diff
@@ -95,8 +95,11 @@ class BlueprintsPage extends React.Component {
           manageSources={manageSources}
         />
         {(blueprintsLoading === true && <Loading />) ||
+          (blueprintsError === undefined && (
+            <EmptyStateInactive fetchingBlueprints={this.props.fetchingBlueprints} />
+          )) ||
           (blueprintsError !== null &&
```
jgiardino, line 101: Just asking, but if the default state is changed from `null` to `undefined` in `store.js`, then should this line also be changed from `null` to `undefined`?

jgiardino (commented):
I reviewed the UI for a non-root user and this looks good. Thanks for updating!

jgiardino (approved):
Clicked the button too fast in my previous comment :-)

henrywang (approved):
Thanks. LGTM.

PR #1111, opened 2020-09-24
Update translations from Fedora Weblate

cockpituous:
Update translations from Fedora Weblate
 * [x] po-refresh


jkozol (approved):
I reviewed the translated strings and they are marked as expected.

PR #1112, opened 2020-09-25
README.md: Grammar fix

martinpitt:
This is mostly an excuse to trigger tests on the current rhel-8-3 image, to compare with the refresh for https://github.com/cockpit-project/bots/pull/1256


martinpitt:
OK, this obviously works, so it does depend on the base image. Feel free to land this anyway though, as I believe it still is a legit grammar fix.

larskarlitski (approved):
Lovely. Thank you!

PR #1115, opened 2020-09-28
test: Rename rhel-8-beta.json to rhel-8.json to work with osbuild-composer

henrywang:
New rhel-8-3 image needs name repo replacement file to `rhel-8.json` instead of `rhel-8-beta.json`.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -12,7 +12,7 @@ lvextend -r -l +100%FREE $VG/root
 
 # overriding osbuild-composer repo with nightly
 mkdir -p /etc/osbuild-composer/repositories
-cp /home/admin/files/rhel-8-beta.json /etc/osbuild-composer/repositories/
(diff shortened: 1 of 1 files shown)


test/vm.install
```diff
@@ -12,7 +12,7 @@ lvextend -r -l +100%FREE $VG/root
 
 # overriding osbuild-composer repo with nightly
 mkdir -p /etc/osbuild-composer/repositories
-cp /home/admin/files/rhel-8-beta.json /etc/osbuild-composer/repositories/
+cp /home/admin/files/rhel-8-beta.json /etc/osbuild-composer/repositories/rhel-8.json
```
martinpitt, line 15: Thanks! I'll land this now to unblock testing -- but should it be renamed in the source as well?

martinpitt (approved):

test/vm.install
martinpitt: Or perhaps, copy it as both names, so that it still continues to work once we move to 8.4 alpha/beta?

PR #1116, opened 2020-09-28
better error message

Gundersanne:
osbuild-composer often specifies more detailed errors in the body of the response, so also log that, and try to parse it as json.

Screenshot from 2020-09-28 10-50-51

---

the bash arithmetic commit was a requirement to get cockpit-composer to make rpms on debian, also `--nodeps` for rpmbuild was required but it wouldn't be a good idea to commit that :)

### DIFF
--- a/core/composer.js
+++ b/core/composer.js
@@ -36,13 +36,24 @@ function request(options) {
         else if (replyFormat === "raw") resolve(data);
         else throw new Error(`invalid replyFormat: '${replyFormat}'`);
       })
-      .catch((error) =>
+      .catch((error, data) => {
+        /*
+         * osbuild-composer describes the error in more detail, so add the body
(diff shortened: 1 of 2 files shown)


jkozol (approved):
Looks good to me! Thanks! Also, just curious, why is `bash` needed over `sh` for debian?

Gundersanne:
@jkozol on debian `dash` is the default shell (the one which gets invoked when you call `sh`), which doesn't have support for bash-style arithmetic. On fedora bash is the default, so bash arithmetic works there when you invoke with sh.

In general I think it's better when using bash-specific syntax to explicitly use bash as an interpreter.

Gundersanne:
oh @jkozol it seems I can't actually press the merge button, not authorized >.> So I'll leave it up to you :p

PR #1121, opened 2020-10-06
utils: add fedora 33 to cockpituous release

jkozol:
Since we stopped releasing into fedora 33, we caused a downgrade issue between fedora 32 (cockpit composer version 24 available) and fedora 33 (version 23 available). We will now release into 33 alongside 31, 32, and rawhide.

### DIFF
--- a/utils/cockpituous-release
+++ b/utils/cockpituous-release
@@ -19,6 +19,7 @@ job release-srpm -V
 job release-koji master
 job release-koji f31
 job release-koji f32
+job release-koji f33 
 
 job release-github
 job release-copr @osbuild/cockpit-composer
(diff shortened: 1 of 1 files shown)


larskarlitski (approved):
Thanks!

marusak:
https://github.com/cockpit-project/bots/pull/1213 I believe that it should be tested as well ;)

jkozol:
@marusak Thanks for bringing this up! fedora 33 support has been added to osbuild-composer's master branch but has not yet been released so the tests will still fail. Once it is released we can test against f33. We decided to release cockpit-composer into f33 because we had already released version 23 into f33 and upgrading from f32 to f33 would cause a package downgrade. We figured that, since what was already in f33 was broken, releasing another broken version wouldn't be an issue and would resolve the downgrade issue.

PR #1131, opened 2020-10-20
Run po-refresh and npm-update in GitHub workflows

martinpitt:
This avoids a lot of moving parts in our infrastructure (webhook,
npm-trigger roundtrip, tasks container) and does not need *any* secret
other than the automatically provided GitHub token.

 - [x] Wait until https://github.com/cockpit-project/cockpit-podman/pull/568 lands and successfully runs


martinpitt:
This is the same as https://github.com/cockpit-project/starter-kit/pull/384 (which is confirmed working) and https://github.com/cockpit-project/cockpit-podman/pull/568 (which didn't land yet) -- hence blocked, but this is an RFC if you are ok with that change.

larskarlitski:
I like this a lot! Thanks Martin.

I think this is missing the workflow file itself, as it is included in https://github.com/cockpit-project/cockpit-podman/pull/568.

martinpitt:
@larskarlitski Oh no, I won't tell you that, it's of course a secret! *blush* (Fixed..)

It still needed a fix anyway, see https://github.com/cockpit-project/cockpit-podman/pull/569 . Adjusted the workflow to use the cockpituous token instead of the workflow default one, and I uploaded that token to https://github.com/osbuild/cockpit-composer/settings/secrets .

martinpitt:
I have this working now, see the [recent run for cockpit-podman](https://github.com/cockpit-project/cockpit-podman/runs/1279898608). So good to review/land!

larskarlitski (approved):
> I have this working now, see the recent run for cockpit-podman. So good to review/land!

Really cool stuff! Thanks again Martin :heart:

@jkozol all good from your side as well?

jkozol (approved):
Thanks! This looks good to me.

PR #1143, opened 2020-10-28
components: remove shortid

jkozol:
The short id package is only used for a key value and is not needed. Instead the component name is now used for the key. The component names will always be unique so they will work as a key.


jgiardino (dismissed):
Looks good to me!

henrywang (approved):

PR #1155, opened 2020-11-03
test: fix Test Source

jkozol:
The test can't find the Add Source button on firefox. This may have to do with multiple buttons having the same text.

### DIFF
--- a/test/verify/check-source
+++ b/test/verify/check-source
@@ -57,6 +57,7 @@ class TestSource(composerlib.ComposerCase):
         b.set_val(dropdown_menu_sel, value_id)
         b.wait_val(dropdown_menu_sel, value_id)
         # groups action done
+        b.wait_present("button:contains('Add Source')")
         b.click("button:contains('Add Source')")
 
         # HACK: workaround issue https://github.com/osbuild/cockpit-composer/issues/989
@@ -64,6 +65,8 @@ class TestSource(composerlib.ComposerCase):
         if b.cdp.browser == "firefox":
             # open manage sources dialog
             drop_down_sel = ".toolbar-pf-action-right #dropdownKebab"
+            b.wait_present("#main")
+            b.wait_attr(drop_down_sel, "aria-expanded", "false")
             b.click(drop_down_sel)
             b.wait_attr(drop_down_sel, "aria-expanded", "true")
             b.click("a:contains('Manage Sources')")


henrywang:
@jkozol, It's not caused by code, but new version of firefox. But I spent a lot of time on this issue, but can't find how to fix it.

marusak:
Here are some ideas how to debug this:
- If you cannot debug locally it is completely fine to do that in CI, just to make the iteration faster I suggest:
    - Add 'no-test' label and always just trigger `fedora-32/firefox` it saves both your time and CI resources
    - Disable all other tests (just remove them or rename them so they are not discovered by test-run)
    - Run with `-vt` to get better logs (logs become huge, but if you disable all other tests and run only one context it is fine)

- It fails on `wait_js_cond(ph_is_present("button:contains('Add Source'):not([disabled])"))` so I think it make sense to check:
    - `wait_present("button:contains('Add Source')"`
    - `wait_visible("button:contains('Add Source')")`
    - `wait_present("button:contains('Add Source'):not([disabled])"`
     - `wait_visible("button:contains('Add Source'):not([disabled])"`

    in this order. My guess is that when you change the dropdown the event is not propagated/recognized and the button will remain disabled.

- Check that the modal is still opened. We had some issue that modals sometimes get closed randomly. (There is fix for that, but I don't know it from top of my head).

- Try to put debug info into code. `console.log` when inputs and the dropdown change. log what value you have received. log when you enable the button. log when the modal gets closed.

-  Try replacing `b.set_input_text("#textInput1-modal-source", repo_name)` and `b.set_input_text("#textInput2-modal-source", repo_url)` with `b.set_val(sel, text)`. Check `test/common/testlib.py` for the difference. Also you may try `b.blur(sel)` to trigger event.

- This is just to make the test cleaner, but you may want to replace
```
        # group actions for select from dropdown menu                           
        dropdown_menu_sel = "#textInput3-modal-source"                          
        b.wait_visible(dropdown_menu_sel)                                       
        test_selector = "{} option[value='yum-baseurl']".format(dropdown_menu_sel)
        b.wait_present(test_selector)                                           
        value_id = b.attr(test_selector, "value")                               
        b.set_val(dropdown_menu_sel, value_id)                                  
        b.wait_val(dropdown_menu_sel, value_id) 
```
    with just one line `select_from_dropdown(...)`

- You can try to use `set_val` instead of `select_from_dropdown` (see the previous point). But if you use bootstrap dropdowns that would not work. Maybe time to migrate :) (especially if the issue is related to events not coming in).

jkozol:
@marusak Thanks for all of this advice.

jkozol:
This is dependent on #1162

henrywang (approved):
LGTM. @jkozol, Thanks!

PR #1162, opened 2020-11-09
eslint: fix lint issues with function order

jkozol:
After cleaning my node_modules folder cockpit-composer began complaining about the eslint issues fixed in this PR.


jkozol:
This issue seems isolated to my local env

PR #1163, opened 2020-11-11
Fix eslint loader issues

jkozol:
`eslint-loader` is replaced with the `eslint-webpack-plugin` and the order of functions is updated.


jkozol:
@henrywang This only passes the travis eslint test when the node_modules are cached. This passes `make eslint` locally for me. Do you have any thoughts on fixing this?

henrywang:
@jkozol, I think we can clear travis cache to fix this issue by following [this guide](https://docs.travis-ci.com/user/caching/#clearing-caches). But I have no permission on Travis to clear cache. Could you please help to do that or do you know who have permission to do that? Thanks.

henrywang (approved):
LGTM. Thanks!

PR #1170, opened 2020-11-19
package.json: update @patternfly/react-core

jkozol:
@patternfly/react-core 4.75.2 broke our builds due to a missing style for the Splitter component. Upgrading to 4.79.2 fixes the error.


Gundersanne (approved):
Looks good! Wait until ti's green though, and maybe worth clicking through the page a bit, because this seems like a rather large version jump.

PR #1183, opened 2020-12-04
test: Stop both osbuild-composer.socket and service because systemd refuses to start the socket when the service is still running

henrywang:


jkozol (approved):
Looks good to me.

PR #1188, opened 2020-12-09
test: fix TestSource modal reload issue

jkozol:
In the firefox nightlies the add source modal will drop after adding a new source. In order to continue the tests we need to reopen it. We now wait until the modal has closed before we reopen it.


henrywang (approved):
Looks good to me. Thanks!

PR #1189, opened 2020-12-09
test: update README

jkozol:
With the incoming support for rhel-8-4 the testing README now contains the test distros we actively test on.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -55,7 +55,7 @@ interactively follow what a test is doing, set environment variable `$TEST_SHOW_
 You can set these environment variables to configure the test suite:
 
     TEST_OS    The OS to run the tests in.  Currently supported values:
-                  "fedora-32"
+                  "fedora-32, fedora-33, rhel-8-4"
                "fedora-32" is the default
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -55,7 +55,7 @@ interactively follow what a test is doing, set environment variable `$TEST_SHOW_
 You can set these environment variables to configure the test suite:
 
     TEST_OS    The OS to run the tests in.  Currently supported values:
-                  "fedora-32"
+                  "fedora-32, fedora-33, rhel-8-4"
```
henrywang, line 58: Should we put `rhel-8-3` here?
@martinpitt, do we still keep `rhel-8-3` image in our CI? Thanks.

martinpitt: At the moment, yes. We'll keep the image as long as any project still wants to use it. IMHO it's also convenient for manual debugging purposes to have a RHEL 8.3 image readily at hand.

henrywang: Make sense to me. Thanks!

PR #1191, opened 2020-12-10
test: fix call to stop osbuild-local-worker.socket

jkozol:
The test setup should only stop the osbuild-local-worker socket if the distro is fedora 32 or 33.


henrywang (approved):
Looks good to me. Thanks.

PR #1194, opened 2020-12-11
tests: Always check if elements are visible

marusak:
`wait_present` should be used just rarely as the meaning of this
function is "Wait until present in dom", while meaning of `wait_visible`
is "Wait until in dom and visible to user". Tests should really just
check what user sees. Using `wait_present` incorrectly can lead to
broken code (just now I was fixing accordion which would never open up,
but tests were happy as they were just checking with `wait_present`)

I haven't tried this, just `sed` it. Lets see what CI think :)

### DIFF
--- a/test/verify/check-blueprint
+++ b/test/verify/check-blueprint
@@ -18,11 +18,11 @@ class TestBlueprint(composerlib.ComposerCase):
         b = self.browser
 
         self.login_and_go("/composer", superuser=True)
-        b.wait_present("#main")
+        b.wait_visible("#main")
 
         # create blueprint
         b.click(".toolbar-pf-action-right #cmpsr-btn-crt-blueprint")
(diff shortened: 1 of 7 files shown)


marusak:
10 failing tests, seems that mostly on incorrect using of selects (not opening the select, just interacting with it). We have functions to do this properly in `test/common/testlib.py` so I'll take this as an opportunity to rewrite it into that :)

marusak:
List of image types is fetched asynchronously. UI does not indicate this in any way, which is not great. But I think we can just simply check for it in tests.

marusak:
oh, there is still one broken test, that needs fixing. Someone retried tests just now, so I don't have those runs available. So I'll wait until it finishes and then fix those remaining tests.

henrywang (approved):
Nice! Thanks @marusak!

PR #1195, opened 2020-12-11
Use sentence case rather than title case

marusak:
See https://www.patternfly.org/v4/ux-writing/capitalization

19% of strings were using title case.
This breaks translations, but they will be just fuzzy, so very easy to
fix up by translators. It would be possible to also change `.po` files,
but then we would need to manually fix Weblate. Also I believe that this
is not that big number of strings and gives translators chance to go
through and fix strings properly (as I've seen people translating title
case as title case).

This was done by:
```
make po/cockpit-composer.pot
msggrep --no-wrap --msgid -e '[A-Z].*\s[A-Z]' po/cockpit-composer.pot | grep "^msgid " | cut -d " " -f2- | sort --reverse > titles
<manually reviewing strings in titles and removing the ones, that don't apply>
<copying "tools/title2sentence.py" from cockpit-project/cockpit>
./title2sentence.py -i titles -o f.sh
<changing the first line of f.sh to "git ls-files | xargs sed -i \ ">
chmod u+x f.sh
./f.sh
git checkout -- po/*.po
git add -u
```

### DIFF
--- a/components/EmptyState/EmptyStateInactive.js
+++ b/components/EmptyState/EmptyStateInactive.js
@@ -9,7 +9,7 @@ import EmptyState from "./EmptyState";
 
 const messages = defineMessages({
   errorInactiveTitle: {
-    defaultMessage: "Image Building Service is Not Active",
+    defaultMessage: "Image building service is not active",
   },
   errorInactivePrimary: {
     defaultMessage: "Start",
--- a/components/ListView/BlueprintContents.js
+++ b/components/ListView/BlueprintContents.js
@@ -17,16 +17,16 @@ const messages = defineMessages({
     defaultMessage: "An error occurred while trying to get blueprint contents.",
   },
   emptyStateErrorTitle: {
-    defaultMessage: "An Error Occurred",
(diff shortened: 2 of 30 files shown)


marusak:
I should have mentioned that this is just proposal from my side :) Of course it is up to the team what they decide. @jkozol and @jgiardino WDYT?

jkozol (commented):
This change makes sense to me. I believe patternfly also recommends using sentence case. Thanks for making these changes. I'll wait for @jgiardino's thoughts before I approve.

henrywang (dismissed):
Thanks @marusak! Wait for @jgiardino's approve.

jkozol:
@marusak since Jenn has now moved on from the project I am happy to merge this without her approval. Thanks for fixing all of these strings!!

henrywang:
@marusak, could you please make a rebase and I'll merge this PR. Thanks.

marusak:
Rebased

jkozol (approved):
LGTM! Thanks for this work!!

PR #1205, opened 2021-01-05
test: Fix vm.install for non-LVM cloud images

martinpitt:
Some of our images, like centos-8-stream, are already built from the official
cloud images instead of virt-install with LVM. More images are going to do that
soon [1], so fix vm.install to only do the LVM grow steps if the image actually
uses LVM.

[1] https://github.com/cockpit-project/bots/pull/1518

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -3,12 +3,14 @@
 # The application RPM will be installed separately
 set -eux
 
-# resize root partition to fill free space
-echo -en "n\n\n\n\n\nw\n" | fdisk /dev/vda
-pvcreate /dev/vda3
+# on virt-install images with LVM, resize root partition to fill free space
 VG=$(vgs --noheadings -o vg_name)
-vgextend $VG /dev/vda3
-lvextend -r -l +100%FREE $VG/root
+if [ -n "$VG" ]; then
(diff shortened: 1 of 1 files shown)


henrywang (approved):
Looks good to me. Thanks @martinpitt!

martinpitt:
That [testNavigate failure](https://logs.cockpit-project.org/logs/pull-1205-20210105-152305-fb473194-fedora-33/log.html#18-2) is unrelated. The test log shows that in this run it (rightfully) does run the `pvcreate` etc. steps, so nothing changed in the image.

martinpitt:
@henrywang: I don't understand that test. It is supposed to filter by release, and that is 1.fc33. So it's quite natural that both httpd and tmux appear. Maybe this was supposed to filter on the *version* instead? But that already happens in the test case above. So I believe that test has always been racy. Perhaps we have some test package which is not available in fc33? Then the test could use that. Or it just makes sure that both are shown. I am AFK tomorrow (holiday), but I can send a PR on Thursday unless you beat me to it.

In the meantime, I'm landing this PR as it's obviously an unrelated failure, so that the image build PR gets unblocked.

henrywang:
Thanks @martinpitt. I found the root cause. PR #1207 should fix this issue.

PR #1207, opened 2021-01-06
test: Fix check package release issue

henrywang:
Test should check release field when filter package by release, not package name


martinpitt (approved):
Ack, that makes more sense, thanks! The subsequent "httpd" check is still a bit confusing/redundant, but harmless.

henrywang:
Failed cases are not this PR related.

PR #1211, opened 2021-01-08
Update RHEL 8 branch to curent release

jkozol:
We have released version 27 into the RHEL nightlies. This branch is now updated to be up to date with what has been released with the addition of 2928f9781ad97db500b0b753a8432a6e3df94aa6, cf1d061ad0ab585fc242d00b35e76578c9ec834f, and ac37f4134a24ad8b2992e5c65eb2b2c01e6bef16. Since these commits are placed on top of version 27, this update to the branch can't be automatically merged. Is this best course of action to force push to the upstream branch when this is approved?


henrywang:
> We have released version 27 into the RHEL nightlies. This branch is now updated to be up to date with what has been released with the addition of [2928f97](https://github.com/osbuild/cockpit-composer/commit/2928f9781ad97db500b0b753a8432a6e3df94aa6), [cf1d061](https://github.com/osbuild/cockpit-composer/commit/cf1d061ad0ab585fc242d00b35e76578c9ec834f), and [ac37f41](https://github.com/osbuild/cockpit-composer/commit/ac37f4134a24ad8b2992e5c65eb2b2c01e6bef16). Since these commits are placed on top of version 27, this update to the branch can't be automatically merged. Is this best course of action to force push to the upstream branch when this is approved?

Yeah! I agree with you.

marusak:
Sorry, I am not properly following what is happening in c-composer, but I have a few questions.

What you have released into rhel, I assume, is current main branch, isn't it?
Is there already reason to branch out?
I would say the course in which one wants to go is to test the main branch with rhel-8-4 tests, then we know it works, thus when branching time comes, you can just `git push -f upstream main:rhel-8` and it will continue to work (maybe editing testmap just to trigger the right tests).

jkozol:
@marusak Hi! Thanks for commenting.

> What you have released into rhel, I assume, is current main branch, isn't it?

We have released the main branch up to release 26.

> Is there already reason to branch out?

So I am under the impression that we should update the rhel 8 branch whenever we release to rhel 8. When does cockpit branch out?

marusak:
> So I am under the impression that we should update the rhel 8 branch whenever we release to rhel 8.

I am not here to tell you how to do things, thats always team's decision. But I find this excessive and unnecessary. If you released tag X into rhel and you have this tag in main branch, then you have point of reference. If you want to release something that differs from the main branch, then you should branch out, to have point of reference.

> When does cockpit branch out?

Only when we have to - that means, when we cannot release master branch to rhel anymore. Usually this happens when all of the following conditions happen:
1. It is after string freeze
2. We have some commit that changes strings
3. We have something we want to get into rhel (some bug fix)

We cannot use master anymore, as we would break translations, but we need to pick just some commits. Therefore we branch out from the last tag we delivered from the master branch (or the one right before we changed string) and then cherry pick bug fixes (and usually some test and build fixes).

Note: This is not the usual rhel workflow. "Proper rhel maintainer" would stop updating it after dev freeze and then would really only cherry-pick bug fixes that have proper BZs :) We in Cockpit are really stretching this as much as we can, but we test heavily and therefore we are reasonably confident to deliver new changes and not break stuff.

jkozol:
@marusak Okay. Thank you for this explanation! I will close this then and update the branch only if we have the three conditions you listed met. That all makes sense to me.

PR #1217, opened 2021-01-20
wizard: add option to upload VMWare images

jmikovic:


components/Wizard/ImageStep.js
```diff
@@ -308,6 +311,41 @@ class ImageStep extends React.PureComponent {
...
+        label={<FormattedMessage defaultMessage="Upload image" />}
+        labelIcon={
+          <Popover
+            id="popover-help"
+            bodyContent={
+              <FormattedMessage
+                defaultMessage="
+                      Image Builder can can upload images you create to VMWare vSphere.
+                      When the image build is complete and the upload action is successful,
+                      the image file is available in the Cluster on vSphere instance that you specified.
```
jkozol, line 325: Minor nitpick. Could you say "in the Cluster on `the` vSphere instance that you specified"

jkozol: Also could you make the statement "VMWare vSphere" a link to the VMWare vSphere.

components/Wizard/ReviewStep.js
```diff
@@ -183,7 +186,7 @@ class ReviewStep extends React.PureComponent {
             bodyContent={
               <FormattedMessage
                 defaultMessage="
-                      Image Builder can upload images you create to a Blob container in {azure}. When the image build is complete 
+                      Image Builder can upload images you create to a Blob container in {azure}. When the image build is complete
```
jkozol, line 189: Can you update the aws test to not include the extra space too. If you have any questions about the test let me know!

components/Wizard/ReviewStep.js
```diff
@@ -234,6 +237,64 @@ class ReviewStep extends React.PureComponent {
...
+          </h3>
+          <Popover
+            className="pf-l-flex__item"
+            id="vmware-review-popover"
+            bodyContent={
+              <FormattedMessage
+                defaultMessage="
+                      Image Builder can can upload images you create to VMWare vSphere.
+                      When the image build is complete and the upload action is successful,
+                      the image file is available in the Cluster on vSphere instance that you specified.
```
jkozol, line 254: Same comments as for the other help text

> Minor nitpick. Could you say "in the Cluster on the vSphere instance that you specified"
> Also could you make the statement "VMWare vSphere" a link to the VMWare vSphere.

components/Wizard/VMWareDestinationStep.js
```diff
@@ -0,0 +1,218 @@
...
+                </span>
+                <span className="pf-c-form__label-required" aria-hidden="true">
+                  &#42;
+                </span>
+              </label>
+              <Popover
+                id="bucket-popover"
+                bodyContent={
+                  <FormattedMessage
+                    defaultMessage="Provide url of your {vmware} instance to which the image file will be uploaded."
```
jkozol, line 82: Minor nitpick. "Provide `the` url" .

jkozol (changes requested):
Really great work here! I just have a couple of changes but overall it looks good!!

Also, we need a test for it. I'd be happy to talk with you about the tests so you can add one. Should be very similar to the aws and azure image tests.

Thanks!

components/Wizard/ImageStep.js
jmikovic: I was thinking about that link but AFAIK vSphere doesn't have a public instance similar to AWS or Azure, it's a product you privately so I didn't see much point in it, but I can link it to https://www.vmware.com/products/vsphere.html if you think it makes sense.

jkozol: Mmm. Okay. I wasn't aware of that. I think you're right that there isn't much of a point to link to the product page. Thanks for the clarification.

jkozol:
Thanks for the updates. Looks good to me but the tests are failing to build. I'm looking into this. Seems unrelated to your PR.

jkozol (approved):
Great work here. Once the tests pass this can merge. Thanks!!

jkozol:
The failing test is unrelated to this PR. Merging.

PR #1220, opened 2021-01-24
[rhel-8] test: Fix check package release issue

martinpitt:
Test should check release field when filter package by release, not
package name

Cherry-picked from master commit 7b416df4dd06

-----

 - [ ] Needs some more backports: PR #1221 

This is a nice and unintrusive test stabilization for the RHEL 8 branch. But I mostly want a PR for running tests against rhel-8-4 now, as deadlines for 8.3.1 have passed. So let's switch over to 8.4 ASAP.

### DIFF
--- a/test/verify/check-package
+++ b/test/verify/check-package
@@ -71,7 +71,8 @@ class TestPackage(composerlib.ComposerCase):
         release = b.text("li[data-component=tmux] .cc-component__release strong")
         b.set_input_text("#filter-blueprints", release)
         b.key_press("\r")
-        b.wait_text("ul[data-list=components] li:nth-child(1) #tmux", "tmux")
+        b.wait_text("ul[data-list=components] li:nth-child(1) .cc-component__release strong",
+                    release)
         b.click(".cmpsr-panel__body--main .toolbar-pf-results .pficon-close")
         b.wait_text("ul[data-list=components] li:nth-child(1) #httpd", "httpd")
 


martinpitt:
Ugh, [what a disaster](https://logs.cockpit-project.org/logs/pull-1220-20210124-183506-04aa2385-rhel-8-4-firefox/log.html).. `Job for osbuild-composer.socket failed`. Did that somehow break in 8.4? (which shows how c-composer should have begun to start testing on 8.4 months ago..) @jkozol , @larskarlitski , does that ring a bell? Known breakage or API change?

teg:
Hm, this seems like a potential regression introduced in dc1b84fcfeb946a1a0553c7af9df0c448cacc21a. `osbuild-composer.service` needs to `Require/After=osbuild-composer.socket`, which used to be pulled in transitively by the worker. What do you think @ondrejbudai ?

ondrejbudai:
I don't understand what's the issue here. `osbuild-composer.service` can run just alright without `osbuild-composer. socket`. Also, the local worker.socket/.service doesn't imply that the `osbuild-composer.socket` must be enabled.

martinpitt:
@ondrejbudai : FYI, @henrywang backported a few unit related fixes in #1221. He'll cherry-pick this test robustification as well, then I'll close my PR and we sort it all out in #1221.

ondrejbudai:
Ah, thanks, @martinpitt and @henrywang!

PR #1227, opened 2021-02-01
Move some tests from Travis CI to Github Actions

henrywang:


.github/workflows/eslint-styleline.yml
```diff
@@ -0,0 +1,42 @@
+name: eslint-stylelint
+
+on:
+  pull_request:
+    branches:
+      - master
+      - rhel-8
```
martinpitt, line 7: Any reason why you are so picky here? `on: [pull_request]` should be quite enough, and then also cover future stable branches?

.github/workflows/rpm-build.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  rpm-build:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run rpm-build test
+        run: make test_rpmbuild
```
martinpitt, line 17: Does that invoke the container by itself? I don't see a `container:` in the workflow, but then again Travis didn't do that either.

martinpitt (dismissed):
Nice, thank you! One proposed simplification and one question, but this can also be refined later.

henrywang:
Thanks @martinpitt! :+1:  All update!

martinpitt:
@henrywang : Your call of course, but I found the `make test_rpmbuild` better. I didn't realize at first that this calls `sudo docker run ...` by itself, which is fine -- it makes it much easier to run that test locally. Of course at some point the Makefile should be changed to work with podman, and unprivileged :) (But that's for another time).

martinpitt (approved):
(but good enough for sure)

PR #1229, opened 2021-02-04
Wizard: temporarily disable uploading to vmware.

jkozol:
Support for uploading to vmware has not yet landed in osbuild-composer. This patch temporarily removes the upload option.

### DIFF
--- a/components/Wizard/ImageStep.js
+++ b/components/Wizard/ImageStep.js
@@ -311,6 +311,7 @@ class ImageStep extends React.PureComponent {
       </FormGroup>
     );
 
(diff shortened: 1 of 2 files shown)


teg (dismissed):
Thanks for this! @jmikovic, let's reenable this once your PR lands in osbulid-composer. I discussed with @jkozol, and there is no good way to test that the new API is ready from CI, so we'll just need to keep track of things.

Gundersanne (approved):

teg (approved):

jkozol:
These test failures are unrelated to this change.

PR #1232, opened 2021-02-16
Add support for additional RHEL Edge image types

jkozol:
A user can now select the `RHEL for Edge Container (.tar)` image type. This image type supports the optional ostree fields for a parent commit, a ref for the name of the branch, and a url for the parent repository.

The user can also select the `RHEL for Edge Installer (.iso)` image type. This image type supports the required ostree fields of a ref for the name of the branch and a url for the parent repository.

The help text for the new URL field is `Provide the URL of the upstream repository. This is where the OSTree commit will be pulled from.` Does this explain what the URL represents?

There are currently no tests for this PR because these image types are not yet supported by osbuild-composer. They will not display for selection until they are supported.

### DIFF
--- a/components/Wizard/CreateImageUpload.js
+++ b/components/Wizard/CreateImageUpload.js
@@ -64,6 +64,7 @@ class CreateImageUploadModal extends React.Component {
       ostreeSettings: {
         parent: undefined,
         ref: undefined,
+        url: undefined,
       },
       showUploadAwsStep: false,
       showUploadAzureStep: false,
@@ -84,6 +85,7 @@ class CreateImageUploadModal extends React.Component {
     this.setImageType = this.setImageType.bind(this);
     this.setOstreeParent = this.setOstreeParent.bind(this);
     this.setOstreeRef = this.setOstreeRef.bind(this);
+    this.setOstreeURL = this.setOstreeURL.bind(this);
     this.setUploadSettings = this.setUploadSettings.bind(this);
     this.handleUploadService = this.handleUploadService.bind(this);
     this.handleCreateImage = this.handleCreateImage.bind(this);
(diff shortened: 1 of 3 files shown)


jkozol:
Ignore the selected image type. I had to fake it to check the fields.


rheledgeinstaller
rheledgecontainer

teg:
This is great! thanks @jkozol!
I wonder about the second screen, should we make the label `Parent Repository URL`, to make it clear that this is where the parent is taken from, and that we are not actually pushing the container anywhere. Could also put the `Parent commit` box under the `Parent Repository URL` one and in the help text say that it will be inferred from the parent repository if left blank.

teg:
The help text for the `Parent Repository URL` should probably be something like `Provide the URL of the upstream repository. This is where the OSTree commit will be pulled from.` We don't strictly speaking pull in the commit (yet), but this is an implementation detail, and we very well might do in the future.

jkozol:
@teg Would `Parent repo URL` be descriptive enough? There are some limits on how long the labels can be before they wrap and I would prefer to not override patternfly's default styling for this one.

teg:
> @teg Would `Parent repo URL` be descriptive enough? There are some limits on how long the labels can be before they wrap and I would prefer to not override patternfly's default styling for this one.

Maybe just keep it as is and just update the help text?

jkozol:
@teg  `Provide the URL of the upstream repository. This parent repository is where the OSTree commit will be pulled from.`?

teg:
> @teg `Provide the URL of the upstream repository. This parent repository is where the OSTree commit will be pulled from.`?

I think `This repository is where the parent OSTree commit will be pulled from.`

teg (approved):
Thanks for this!

components/Wizard/CreateImageUpload.js
```diff
@@ -316,7 +338,13 @@ class CreateImageUploadModal extends React.Component {
   }
 
   requiresImageSize(imageType) {
-    if (imageType === "" || imageType === "fedora-iot-commit" || imageType === "rhel-edge-commit") {
+    if (
+      imageType === "" ||
+      imageType === "fedora-iot-commit" ||
+      imageType === "rhel-edge-commit" ||
+      imageType === "rhel-edge-container" ||
+      imageType === "rhel-edge-installer"
```
gicmo, line 346: I wonder if we should not just add `fedora-iot-{container, installer}` as well?

jkozol:
Closing as this has diverged uncomfortably far from master and rebasing would be more painful than starting off of master.

PR #1235, opened 2021-02-18
package.json: Lock @patternfly/react-styles version

martinpitt:
It's part of patternfly-react, so we should update it in a controlled
fashion. Letting it auto-update freely often breaks the build.

Similar to https://github.com/cockpit-project/cockpit/commit/bbde4c01947

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "4.80.3",
-    "@patternfly/react-core": "^4.90.2",
+    "@patternfly/react-core": "4.97.2",
     "@patternfly/react-icons": "4.8.4",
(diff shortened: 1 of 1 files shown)


martinpitt:
This avoids [this kind of failures](https://logs.cockpit-project.org/logs/pull-1684-20210218-052956-38f52238-rhel-8-4-osbuild-cockpit-composer-rhel-8/log.html). That happened on the rhel8 branch (I'll send a backport in a moment), but could just as easily hit master. It recently did so for both cockpit and cockpit-podman, we did the same fix there.

martinpitt:
Eh, what? That worked just fine locally...

martinpitt:
Works now.

jkozol (approved):
This makes sense to me. Thanks.

PR #1237, opened 2021-02-18
wizard: translate image type labels

jkozol:
Most of the image type labels are proper nouns and haven't needed to be translated. But, with the addition of a descriptive label for tar, these labels should now be translated.


Gundersanne (approved):

jkozol:
This ostree error is unrelated to this PR. I tested it locally and was able to successfully build the ostree image on this PR on a fedora 33 vm. So, I am going to merge this.

PR #1239, opened 2021-02-19
wizard: fix image type translations

jkozol:
In order for image type labels to be translated they cannot be passed by an object. The image type labels needing translation are now explicitly declared.

This is a subpar fix but a quick fix is needed before string freeze.


Gundersanne (approved):

jkozol:
Again, the edge test failure is unrelated to this PR. Also, the fedora-33 tests passed for it, so I will merge.

PR #1241, opened 2021-02-19
test: allow avc denial messages on cancel image

jkozol:
The fedora 32 test is consistently failing due to an avc denial error as seen on this PR https://github.com/osbuild/cockpit-composer/pull/1240.

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -668,6 +668,7 @@ class TestImage(composerlib.ComposerCase):
         b.wait_attr("#{}-actions".format(uuid), "aria-expanded", "true")
         b.click("li[aria-labelledby={}] a:contains('Remove')".format(selector))
         b.wait_not_present("#{}".format(selector))
+        self.allow_journal_messages(".*avc:  denied.*")
 
         # collect code coverage result
         self.check_coverage()


test/verify/check-image
```diff
@@ -668,6 +668,7 @@ class TestImage(composerlib.ComposerCase):
         b.wait_attr("#{}-actions".format(uuid), "aria-expanded", "true")
         b.click("li[aria-labelledby={}] a:contains('Remove')".format(selector))
         b.wait_not_present("#{}".format(selector))
+        self.allow_journal_messages(".*avc:  denied.*")
```
marusak, line 671: This does not seem like a correct solution.
We normally report AVCs into BZ and then create naughty, like https://github.com/cockpit-project/bots/issues/1688 , https://github.com/cockpit-project/bots/issues/1662 or https://github.com/cockpit-project/bots/issues/1657

marusak: https://github.com/cockpit-project/bots/issues/1662 is very similar btw.

marusak: Also regrading #1240 - I see you merged it with non-finished tests. We now try to get PRs to green as we collect these data in https://images-frontdoor.apps.ocp.ci.centos.org/tests.html
It also collects retries, so no worries that retry would hide the fact, that some tests are failing.

https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=osbuild%2Fcockpit-composer is just for c-composer.

jkozol: @marusak Thanks for these comments. I will open a BZ and then create naughty for the specific avc denial. In regards to merging it I felt that it was fine since it was only a translations PR. Thanks for sharing this dashboard.

PR #1244, opened 2021-02-21
Add URL field to Edge commit

larskarlitski:
This is based on @jkozol's work in #1232, but doesn't yet include the new container or installer image types.

Also includes a commit which might fix tests and new strings (in preparation for string freeze).

### DIFF
--- a/components/Wizard/ImageStep.js
+++ b/components/Wizard/ImageStep.js
@@ -37,6 +37,9 @@ const ariaLabels = defineMessages({
   ostreeRef: {
     defaultMessage: "OSTree ref help",
   },
+  ostreeURL: {
+    defaultMessage: "OSTree repo url help",
+  },
 });
 
 const messages = defineMessages({
@@ -181,6 +184,7 @@ class ImageStep extends React.PureComponent {
       ostreeSettings,
       requiresImageSize,
       setOstreeParent,
+      setOstreeURL,
       uploadService,
     } = this.props;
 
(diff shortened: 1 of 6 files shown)


jkozol:
@larskarlitski To make sure I understand correctly, since https://github.com/osbuild/osbuild-composer/pull/1235 has landed we now support the parent repository url, the commit, and the ref as parameters for our ostree image. For the `rhel-edge-commit` and `fedora-iot-commit` image types it is okay to pass all three parameters as we would for the future container image type?

teg:
@jkozol exactly right! Strictly speaking you can't pass both `parent` and `url`, but you can pass either. And it is exactly the same for the three image types you mentioned. For the future `rhel-edge-installer` it is almost the same, except `parent` does not make sense.

components/Wizard/ImageStep.js
```diff
@@ -579,6 +579,8 @@ class ImageStep extends React.PureComponent {
           {/* {imageType === "vmdk" && vmwareProviderCheckbox} */}
           {requiresImageSize(imageType) && imageSizeInput}
           {(imageType === "fedora-iot-commit" || imageType === "rhel-edge-commit") && ostreeFields}
+          {imageType === "rhel-edge-container" && ostreeFields}
+          {imageType === "rhel-edge-installer" && ostreeFields}
```
teg, line 583: It is worth noting that this will get its own UI (see @jkozol's draft PR which introduces that), but I agree that this is the best placeholder until we land that (which we may want to wait for support in osbuild-composer first to do, as it would otherwise be untested). The generic ostree UI is a bit misleading for this image type, but if we are going to have a place holder it is better than the non-ostree one. In particular, the `parent` field does not make sense for the installer, and the help text for the `url` is not quite right (it should not refer to a parent commit). It would be fully functional though.

jkozol:
> Strictly speaking you can't pass both parent and url, but you can pass either.

My concern with this is that we do not make it clear to the user that they cannot pass in both.

components/Wizard/ImageStep.js
jkozol: I think we should leave these out. I agree with adding strings but since the `rhel-edge-installer` image-type does not accept a commit field I do not think it should be displayed. Until osbuild-composer's weldr api returns `rhel-edge-container` and `rhel-edge-installer` in the list of supported image-types these components will not display anyways. If osbuild-composer starts to support `rhel-edge-installer` before cockpit-composer has proper UI for users will be presented the commit field which isn't supported.

teg: > If osbuild-composer starts to support rhel-edge-installer before cockpit-composer has proper UI for it, users will be presented the commit field which isn't supported.

Yeah, it really isn't nice. The way I see it, we have two alternatives in this (hopefully very short-lived) state:
 - show a screen with an unsupported field (which can be made to work if you just ignore that field)
 -  show a screen without the required fields (which can never work at all)

I figured the former was preferable, but happy to let you call this one, there are no good solutions I think, and either way it will be short-lived / never seen.

teg:
> > Strictly speaking you can't pass both parent and url, but you can pass either.
> 
> My concern with this is that we do not make it clear to the user that they cannot pass in both.

Yeah, that's a good point. How should we handle it? Merge as-is for now, and then do something nicer in a follow-up, or hold off on the whole thing until we have something better? Would we manage to get that by tomorrow though? I'm not sure how much work the alternative entails.

jkozol (dismissed):
I visually reviewed this in fedora 33 and rhel 84. All fields and text appears as expected. Future improvements will be needed as the edge image types gain support in osbuild-composer. But, for now, I agree that this is the best way to initially implement these image types and new field.

teg:
New tests are failing. Seems to confirm the timeout theory? I guess we should bump the timeout across the board?

larskarlitski:
> New tests are failing. Seems to confirm the timeout theory? I guess we should bump the timeout across the board?

These are new failures related to deleting images. I'm investigating,.

PR #1245, opened 2021-02-21
Add edge warnings

jkozol:
Either the parent commit or repository url can be specified. Not both. If both are entered the create button is disabled and an error message is displayed.

This is based on #1244

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -589,7 +589,7 @@ class TestImage(composerlib.ComposerCase):
         self.assertEqual(image_type, image_type_ostree)
 
         # image building needs more time
-        with b.wait_timeout(1800):
+        with b.wait_timeout(3600):
             b.wait_text("li[aria-labelledby={}] [data-status=true]".format(selector),
                         "Image build complete")
         # log should contains rpm, hostname, users stages and tar assembler
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -9,7 +9,6 @@ import composerlib
 import testlib
 
 
-@testlib.nondestructive
 @testlib.timeout(2400)
 @testlib.no_retry_when_changed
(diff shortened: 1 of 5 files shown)


jkozol:
Whichever field is entered second will display the error.

doubleEdge

teg:
Love this!

teg (dismissed):

teg (dismissed):

test/verify/check-image
```diff
@@ -611,13 +611,18 @@ class TestImage(composerlib.ComposerCase):
...
-        b.click("li[aria-labelledby={}] a:contains('Delete')".format(selector))
-        b.wait_attr("#{}-actions".format(uuid), "aria-expanded", "false")
-        b.click("#cmpsr-modal-delete button:contains('Delete image')")
-        b.wait_not_present("#{}".format(selector))
+        # b.click("#{}-actions".format(uuid))
+        # b.wait_attr("#{}-actions".format(uuid), "aria-expanded", "true")
+        # b.click("li[aria-labelledby={}] a:contains('Delete')".format(selector))
+        # b.wait_attr("#{}-actions".format(uuid), "aria-expanded", "false")
+        # b.click("#cmpsr-modal-delete button:contains('Delete image')")
+        # b.wait_not_present("#{}".format(selector))
```
larskarlitski, line 624: We've seen this in testOpenStack below as well. Probably makes sense to disable that, too?

jkozol: I believe the openstack tests failed for a different reason due to there being multiple identical selectors when the ostree test fails to delete. But, I am going removing this for now since we are still unsure what is causing this issue and it may be causing openstack to fail.

henrywang:
@jkozol, all green already!

PR #1246, opened 2021-02-21
test: double image build timeout

teg:
Set it to 60m rather than 30m.

Either should be plenty, but in order to eliminate slow builds as a source of false negatives, let's up the limit.


henrywang:
@jkozol, @teg, do we still need this? Thanks.

teg:
No, this was not the issue after all :)

PR #1247, opened 2021-02-22
Add new edge image type support and validation, reenable vmware

jkozol:
This reverts commit 08e46cdb3ef761f2f4d6d75c97e90d786ca86432. VMWare is now a selectable upload type.

### DIFF
--- a/cockpit-composer.spec.in
+++ b/cockpit-composer.spec.in
@@ -12,10 +12,10 @@ BuildRequires:  libappstream-glib
 
 Requires:       cockpit
 %if 0%{?fedora} >= 33 || 0%{?rhel} >= 8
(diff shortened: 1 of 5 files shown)


teg (approved):
Thanks! cc @jmikovic

teg (approved):
🥳

larskarlitski (approved):
Very nice! Thanks for seeing this through!

teg:
We are seeing some really bad CI flakes here apparently. They are not specific to this PR, and manual testing seems to work just fine. Considering the Firefox tests pass for both Fedora and RHEL, I vote in favor of merging as is, and doing having a discussion about the state of our CI next week to see how we can avoid a situation like this in the future.

PR #1251, opened 2021-02-22
test: readd nondestructive

jkozol:
Also select logs using selector


larskarlitski (dismissed):
Yep. Thanks for the quick fix.

henrywang (approved):
@jkozol, @larskarlitski. I push a commit here to fix "Download" drop-down item css selector issue. All green now.

PR #1252, opened 2021-02-23
cockpituous: Add Fedora 34, drop Fedora 32

martinpitt:
bodhi activation for Fedora 34 is today [1]. Drop Fedora 32 instead,
which is soon EOL.

[1] https://fedorapeople.org/groups/schedule/f-34/f-34-key-tasks.html

----

This is what I did to cockpit{,-podman,ostree}. Of course feel free to adjust if you want to keep F32 for a while. This is mostly a reminder that this needs doing.


henrywang (approved):
Thanks @martinpitt!

PR #1258, opened 2021-03-03
test: Run OSTree test in VM with more RAM

martinpitt:
The OSTree build needs more memory than the standard 1.1GiB RAM
nondestructive VMs. This was leading to excessive memory pressure, CPU
churn (from constantly trying to move things around), and eventually
crashing VMs and long timeouts.

This gets aggravated on cloud images which don't have swap; fedora-33 is
being moved to a cloud image in https://github.com/cockpit-project/bots/pull/1719

Split out the test into a destructive class which provisions a bigger
machine.

Also drop the obsolete test skip on Fedora 31, we are not supporting
that any more.

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -4,7 +4,6 @@
 # 1. create supported image
 
 import os
-import unittest
 import composerlib
 import testlib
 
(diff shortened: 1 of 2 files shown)


test/verify/check-image
```diff
@@ -548,8 +548,62 @@ class TestImage(composerlib.ComposerCase):
         # collect code coverage result
         self.check_coverage()
 
-    @unittest.skipIf(os.environ.get("TEST_OS") == "fedora-31", "Does not support ostree image")
-    def testOSTree(self):
+    def testCancel(self):
```
martinpitt, line 551: Note: The diff looks weird. Conceptually I didn't move `testCancel()`, but `testOSTree()` into a new class below, but apparently testCancel() is shorter :grin:

martinpitt:
Note that this blocks testing on fedora 34, which is also a cloud image. So is rhel-8-4, but I suppose "rhel-edge-commit" is just a bit smaller and thus barely fits?

martinpitt:
Meh, this really hates me.. now I get timeouts in other tests, and looking at the bots I now have *three* libvirt VMs in "paused" state.

martinpitt:
This is the reason:

    tmpfs                                                                                                10G   10G     0 100% /tmp

VM overlays are getting too big, and as there are now 3 VMs running in parallel they aggravate the overflow on the host.

henrywang (approved):
Thanks @martinpitt.

martinpitt:
[testOpenstack](https://logs.cockpit-project.org/logs/pull-1258-20210303-065426-868f2995-fedora-33-bots-1719/log.html) is in the same boat. I moved this to a bigger machine as well.

henrywang (approved):

PR #1260, opened 2021-03-10
Do not copy testmap.py for test dependence

henrywang:
testmap.py should not be used by test


martinpitt (approved):
Oops, thanks! This also broke cockpit's selenium tests, so it *may* happen that we revert the lib commit if we can't figure this out really quickly. But approving for now, so you can land this whenever it starts to block you.

FYI @allisonkarlitskaya

henrywang:
Thanks @martinpitt. No urgent PR to land recently, we can wait until cockpit's broken issue gets fixed.

martinpitt:
@henrywang : Seems we [have it under control](https://github.com/cockpit-project/cockpit/pull/15494), so go ahead.

henrywang:
> @henrywang : Seems we [have it under control](https://github.com/cockpit-project/cockpit/pull/15494), so go ahead.

Nice! Thanks for update.

PR #1262, opened 2021-03-12
test: Bump cockpit test API to 239

martinpitt:
I will work on bringing back the `@nondestructive` tests, that will require a new option in `run-tests`. Let's first make sure that composer tests work with the most recent release.

 - [ ] https://github.com/cockpit-project/bots/pull/1787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ machine: bots
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # this needs a recent adjustment for firefox 77 and working with network-enabled tests
 test/common:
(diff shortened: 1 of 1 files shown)


martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-1262-20210312-060925-9dd70cbb-fedora-33-firefox/log.html) is an utter mess. I'm trying to fix this with https://github.com/cockpit-project/bots/pull/1787 .

martinpitt:
`TestSource.testBasic` seems really hard to please.. But aside from that, the current test API works, which is all I wanted to know. I'll rather focus on #1263 instead.

PR #1265, opened 2021-03-16
test: update README

jkozol:
In order to run individual tests bots/vm-run must be called and the tests must point to that machine. The README is updated to explain this.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -31,24 +31,32 @@ To run all tests run the following:
 
     $ make check
 
-Alternatively you can run an individual test like this:
+Alternatively, you can run individual tests. To list the individual test names you can run:
 
-    $ ./test/verify/check-image
+    $ test/common/run-tests --test-dir=test/verify -l
+
+And then run:
+
+    $ test/common/run-tests --test-dir=test/verify $TEST_NAME
+
+To run the tests with network, use the `--enable-network` flag:
+
+    $ test/common/run-tests --test-dir=test/verify --enable-network $TEST_NAME
 
 To see more verbose output from the test, use the `--verbose` and/or `--trace` flags:
 
-    $ ./test/verify/check-image --verbose --trace
+    $ test/common/run-tests --test-dir=test/verify --verbose --trace $TEST_NAME
 
 In addition if you specify `--sit`, then the test will wait on failure and allow you to log into
 cockpit and/or the test instance and diagnose the issue. An address will be printed of the test
 instance.
 
-    $ ./test/verify/check-image --trace --sit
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -31,9 +31,13 @@ To run all tests run the following:
...
 
-Alternatively you can run an individual test like this:
+Alternatively, you can run individual tests. After `make vm` you must run:
 
-    $ ./test/verify/check-image
+    $ bots/vm-run "TEST_OS"
+
+This will output an eth0 for the machine. Using this address, you then run: 
+
+    $ ./test/verify/check-image --machine "eth0 address"
```
marusak, line 40: Not really though. `./bots/vm-run` is used when you want to test more tests against one running machine. You can just do `./test/verify/check-image` and it will run its own VM that is killed when tests finish.

marusak: and of course you can do `TEST_OS=foo ./test/verify/check-machine` if you want to use different image then default.

jkozol: When I run just `./test/verify/check-image` the tests fail due to no network access. Yet, when I use `bots/vm-run` and point at that machine they have access to the network. Do you know why that might be?

marusak: > When I run just ./test/verify/check-image the tests fail due to no network access.

That is expected and wanted. If you want to have network access you can give it `--enable-network` which btw. is what you do in your tests: https://github.com/osbuild/cockpit-composer/blob/master/Makefile#L133

> Yet, when I use bots/vm-run and point at that machine they have access to the network.

Because by default `vm-run` connects to internet. You need to do `--no-network` to stop that.

While this might seem odd, it makes sense - By default you want tests to run in separate environment thus no network. But one uses vm-run for stuff like debugging issues, checking if something is available... so having network is useful to have to make life easier.

jkozol: Hmmm. What is the difference between running `test/common/run-tests --test-dir=test/verify --enable-network` and running an individual test with `./test/verify/check-image --enable-network`? When using `make check` the tests do run with the network enabled but when I run the individual tests the `--enable-network` flag seems to have no effect and the tests do not have network access.

marusak: hmm, not sure from top of my head, should work.
But in any case, you can just use `test/common/run-tests --test-dir=test/verify --enable-network <test_name>` to run one test. (or more, just add all test names you want to run).
If you are not sure about names, you can simply do `test/common/run-tests --test-dir=test/verify -l` to list all the names.

jkozol: Okay. I updated the README to run the individual tests using `test/common/run-tests --test-dir=test/verify --enable-network`. The --enable-network flag works when used this way.

henrywang (approved):
LGTM.

PR #1267, opened 2021-03-18
Add RHEL Edge Container and Installer image types

jkozol:
The RHEL for Edge Container and Installer image types are now supported. The tests will fail until osbuild-composer 28.1 is in the RHEL 8.4 nightly.

The RHEL for Edge Container (.tar) image type supports the optional ostree fields for a parent commit, a ref for the name of the branch, and a url for the parent repository.

edgeContainerError

The RHEL for Edge Installer (.iso) image type supports the required ostree fields of a ref for the name of the branch and a url for the parent repository. If the url is not provided the user will not be able to create an image. If the user enters an empty value for the url they will receive an error message. This error message has been used elsewhere in the CreateImageUpload Wizard so is already translated. This string should have a period at the end but I will leave it as is for now since we should not add new or updated strings.

edgeInstaller
edgeInstallerError

### DIFF
--- a/components/Wizard/ImageStep.js
+++ b/components/Wizard/ImageStep.js
@@ -567,6 +567,82 @@ class ImageStep extends React.PureComponent {
       </>
     );
 
+    const ostreeInstallerFields = (
+      <>
+        <FormGroup
+          label={<FormattedMessage defaultMessage="Repository URL" />}
+          labelIcon={
+            <Popover
+              id="ostree-url-popover"
+              bodyContent={
+                <FormattedMessage defaultMessage="Provide the URL of the upstream repository. This repository is where the parent OSTree commit will be pulled from." />
+              }
+              aria-label={formatMessage(ariaLabels.ostreeURL)}
+            >
+              <Button variant="plain" aria-label={formatMessage(ariaLabels.ostreeURL)}>
+                <OutlinedQuestionCircleIcon className="cc-c-text__align-icon" id="popover-icon" />
+              </Button>
+            </Popover>
+          }
+          fieldId="ostree-url-input"
+          helperTextInvalid={formatMessage(messages.requiredField)}
+          validated={ostreeSettings.url !== "" ? "default" : "error"}
+          isRequired
+          hasNoPaddingTop
+        >
+          <TextInput
+            className="pf-c-form-control"
+            value={ostreeSettings.url !== undefined ? ostreeSettings.url : ""}
+            type="text"
+            id="ostree-url-input"
+            onChange={this.handleOSTreeURL}
+            validated={ostreeSettings.url !== "" ? "default" : "error"}
+          />
+        </FormGroup>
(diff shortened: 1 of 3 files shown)


test/verify/check-image
```diff
@@ -560,9 +560,10 @@ class TestLargeImage(composerlib.ComposerCase):
                         "Image build complete")
         # log should contains rpm, hostname, users stages and tar assembler
         b.click("li[aria-labelledby={}] button:contains('Logs')".format(selector))
-        b.wait_in_text("#{}-logs".format(uuid), "Stage org.osbuild.rpm")
-        b.wait_in_text("#{}-logs".format(uuid), "Stage: org.osbuild.rpm-ostree")
-        b.wait_in_text("#{}-logs".format(uuid), "Assembler org.osbuild.ostree.commit")
+        # temporarily disable log contents checking
+        # b.wait_in_text("#{}-logs".format(uuid), "Stage org.osbuild.rpm")
+        # b.wait_in_text("#{}-logs".format(uuid), "Stage: org.osbuild.rpm-ostree")
+        # b.wait_in_text("#{}-logs".format(uuid), "Assembler org.osbuild.ostree.commit")
```
teg, line 566: I think cockpit-composer should not assert on the contents of the logs (but composer should!).

jkozol: So you think these log checks should be permanently removed?

components/Wizard/ImageStep.js
```diff
@@ -567,6 +567,198 @@ class ImageStep extends React.PureComponent {
       </>
     );
 
+    const ostreeContainerFields = (
```
teg, line 570: Should this not be the same as the `ostreeCommitFields`, how come it was split?

jkozol: Ah yes. I split it thinking the UI for image types would be different. Removing the duplicate code now.

test/verify/check-image
ondrejbudai: Yeah, I just spoke about that with Tom, and we suggest removing these checks.

teg: Yes :) The logs are meant to be human readable only, so we should treat them as opaque and not assert on their contents (could verify that they are not empty though!). Reason being that they may change at any time in the future.

This did uncover a real bug though! So this should be fixed in composer, and tests should be added there to verify the logs are as expected (even if opaque they should be useful...). Then if ever composer changes the format of the logs, only the composer tests would have to be adjusted.

jkozol: Okay will fully remove these.

components/Wizard/ImageStep.js
teg: They should be the same in this case (whether the code is the same is of course up to you, just wanted to verify that the UX is identical).

teg (approved):
Great stuff!

PR #1271, opened 2021-03-24
CreateImageUpload: require slash in url

jkozol:
The edge image type's url requires a slash at the end of the url. If no slash is present show an error icon on the field and disable image creation.

An error message should be added for this validation to explain the failure but I am opening this PR without it in case this needs to be backported to RHEL 8.4


teg:
Thanks for looking into this so quickly @jkozol! I think the right fix should be https://github.com/osbuild/osbuild-composer/pull/1329 and that this won't be needed after all.

PR #1290, opened 2021-05-04
test: updated rhel 8.4 repos

jkozol:
Since RHEL 8.4 is released we no longer point at the nightly compose of 8.4.


jkozol:
I am not a fan of merging PRs with tests failing but this PR should fix the image refresh issue in #1984. The issue seems to be that osbuild-composer's repos are overridden to use the RHEL nightly. Since the RHEL 8.4 nightly no longer exists, osbuild-composer breaks.


I manually tested this PR using this branch of bots checked out https://github.com/jkozol/bots/tree/updateRHEL84Nightly and the tests pass.

@marusak @henrywang any thoughts?

marusak:
Those 8.4 that fail to get prepared are due to https://github.com/cockpit-project/bots/pull/1984 ? If so, then can you please test them with `@bots` as well?

And since `TestLargeImage` is busted, how about skipping it as known broken test to unblock everything? `self.skipTest()` could do or anything other that works for you. + possibly opening issue for good measure.

martinpitt:
The URL changes here look correct, and the fedora failures are "just" the long-standing TestLargeImage failures. The rhel-8-4 failures are due to using the old image, which is fixed in https://github.com/cockpit-project/bots/pull/1984 . But that in turn is blocked on this PR :grin: 

To give them a spin on the rhel-8-4 image, I triggered these additional tests here. Thanks!

martinpitt:
I wonder if instead of hardcoding the repo URLs again, it could just grab them from the running image? Then you wouldn't have to keep updating them for 8.y updates for nightly → release.

/etc/yum.repos.d/nightly.repo has the info on our test images, and so does `dnf repoinfo`.

jkozol:
> I wonder if instead of hardcoding the repo URLs again, it could just grab them from the running image? Then you wouldn't have to keep updating them for 8.y updates for nightly → release.
>
> /etc/yum.repos.d/nightly.repo has the info on our test images, and so does dnf repoinfo.

@martinpitt I like this idea but I don't think it is the most practical solution since osbuild-composer requires a json format of the repo data. I will see if there is a nice way we can pull the same repo overrides that osbuild-composer uses for testing.

martinpitt (approved):
Tests are green against the new RHEL 8.4 image, thanks!

martinpitt:
Can you please also backport this to the rhel-8 branch?

PR #1292, opened 2021-05-05
test: updated rhel 8.4 repos

jkozol:
Since RHEL 8.4 is released we no longer point at the nightly compose of 8.4.


martinpitt (approved):
Danke! New image just landed, to tests ought to get green.

PR #1302, opened 2021-06-15
test: update check-source

jkozol:
The repo url in check-source's testBasic was broken. This has now been replaced by a fedora 30 mirrorlist which should maintain more reliability. Additionally the packages check is removed from this test. Packages are checked in the check-package test.


henrywang (approved):
@jkozol LGTM. Thanks!

PR #1317, opened 2021-07-02
workflows: Run release workflow in release environment

allisonkarlitskaya:
This partitions our secrets and limits their potential exposure/leakage.

This environment can be created with github-upload-action-secrets in [1].

[1] https://github.com/cockpit-project/bots/pull/2164

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   cockpituous:
     runs-on: ubuntu-20.04
+    environment: release
     container:
       image: ghcr.io/cockpit-project/release
     steps:


allisonkarlitskaya:
I pushed the environment secrets and cleaned up the relevant items from the repo-level secrets (since there's no org secrets in play for `cockpit-composer`).  Probably needs some extra-checking at release time, compared to the others, but I imagine it'll be fine.

allisonkarlitskaya:
@jkozol It might make sense to attempt a release after this lands to try and shake out any potential problems.  Otherwise, if something goes wrong during the next release, please ping.

In any case, the release won't work without this commit now, so marking as release-blocker.

martinpitt (approved):

PR #1320, opened 2021-07-05
webpack.config.js: silent sass deprecation warnings

allisonkarlitskaya:
We're getting a lot of warnings like

  DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

from PatternFly 3.

We can disable these warnings for our dependencies with the
`quietDeps` option to Dart Sass, as we do in the cockpit repo.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -152,6 +152,7 @@ module.exports = {
                   path.resolve(nodedir, "bootstrap-sass", "assets", "stylesheets"),
                 ],
                 outputStyle: "compressed",
+                quietDeps: true,
               },
             },
           },


webpack.config.js
```diff
@@ -152,6 +152,7 @@ module.exports = {
                   path.resolve(nodedir, "bootstrap-sass", "assets", "stylesheets"),
                 ],
                 outputStyle: "compressed",
+                quietDeps: true,
```
KKoukiou, line 155: I don't think this will take effect, you need 1.35.0 sass version, and I did not check for the sass-loader https://github.com/sass/dart-sass/blob/main/CHANGELOG.md#1350

allisonkarlitskaya: I tried it and it worked :shrug:

allisonkarlitskaya: the version for sass, btw:

```
  "_from": "sass@^1.27.1",
  "_id": "sass@1.35.1",
```

Because we use a `^` we indeed get 1.35.

KKoukiou (approved):

Issue #989, opened 2020-06-05
Manage source window disappeard after add a new source


henrywang:
How to reproduce:
1. login to cockpit from firefox 77 
2. go to manager source page
3. add a new source
4. click "Add Source"
5. the manage source page disappear

Expected result:
the add source page disappears, but manage source page should be there.

NOTE:
1. This issue only exists on Firefox 77, and not on Firefox 69 and chromium.
2. This issue only can be reproduced when adding source at the first time.

Some comments from @martinpitt:

> we have some workarounds in our code where event handlers call ev.preventDefault() or ev.stopPropagation()
> but the real issue is often that the dialog gets rendered by a parent element which is unstable and going away, or getting re-rendered and such

PR #1326, opened 2021-07-09
test: use 2GB nondestructive memory and fix failing tests

jkozol:
The cockpit version for testing is bumped to 248. All tests are now run with 2048 mb of nondestructive memory instead of just the image tests. Therefore, the image tests no longer need to explicitly provision more memory.

### DIFF
--- a/components/Modal/ManageSources.js
+++ b/components/Modal/ManageSources.js
@@ -5,7 +5,8 @@ import React from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import { FormattedMessage, defineMessages, injectIntl, intlShape } from "react-intl";
-import { Modal, Alert, Spinner } from "patternfly-react";
+import { Alert, Spinner } from "patternfly-react";
+import { Modal, ModalVariant, Title } from "@patternfly/react-core";
 import SourcesListItem from "../ListView/SourcesListItem";
 import EmptyState from "../EmptyState/EmptyState";
 import {
@@ -159,7 +160,7 @@ class ManageSourcesModal extends React.Component {
     }
   }
 
-  handleShowForm(e, showForm) {
+  handleShowForm(ev, showForm) {
     this.setState({ addEntry: showForm });
     if (showForm) {
       this.setState({
@@ -177,14 +178,15 @@ class ManageSourcesModal extends React.Component {
         warningDuplicateUrl: false,
       });
     }
+    ev.preventDefault();
   }
 
-  handleChange(e, input) {
+  handleChange(ev, input) {
     let value;
     if (input === "check_ssl" || input === "check_gpg") {
-      value = e.target.checked;
+      value = ev.target.checked;
     } else {
-      value = e.target.value.trim();
+      value = ev.target.value.trim();
     }
     if (input === "name") {
       this.handleValidateName(value);
@@ -193,6 +195,7 @@ class ManageSourcesModal extends React.Component {
       this.handleValidateUrl(value);
     }
     this.setState({ [input]: value });
+    ev.preventDefault();
   }
 
   handleValidateName(name) {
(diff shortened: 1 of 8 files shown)


jkozol:
@marusak @henrywang I am at a loss for why this test is failing. I have been unable to replicate locally. Do you have any suggestions or thoughts on what could be causing this failure? I have no idea how to continue to debug it.

jkozol:
I would like to merge this PR as is and work in a seperate PR to fix the fedora-33/firefox test failure. These test updates are useful for testing against fedora 34.

components/Modal/ManageSources.js
```diff
@@ -1,11 +1,13 @@
+/* eslint-disable no-unused-vars */
```
marusak, line 1: Why is this necessary?

components/Modal/ManageSources.js
```diff
@@ -356,100 +361,109 @@ class ManageSourcesModal extends React.Component {
...
+            description="Sources provide the contents from which components are selected"
+          />
+        )) ||
+          (this.state.editName === "" && <FormattedMessage defaultMessage="Add source" />) || (
+            <FormattedMessage defaultMessage="Edit source" />
+          )}
+      </Title>
+    );
+
+    const body = (
```
marusak, line 379: phew, this contains a lot of unnecessary brackets and fragments. It is not directly in return of render, thus you can just write something like:
```
const body = Object.keys(manageSources.sources).length === 0 ?              
          <EmptyState                                                           
            title={formatMessage(messages.errorStateTitle)}                     
            message={formatMessage(messages.errorStateMessage)}                 
          />                                                                    
        : !this.state.addEntry ? 
           <> 
               ...
```

Same for `footer`, But mostly just preference of style, so feel free to ignore :)

components/Modal/ManageSources.js
```diff
@@ -356,100 +361,109 @@ class ManageSourcesModal extends React.Component {
...
+            </button>
+          </>
+        )}
+      </>
+    );
+
+    return (
+      <Modal
+        isOpen
+        variant={ModalVariant.medium}
```
marusak, line 459: admittedly, I haven't run this, but we also specify `position="top"`alongside with `variant="medium"`. But it might be, that without this position it aligns more with other existing dialogs?

test/verify/check-source
```diff
@@ -52,17 +52,6 @@ class TestSource(composerlib.ComposerCase):
         # groups action done
         b.click("button:contains('Add source')")
 
-        # HACK: workaround issue https://github.com/osbuild/cockpit-composer/issues/989
```
marusak, line 55: So this commit completely fixes #989 or just one of those problematic dialogs? If all of them, then the commit message could do with `Fixes` line.

marusak (changes requested):
Agree with merging this, no need to block this since current master has that user test broken anyway.

test/verify/check-source
jkozol: Added fixes

components/Modal/ManageSources.js
jkozol: Added the `position="top"`

components/Modal/ManageSources.js
jkozol: Refactored for better style

components/Modal/ManageSources.js
jkozol: Thanks for catching this. Removed.

components/Modal/UserAccount.js
```diff
@@ -285,6 +285,7 @@ class UserAccount extends React.Component {
                     <label>
                       <input
                         type="checkbox"
+                        aria-label="Server admin checkbox"
```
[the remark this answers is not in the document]
jkozol: This is now translated.

marusak (approved):
Thank you!

PR #1330, opened 2021-08-02
Notifications: refactor to use redux and patternfly 4

jkozol:
The notification system is refactored. Notifications are now passed through the redux store so they are globally accessible. Currently, only blueprint commits and image status updates are supported. The Notifications component now uses patternfly alerts to provide a more consistent UI for the notifications.


Gundersanne (approved):
LGTM

PR #1332, opened 2021-08-18
core: don't display unsuported image types

jkozol:
If an image type was supported by osbuild-composer but not cockpit-composer it would still be displayed. Now, an image type must be listed in the list of imageTypeLabels for it to be displayed in the list of available image types.


Gundersanne (approved):

Gundersanne (commented):
The firefox error seems real. I can't find `User account actions user` anywhere in the code. Shouldn't it be present on a dropdown somewhere?

marusak:
> The firefox error seems real. I can't find User account actions user anywhere in the code. Shouldn't it be present on a dropdown somewhere?

This has been broken for months now.

PR #1334, opened 2021-08-20
test: simplify testExport

jkozol:
The export modal test does not need to verify the contents of cockpit-composer and composer-cli. It is unnecessary to test against composer-cli. Also, the export test in TestCustom is removed since it is a duplicate test.


Gundersanne (approved):
makes sense to me, thanks!

PR #1335, opened 2021-08-23
Simplify user test

jkozol:


jkozol:
Removing testing is never ideal but I don't think these checks are important and I have opened issue #1336 to track reimplementing these checks in the future.

Gundersanne (changes requested):
I think this is acceptable in this instance, but please create an issue and refer to it in a comment, stating why, or refer to the user creation rework issue.

jkozol:
@Gundersanne Thanks for the review. I added a comment and linked to the issue.

Gundersanne (approved):
perfec, thanks!

jkozol:
This 8-4 test failure is unrelated to the fix here. I will fix this failure in a separate PR.

PR #1337, opened 2021-08-24
alerts: include error message

jkozol:
Alerts can now pass along an error. If a compose fails the error will be included in the alert. Currently, only a failed depsolve due to no subscription will display a message. All failures will display the alert but only errors mapped to a translated string will display an error message.

### DIFF
--- a/components/Notifications/Notifications.js
+++ b/components/Notifications/Notifications.js
@@ -1,10 +1,16 @@
 import React from "react";
-import { FormattedMessage, injectIntl } from "react-intl";
+import { defineMessages, FormattedMessage, injectIntl, intlShape } from "react-intl";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
 import { Alert, AlertActionCloseButton, AlertGroup } from "@patternfly/react-core";
 import "./Notifications.css";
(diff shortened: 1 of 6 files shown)


components/Notifications/Notifications.js
```diff
@@ -158,6 +169,7 @@ class Notifications extends React.PureComponent {
 Notifications.propTypes = {
   alerts: PropTypes.arrayOf(PropTypes.object),
   alertDelete: PropTypes.func,
+  intl: intlShape.isRequired,
```
Gundersanne, line 172: hm I'm a little bit confused, where is this property actually set? In `components/Layout/Layout.js` I see Notifications being created with the defaultprops, but intl doesn't have a default one.

jkozol: Okay. So. Unfortunately I am actually not 100% sure on this because our translations are done with `react-intl` version 2 and the current version is 5. So, the documentation on `react-intl` is not really accurate for our version. But, as far as I understand, in `main.js` we wrap everything in an `<IntlProvider>` component which will provide intl as a prop.

Gundersanne (approved):
Then it makes sense, thanks!

PR #1351, opened 2021-09-27
Modal: rewrite UserAccount modal

jkozol:
The UserAccount modal is now rewritten to use patternfly 4. It keeps the same basic functionality and validation as the existing modal but has removed some unneccessary store updates. Tests are also updated for the changes in the modal.

createUser

createUserFilled

editUser


Gundersanne (approved):
lgtm!

PR #1359, opened 2021-09-30
Wizard: add Edge Raw Image type

jkozol:
Add support for the edge-raw-image image type. The Repo URL is required and the user can add an optional branch ref. Testing is added to verify the fields display.

rawimage

### DIFF
--- a/components/Wizard/CreateImageUpload.js
+++ b/components/Wizard/CreateImageUpload.js
@@ -298,7 +298,7 @@ class CreateImageUploadModal extends React.Component {
     if (this.state.imageType === "") return true;
     if (
       this.requiresImageSize(this.state.imageType) &&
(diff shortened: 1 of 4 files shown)


achilleas-k:
Thanks.  Did a quick test locally and the frontend works as expected.  Though there's an issue with the build because the image size gets set to 1 GB.  When I build from the command line, it defaults to 10 GB from the image definition.  Does cockpit-composer specify a size when starting the build?

jkozol:
@achilleas-k I added in image-size support. At some point we should figure out default and min image sizes.

achilleas-k (approved):
Excellent!  Looks great now.  Sorry for the mess with the size options :)

jkozol:
@achilleas-k Ehh no worries. All part of the process. Thanks for the review :)

PR #1361, opened 2021-10-01
blueprint: add installation_device customization

jkozol:
A user can now enter an installation device as a blueprint customization.

This is a very simple implementation. The functionality is a little clunky and follows the existing work for the hostname.

Screenshot from 2021-10-01 18-22-54
Screenshot from 2021-10-01 18-22-37

### DIFF
--- a/core/sagas/blueprints.js
+++ b/core/sagas/blueprints.js
@@ -13,6 +13,7 @@ import {
   UPDATE_BLUEPRINT_COMPONENTS,
   SET_BLUEPRINT_USERS,
   setBlueprintUsersSucceeded,
+  SET_BLUEPRINT_DEVICE,
   SET_BLUEPRINT_HOSTNAME,
   setBlueprintHostnameSucceeded,
   SET_BLUEPRINT_DESCRIPTION,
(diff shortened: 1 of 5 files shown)


core/sagas/blueprints.js
```diff
@@ -210,6 +211,36 @@ function* setBlueprintUsers(action) {
...
+    groups: blueprint.groups !== undefined ? blueprint.groups : [],
+    customizations: blueprint.customizations,
+  };
+  return weldrBlueprint;
+}
+
+function* setBlueprintDevice(action) {
+  try {
+    const { blueprint, device } = action.payload;
+    // commit the oldest blueprint with the updated hostname
```
Gundersanne, line 231: Comment isn't entirely correct here I think?

Gundersanne (changes requested):
Test needs fixing, I think because the device is empty? https://logs.cockpit-project.org/logs/pull-1361-20211001-162408-bf2df98a-fedora-33/TestCustom-testHostname-fedora-33-127.0.0.2-2201-FAIL.png Should that be a valid option too?

core/sagas/blueprints.js
jkozol: Thanks. Forgot to remove this comment.

jkozol:
> Test needs fixing, I think because the device is empty? https://logs.cockpit-project.org/logs/pull-1361-20211001-162408-bf2df98a-fedora-33/TestCustom-testHostname-fedora-33-127.0.0.2-2201-FAIL.png Should that be a valid option too?

Blueprint device being empty is a valid option. This function only gets called if a user sets the blueprint device in customization.

The test failures seem to related to duplicate css classes since I used the TextInlineEdit component for a second time. Adding more specificity to the test selectors now.

teg:
This is specific to installer blueprints, right? But shown regardless. Should we indicate this in some way? An "installer" section perhaps?

jkozol:
@teg 
>This is specific to installer blueprints, right? But shown regardless. Should we indicate this in some way? An "installer" section perhaps?

Maybe help text for now? Having a section for installer might be a little confusing if there is only one labelled section. Help text can at least explain when and what this field would be used for. Ideally I think this customization should live in the create image wizard and be specified on image creation but that isn't exactly blueprint conducive.

teg:
Sounds good for now. Probably should discuss the way we present blueprints from the ground up at some point. cc @Gundersanne @gicmo @ondrejbudai @achilleas-k

jkozol:
> Sounds good for now. Probably should discuss the way we present blueprints from the ground up at some point.

Would love this.

gicmo:
We should probably hint (help text or something) that this needs to be the actual device node, e.g. `/dev/sda1`.

jkozol:
> We should probably hint (help text or something) that this needs to be the actual device node, e.g. `/dev/sda1`.

Is it possible to verify that this is a valid device node? Or should we just instruct users to enter an actual device node

gicmo:
> > We should probably hint (help text or something) that this needs to be the actual device node, e.g. `/dev/sda1`.
> 
> Is it possible to verify that this is a valid device node? Or should we just instruct users to enter an actual device node

Verifying is tricky since some device nodes might only exist on the target hardware (think raid devices).

jkozol:
Screenshot from 2021-10-05 14-59-12

Gundersanne (approved):

PR #1365, opened 2021-10-06
Wizard: add support for edge-simplified-installer

jkozol:
The RHEL for Edge Simplified Installer is now supported by cockpit-composer. The user required fields are the same as for the Edge Installer. Testing is added.
Screenshot from 2021-10-06 18-33-49


achilleas-k:
I think we might have the same issue we did with the size of the Edge raw image.  I never notice this because in composer the raw image defaults to 10 GB but the compose request from the wizard defaults to 2 GB, which isn't enough.

For context, the Edge Simplified Installer builds an Edge Raw Image and embeds it in an installer image.

PR #1374, opened 2021-11-18
WIP: Add support for OCI upload target

rgolangh:
Signed-off-by: Roy Golan <rgolan@redhat.com>


teg:
Thanks for this @rgolangh! Nice to see it come together this quickly. I'm not able to review the JS though, could you have a look @jkozol?

jkozol:
@rgolangh Thanks for opening this. I see it is still marked as WIP. Once you feel it is ready please @ me and I will provide a review. Looks very good so far with only minor issues

data/providers.js
```diff
@@ -25,6 +25,42 @@ const providerSettings = {
       },
     },
   },
+  oci: {
```
jkozol, line 28: I believe this file is currently unused so no need to update it.

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,156 @@
...
+              </Popover>
+            </div>
+            <TextArea
+              className="pf-c-form-control"
+              value={uploadSettings.private_key}
+              type="text"
+              id="private_key-input"
+              name="private_key"
+              onChange={setUploadSettings}
+            />
```
jkozol, line 101: You had mentioned wanting a file upload feature here. I am bumping the patternfly version so you can now follow the documentation [here](https://www.patternfly.org/v4/components/file-upload). I would recommend not using the setUploadSettings function and instead writing a custom one specifically for the private key. You could do something like:

```
  handleFileInputChange(event, file) {
    this.setState({ filename: file.name });
  }

  handleDataChange(value) {
    this.props.setPrivateKeyFile(value);
  }

...

<FileUpload
  type="text"
  value={this.props.uploadSettings.private_key}
  filename={this.state.filename}
  filenamePlaceholder="Placeholder"
  browseButtonText="Upload"
  hideDefaultPreview
  onFileInputChange={this.handleFileInputChange}
  onDataChange={this.handleDataChange}
/>
```

jkozol (changes requested):
This is great! Thanks for the contribution. Overall this is basically ready to land. I do not have an OCI account so I cannot verify the actual upload ability but from a code review perspective only a few nitpicks. Once you fix the lint issues I would be happy to land this. But, I would like a review page added. You can follow the same patterns used by the other upload providers in the `ReviewStep.js` file. Also, if you could add a test under `test/verify/check-image` that would be great. I am happy to land this without those and we could create issues and land them in follow up PRs.

data/providers.js
rgolangh: done

components/Wizard/OCIAuthStep.js
rgolangh: Thanks for the version bump. I'm still not sure if to stay with the textarea or the file upload. Any preferences from the whole application UX perspective?

jkozol: If this needs to be a secret value than I think using a file upload would be better. Such a large entry all as secret wouldn't be particularly easy for users I believe. If it can be plaintext than I think you could use both a text area and a file upload. If you don't include `hideDefaultPreview` the fileupload also includes a textarea.

jkozol: The patternfly packages are now updated (was waiting on tests). Feel free to rebase on top of main.

rgolangh: Used it now and it works fine. I specifically set the value in the uploadSettings in the end. Do you still think a dedicated function is needed?

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,174 @@
...
+                <Button variant="plain" aria-label={formatMessage(ariaLabels.private_key)}>
+                  <OutlinedQuestionCircleIcon id="popover-icon" />
+                </Button>
+              </Popover>
+            </div>
+            <FileUpload
+              id="private_key-input"
+              type="text"
+              name="private_key"
+              className="pf-c-form-control"
```
jkozol, line 108: Remove this line to prevent the double form input from showing.

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,174 @@
...
+  },
+});
+
+class OCIAuthStep extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = { value: "", filename: "", isLoading: false };
+    this.handleFileInputChange = (event, file) => this.setState({ filename: file.name });
+    this.handleTextOrDataChange = (value) => this.props.uploadSettings.private_key=value;
+    this.handleClear = () => this.setState({ filename: "", value: "" });
```
jkozol, line 28: This should also clear `props.uploadSettings.private_key`

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,174 @@
...
+    id: "fingerprint-help",
+    defaultMessage: "API private key's fingerprint",
+  },
+});
+
+class OCIAuthStep extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = { value: "", filename: "", isLoading: false };
+    this.handleFileInputChange = (event, file) => this.setState({ filename: file.name });
```
jkozol, line 26: I would recommend also adding a parameter to the props so that the filename is saved in the props and loaded from there. Currently, if a user changes to a different step and then comes back there is nothing showing they already uploaded a file.

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,174 @@
...
+            </div>
+            <FileUpload
+              id="private_key-input"
+              type="text"
+              name="private_key"
+              className="pf-c-form-control"
+              value={this.state.value}
+              filename={this.state.filename}
+              hideDefaultPreview="true"
+              filenamePlaceholder="Drag and drop a file or upload one"
```
jkozol, line 112: Can you mark this text for translation?

jkozol (changes requested):
Thanks for the updates! The FileUpload works for me!! Just a few minor changes but overall functions as expected :)

jkozol:
Also, for testing, you can run `npm run eslint` to test the style locally.

components/Wizard/OCIAuthStep.js
rgolangh: thanks! I was wondering why this was so ugly...

components/Wizard/OCIAuthStep.js
rgolangh: I may be missing something here - setting the state of `filename` here isn't not enough?

jkozol: So each step only renders when we are on the step. Changing steps will cause the step to unmount from the DOM. State variables will then be cleared. The whole wizard, however, does not unmount until the whole process is completed. So, any state stored int the CreateImageUpload component will not be cleared when changing steps but each step's state will be cleared. I will generally store temporary UI focused state (i.e. `isLoading`) in subcomponents and will store more important, longer-term data in the parent components or the redux store.

rgolangh: done.

components/Wizard/OCIAuthStep.js
```diff
@@ -0,0 +1,181 @@
...
+              </Popover>
+            </div>
+            <FileUpload
+              id="private_key-input"
+              type="text"
+              name="private_key"
+              value={this.state.value}
+              filename={this.state.filename}
+              hideDefaultPreview="true"
+              filenamePlaceholder=<FormattedMessage defaultMessage="Drag and drop a file or upload one" />
```
jkozol, line 119: Can you change this to something like:

```
filenamePlaceholder={formatMessage(messages.filenamePlaceholder)}

```
and add at the top
```
  filenamePlaceholder: {
    id: "filename-placeholder",
    defaultMessage: "filenamePlaceholder",
  },
});
```

jkozol:
Except for the last translation issue looks great! If you have some spare time a section on the review page would be appreciated. Otherwise I am happy to leave this PR as is and add the review step separately.

components/Wizard/OCIAuthStep.js
rgolangh: done

jkozol:
Amazing! Waiting for tests to run :)

jkozol (approved):
Looks good to me but I don't have an OCI account to verify with. @rgolangh can you try uploading through the UI before merging?

rgolangh:
> Looks good to me but I don't have an OCI account to verify with. @rgolangh can you try uploading through the UI before merging?

Tested again and again, this is working without issues. Here's an image I uploaded based on the blueprint I'm using for CI
image

PR #1376, opened 2021-11-26
test: remove unused distro checks

jkozol:
We only test against rhel 8.6, 9.0 and against fedora 34. All other distro options are removed.


achilleas-k (dismissed):

jkozol:
rhel-8-5 can be ignored since testing has been removed for it

achilleas-k (approved):

PR #1377, opened 2021-11-28
Bump cockpit test lib to fix Browser.logout() for cockpit 258

martinpitt:
The shell DOM changed in Cockpit 258. Pick up the Browser.logout()
change which can deal with both old and new shell versions.

Version 257 changed the set_input_text() API: The input line now does
not stay focused implicitly:
https://github.com/cockpit-project/cockpit/commit/9b86016ce13
Adjust tests to explicitly disable the blurring if a subsequent
key_press() wants to amend the input line value.

---

This blocks the recent image refreshes like https://github.com/cockpit-project/bots/pull/2669


martinpitt:
Triggering an extra test against the new rhel-9-0 image in https://github.com/cockpit-project/bots/pull/2669

marusak (approved):
Thanks!

PR #1399, opened 2022-02-16
Makefile: bump cockpit to newest release

jkozol:


jkozol:
@martinpitt @marusak Do either of you know why these tests are failing to start?

marusak:
Likely related: https://github.com/cockpit-project/bots/pull/2941#issuecomment-1040509775

martinpitt:
@jkozol I'll look at it ASAP, indeed see https://github.com/cockpit-project/bots/pull/2941#issuecomment-1040509775 . A recent bots change caused some incompatibility which broke c-composer, sorry. We haven't seen this anywhere else.

martinpitt:
But correct, the overlay dir must exist.

jkozol:
For some reason I can't debug this locally :( I will continue to look into it but if you have any advice it would be much appreciated.

test/run
```diff
@@ -8,6 +8,7 @@ fi
 
 # overlays are too big for bot's 10GB /tmp tmpfs
 TEST_OVERLAY_DIR="$(pwd)/tmp/run"
+mkdir -p TEST_OVERLAY_DIR
```
martinpitt, line 11: Have a dollar? :grin:

martinpitt (changes requested):
Reproduces quite fine with

    TEST_OS=rhel-8-6  make vm test/common
    TEST_OS=rhel-8-6 TEST_OVERLAY_DIR="$(pwd)/tmp/run" test/verify/check-blueprint TestBlueprint.testCreate -stv

Note that `$TEST_OVERLAY_DIR` only gets set by test/run, which as a human you'd never run. But it is what triggers this behaviour change.

With fixing the variable typo it works quite fine. I'll force-push that simple fix.

jkozol:
@martinpitt I see that cockpit added `reference-image` to testing. What configuration do I need to add to support this?

martinpitt:
Heh, I was about to push the reference-image fix -- it should be squashed into the lib update, but not a biggie..

martinpitt:
> What configuration do I need to add to support this?

@jkozol : This is for pixel testing. c-composer does not do this at the moment, so it doesn't matter what you put into the file. See https://cockpit-project.org/blog/pixel-testing.html and https://cockpit-project.org/blog/pixel-testing-update.html for details. If you are interested in that, happy to walk you through.

jkozol:
> Heh, I was about to push the reference-image fix -- it should be squashed into the lib update, but not a biggie..

I squashed it.

jkozol:
> @jkozol : This is for pixel testing. c-composer does not do this at the moment, so it doesn't matter what you put into the file. See https://cockpit-project.org/blog/pixel-testing.html and https://cockpit-project.org/blog/pixel-testing-update.html for details. If you are interested in that, happy to walk you through.

I see. I will read up on this.

martinpitt (approved):
Danke!

PR #1402, opened 2022-02-20
[r8.5] test/run: create overlay dir, fix for latest firefox

martinpitt:
The TEST_OVERLAY_DIR must be manually created per:
https://github.com/cockpit-project/bots/commit/16aa791cd

Cherry-picked from main commit dafa4d356bd173895

---

Spotted in https://github.com/cockpit-project/bots/pull/2964 - This is just a backport.

### DIFF
--- a/test/run
+++ b/test/run
@@ -8,6 +8,7 @@ fi
 
 # overlays are too big for bot's 10GB /tmp tmpfs
 TEST_OVERLAY_DIR="$(pwd)/tmp/run"
+mkdir -p $TEST_OVERLAY_DIR
 
 make check
 


martinpitt:
@jkozol : The rhel-8-5 branch [really does not like Firefox](https://logs.cockpit-project.org/logs/pull-1402-20220220-073147-ac5b2e3a-rhel-8-5-firefox/log.html), is that known?

martinpitt:
Oh, for firefox I bet we need the new magic options to counteract https://bugzilla.mozilla.org/show_bug.cgi?id=1746154 . So this will require a testlib bump

martinpitt:
Due to the fixed messages from cockpit-ws, the `assertion 'node != NULL' failed` now [happens everywhere](https://logs.cockpit-project.org/logs/pull-1402-20220221-092754-53c65c98-rhel-8-5/log.html). That is simple to fix and well-known; however, [other tests](https://logs.cockpit-project.org/logs/pull-1402-20220221-092754-53c65c98-rhel-8-5/log.html#2-2) now started to fail, so this requires more backports. At this point this becomes rather hairy -- @teg , @jkozol , how much do you actually care about the rhel-8.5 branch still? There is no opportunity any more to update it, as the next release is 8.6. (We already archived cockpit's rhel-8.5 branch, FTR)

martinpitt:
Going back to the previous state of just fixing the overlay dir. This will at least fix chromium. Let's please land this first.

jkozol:
@martinpitt I am happy to land the overlay fix. We don't really care about the 8.5 branch at this point since we will only release into 8.6.

jkozol (approved):

PR #1412, opened 2022-04-08
test: Disable TestImage.testOSTree{Commit,Container}

martinpitt:
These two tests have constantly failed for several months. They cause
blocked VM image refreshes, test timeouts (so that the other tests
often don't run), and block the infra.

----

This is a desperate measure, but these two tests currently don't help anyone. We know that either OSTree builds or the tests are broken, and it needs some maintainer effort to look at them.

See e.g. #1411 for a recent example.

### DIFF
--- a/test/verify/check-image
+++ b/test/verify/check-image
@@ -505,6 +505,7 @@ class TestImage(composerlib.ComposerCase):
         # collect code coverage result
         self.check_coverage()
 
+    @unittest.skip("FIXME: Broken")
     def testOSTreeCommit(self):
         b = self.browser
         m = self.machine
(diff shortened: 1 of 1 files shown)


martinpitt:
meh at the [failing openstack test](https://logs.cockpit-project.org/logs/pull-1412-20220408-061025-d0bc13d8-rhel-9-0/log.html), but I remember that this also happens quite often, and depends on external infra. I already retried this morning, let's try another round.

jkozol (approved):

martinpitt:
OK, testOpenstack doesn't like me today. But unlike testOSTree that at least still has some chance of getting green (I think), so let's not axe it just yet.

PR #1436, opened 2022-06-21
workflows: split up and rewrite po-refresh

martinpitt:
Take the two jobs from the current po-refresh workflow and split them
into two separate workflows with more descriptive names:

- weblate-sync-pot
- weblate-sync-po

The former now runs daily; this avoids lag for translators and has no
extra cost. The latter continues to run weekly, as that causes a PR
which needs to be reviewed, and it's generally not necessary to update
them more often.

Rework the weblate-sync-po workflow:

- we now manually clone the weblate repo and copy the files over for
ourselves.

- we use the GitHub Actions token (appropriately constrained) to push to
our repository and open the PR.

This gets rid of the last user of the `COCKPITUOUS_TOKEN` secret.

---

This is pretty much exactly the same as in [cockpit-podman](https://github.com/cockpit-project/cockpit-podman/tree/main/.github/workflows) and our other projects.

 - [x] Create the cockpit-composer-weblate environment with deploy key: https://github.com/cockpit-project/bots/pull/3533
 - [ ] After this lands, test both workflows
 - [ ] After this lands, drop the COCKPITUOUS_TOKEN secret


martinpitt:
@jkozol Unfortunately, due to how github works, these two new workflows won't appear on https://github.com/osbuild/cockpit-composer/actions until they are on main. Once they are, a PR can *change* the workflows and one can run them from the proposed branch, but not the first time.

However, these changes are well understood by now, we have run this on all our cockpit projects for some months. So, can we land these, and I commit to immediately testing both of them, and in case of some breakage, send follow-ups?

martinpitt:
Key was deployed, ready to land.

marusak (approved):

martinpitt:
I triggered [weblate-sync-pot](https://github.com/osbuild/cockpit-composer/runs/6997363800?check_suite_focus=true), which is the "complicated" one (cross-repo commit). It succeeded and [committed the POT update](https://github.com/osbuild/cockpit-composer-weblate/commit/b070ab21f21559af09162dccf13256ec79b688aa). I deleted the COCKPITUOUS_TOKEN secret, and it's gone from the primary internal secrets git as well.

martinpitt:
I triggered [weblate-sync-po](https://github.com/osbuild/cockpit-composer/runs/6997405433?check_suite_focus=true). There's nothing to update, as we just landed a PO refresh, but it succeeded.

PR #1440, opened 2022-07-14
Packages step

lucasgarfield:
This is still very much a *draft* PR, it needs a lot of refactoring and there are probably still some bugs. However, the main functionality is all in place and I do not think that any of the logic will majorly change from this point. So, the only things left to do are to add unit tests and refactor.

### DIFF
--- a/components/Wizard/CreateImageWizard.js
+++ b/components/Wizard/CreateImageWizard.js
@@ -22,6 +22,9 @@ import {
 } from "./steps";
 import "./CreateImageWizard.css";
 
+import BlueprintApi from "../../data/BlueprintApi";
+import * as composer from "../../core/composer";
+
 const CreateImageWizard = (props) => {
   const [isWizardOpen, setIsWizardOpen] = useState(false);
 
@@ -38,7 +41,12 @@ const CreateImageWizard = (props) => {
     setIsWizardOpen(true);
   };
 
-  const handleSubmit = (formValues) => {
+  const handleSubmit = async (formValues) => {
+    const result = await composer.diffBlueprintToWorkspace(props.blueprint.id);
+    if (result.diff.length !== 0) {
+      await BlueprintApi.handleCommitBlueprint(props.blueprint);
+    }
+
     // startCompose(props.blueprint.name, composeType, imageSize, ostree, upload);
 
     let uploadSettings;
@@ -127,7 +135,6 @@ const CreateImageWizard = (props) => {
                 buttonLabels: {
                   submit: "Create image",
                 },
-<<<<<<< HEAD
                 fields: [
                   imageOutput,
                   awsAuth,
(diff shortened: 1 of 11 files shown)


lucasgarfield:
Quick demo (also demos the really annoying bug we need to fix where the form resets to the first step):
cockpit packages draft pr

components/Wizard/CreateImageWizard.js
```diff
@@ -23,8 +29,11 @@ const CreateImageWizard = (props) => {
     setIsWizardOpen(true);
   };
 
-  const handleSubmit = (formValues) => {
-    console.log(formValues);
+  const handleSubmit = async (formValues) => {
+    const result = await composer.diffBlueprintToWorkspace(props.blueprint.id);
```
jkozol, line 33: Let's just default commit. I don't want to compare against the workspace.

components/Wizard/CreateImageWizard.js
```diff
@@ -5,9 +5,15 @@ import { Button } from "@patternfly/react-core";
...
 
 import ImageCreator from "./ImageCreator";
-import { imageOutput, details, review } from "./steps";
+import { imageOutput, details, packages, review } from "./steps";
 import "./CreateImageWizard.css";
 
+import BlueprintApi from "../../data/BlueprintApi";
+
+import * as composer from "../../core/composer";
+import { setBlueprint } from "../../core/actions/blueprints";
```
jkozol, line 14: This is unused. Did you intend on setting the blueprint before committing?

pages/blueprintEdit/index.js
```diff
@@ -209,6 +209,8 @@ class EditBlueprintPage extends React.Component {
     // for now, just adding the component to the state
     // component info will load after committing the change to the workspace and reloading the blueprint
     const components = this.props.blueprint.components.concat(addedComponent);
+    console.log("blueprint: ", this.props.blueprint);
+    console.log("components, packages, modules, pendingChanges: ", components, packages, modules, pendingChanges);
```
jkozol, line 213: Remove please.

jkozol (commented):
The wizard refreshes when I add a package due to `Warning: Can't perform a React state update on an unmounted component`.

components/Wizard/CreateImageWizard.js
lucasgarfield: Oof, good catch.

pages/blueprintEdit/index.js
lucasgarfield: Done.

lucasgarfield:
Closing this PR as the commit diffs are not showing up correctly. The diff shows lines from a merge conflict that are not present in the feature branch on my private fork or on my local branch, so I think it must be a bug in Github.
image

PR #1454, opened 2022-07-25
Wizard: only pass blueprint name

jkozol:
The entire blueprint is not needed by the Wizard. Passing the whole
blueprint will also trigger DOM refreshes as the blueprint contents
change. Only depending on the blueprint name prevents this as well as
reduces the passing of unnecessary data.


lucasgarfield (approved):
Good thinking, LGTM and seems to work!

PR #1461, opened 2022-08-01
Packages step - second PR

lucasgarfield:


components/Wizard/formComponents/Packages.js
```diff
@@ -0,0 +1,537 @@
...
+    const formPackages = getState()?.values?.["selected-packages"];
+
+    if (formPackages) {
+      setPackagesChosenSorted(formPackages);
+    } else {
+      fetchPackagesChosen();
+    }
+  }, []);
+
+  const updateBlueprintComponents = (packageList) => {
```
jkozol, line 76: Can we move the actual blueprint update to the final step? I would like the actual blueprint update to be on confirmation i.e. Create Image button or a Save Bluepoint button

components/Wizard/formComponents/Packages.js
```diff
@@ -0,0 +1,537 @@
...
+import api from "../../../core/api";
+import * as composer from "../../../core/composer";
+import { updateBlueprintComponents } from "../../../core/actions/blueprints";
+
+// the fields isHidden and isSelected should not be included in the package list sent for image creation
+const removePackagesDisplayFields = (packages) =>
+  packages.map((pack) => ({
+    name: pack.name,
+    summary: pack.summary,
+  }));
```
jkozol, line 29: This is a helper specifically for the Packages component so can you include it inside of the Packages function.

components/Wizard/formComponents/Packages.js
```diff
@@ -0,0 +1,537 @@
...
+      const result = await composer.depsolveBlueprint(props.blueprintName);
+      const blueprint = result.blueprints[0].blueprint;
+      const blueprintPackages = blueprint.packages.concat(blueprint.modules);
+      updateBlueprintComponents(blueprintPackages);
+      const packageNames = blueprintPackages.map((pkg) => pkg.name);
+      if (packageNames.length !== 0) {
+        const packageInfo = await composer.getComponentInfo(packageNames);
+        const selectedPackages = packageInfo.map((pkg) => ({ name: pkg.name, summary: pkg.summary }));
+        setPackagesChosenSorted(selectedPackages);
+        change(input.name, removePackagesDisplayFields(selectedPackages));
```
jkozol, line 62: no need to change the field parameter like this. The packages with name and summary isn't used elsewhere. Instead, you could set the packages form field to the list of package names. That would enable you, on the review page, to load the packages names from the form state and not the blueprint.

components/Wizard/formComponents/Packages.js
```diff
@@ -0,0 +1,537 @@
...
+      }
+      setPackagesChosenLoading(false);
+    };
+
+    const formPackages = getState()?.values?.["selected-packages"];
+
+    if (formPackages) {
+      setPackagesChosenSorted(formPackages);
+    } else {
+      fetchPackagesChosen();
```
jkozol, line 72: Can you explain this? It feels unnecessary to me.

core/api.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+async function getPackages(filter, selectedInputPage, pageSize) {
+  try {
+    const wildcardsUsed = filter.includes("*");
+    const regex = / +|, +/g;
+    let filterValue = filter.replace(regex, ",");
+    const regexStrip = /(^,+)|(,+$)/g;
+    filterValue = filterValue.replace(regexStrip, "");
+    filterValue = wildcardsUsed ? filterValue : `*${filterValue}*`.replace(/,/g, "*,*");
```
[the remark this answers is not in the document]
lucasgarfield: I don't know what the purpose behind every line is because as you pointed out, it has just been copy/pasted it so that the the search results would be the same as before. But overall the purpose is to transform the search query into a form that will actually work with the Weldr API which expects you to use things like leading and trailing wildcards. So for instance, if you want to search for "vim", you actually need to query the Weldr API "*vim*" if you want it to behave as you would expect show packages such as neovim.

components/Wizard/formComponents/Packages.js
```diff
@@ -0,0 +1,533 @@
...
+  const [packagesAvailableFound, setPackagesAvailableFound] = useState(true);
+  const [packagesAvailableLoading, setPackagesAvailableLoading] = useState(false);
+  const [packagesChosen, setPackagesChosen] = useState([]);
+  const [packagesChosenFound, setPackagesChosenFound] = useState(true);
+  const [packagesChosenLoading, setPackagesChosenLoading] = useState(false);
+  const [focus, setFocus] = useState("");
+  const [totalPackagesAvailable, setTotalPackagesAvailable] = useState(undefined);
+  const [packagesAvailablePage, setPackagesAvailablePage] = useState(1);
+  const [packagesAvailablePerPage, setPackagesAvailablePerPage] = useState(100);
+  const [additionalPackagesAvailableLoading, setAdditionalPackagesAvailableLoading] = useState(false);
```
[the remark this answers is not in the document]
lucasgarfield: Done, see #1462.

components/Wizard/formComponents/Review.js
```diff
@@ -95,6 +97,21 @@ const ociReview = (formValues) => (
...
   const { getState } = useFormApi();
   const formValues = getState()?.values;
+  const [dependencies, setDependencies] = useState(undefined);
+  const [packages, setPackages] = useState(undefined);
+
+  useEffect(() => {
+    const fetchDependencies = async (blueprintName) => {
+      const result = await composer.depsolveBlueprint(blueprintName);
+      const numPackages =
+        result.blueprints[0].blueprint.packages.length - result.blueprints[0].blueprint.modules.length;
```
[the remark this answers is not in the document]
lucasgarfield: See #1463.

However, with respect to this specific line, since we will now convert any modules to packages, modules.length should always be 0 and we can remove the subtrahend. And actually, I think I actually had intended to add them together and not subtract them!

components/Wizard/formComponents/Packages.js
lucasgarfield: Great idea. The current behavior more closely resembles the existing cockpit-composer behavior, where adding or removing packages on the `blueprintEdit` page modifies the workspace. I would actually prefer it to behave more like Image Builder Front End, which would be the result of implementing your suggestion.

components/Wizard/formComponents/Packages.js
lucasgarfield: The useEffect() this belongs to executes every time the Packages step is rendered. However, we only want to fetch the blueprint packages via API calls the *first* time the Packages step is opened. If `formPackages` is `undefined` it is the first time the Packages step has been opened so we fetch the packages via an API call in the else clause.

components/Wizard/formComponents/Packages.js
lucasgarfield: Will do!

core/api.js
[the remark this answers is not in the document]
teg: - `wildcardsUsed` is true if an `*` appears in the filter
- `regex` matches a non-empty sequence of spaces, or a comma followed by a non-empty sequence of spaces
- `filterValue` replaces each match of `regex` with a comma - in other words all spaces are replaced by commas, without inserting repeated commas
- `regexStrip` matches any non-empty sequence of commas at the beginning or end of a string
- `filterValue` is update to replace leading or trailing strings of commas with the empty string
- finally, as Lucas said, each match is replaced with wildcard versions, except if wildcards are already used.

I wonder if a future task could be to make weldr search better, if we push the search query into `dnf-json` we would get more powerful searching (making this logic redundant) and potentially much faster queries.

components/Wizard/formComponents/Packages.js
jkozol: Ahh I see. Great :)

components/Wizard/formComponents/Packages.js
lucasgarfield: Good point! This does add some additional complexity to the review page, though, since the package list is fetched using an async API call. If a user clicks 'next' on the Packages step before the packages are loaded, there is a problem... when the API call completes the form state will be set correctly but the review step won't re-render to reflect this. My solution to this is to use a <FormSpy> to keep track of changes to the form state and re-render the review page when they occur.

components/Wizard/formComponents/Packages.js
lucasgarfield: In fact, this is actually no longer necessary now that we set the packages form field to the list of package names and can just be removed.

components/Wizard/formComponents/Packages.js
lucasgarfield: Actually, after changing the form state so that it is just the list of package names, we no longer keep track of the package summaries. That means that `setPackagesChosenSorted(formPackages)` no longer works, because it expects an array of objects of form { name, summary }. 

I'm going to keep the form state as just the list of package names, and just call `fetchPackagesChosen()` every time the packages step is open.

jkozol (approved):
Overall, great work! This is good enough to merge but there are a few issues. Please add issues to track the FormSpy error. And, when adding packages, going to review page, then back to the packages step a user will see no packages added. Can you add an issue to track that as well?

jkozol:
https://user-images.githubusercontent.com/11712857/182467079-7f798eb0-e85a-4253-9eba-bd089bec5af8.mp4

lucasgarfield:
> Overall, great work! This is good enough to merge but there are a few issues. Please add issues to track the FormSpy error. 

Done, #1467 

>And, when adding packages, going to review page, then back to the packages step a user will see no packages added. Can you add an issue to track that as well?

This bug should be fixed in 95a8878.

PR #1466, opened 2022-08-02
Add ostree support

jkozol:
This PR is a MVP. For some image types the details page will be empty but instead of changing the routing I am leaving it empty since soon other blueprint customization will be added to the wizard and the routing would need to be changed back.

Also see issue https://github.com/osbuild/cockpit-composer/issues/1465

### DIFF
--- a/components/Wizard/formComponents/FormGroupCustom.js
+++ b/components/Wizard/formComponents/FormGroupCustom.js
@@ -1,0 +1,70 @@
+import React from "react";
+import { FormGroup as Pf4FormGroup, TextContent, Text } from "@patternfly/react-core";
+import PropTypes from "prop-types";
+
+const showError = ({ error, touched, warning, submitError }, validateOnMount) => {
+  if ((touched || validateOnMount) && error) {
+    return { validated: "error" };
+  }
+
+  if ((touched || validateOnMount) && submitError) {
+    return { validated: "error" };
+  }
+
+  if ((touched || validateOnMount) && warning) {
+    return { validated: "warning" };
+  }
+
+  return { validated: "default" };
+};
+
+const FormGroupCustom = ({
+  className,
+  label,
+  labelIcon,
+  isRequired,
+  helperText,
+  meta,
+  validateOnMount,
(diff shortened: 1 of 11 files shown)


teg:
Maybe not relevant for the first version, but might make sense to look at how these things are done in edge fleet managment in crc if we can steal anything from there.

components/Wizard/formComponents/FormGroupCustom.js
```diff
@@ -0,0 +1,70 @@
...
+    return { validated: "error" };
+  }
+
+  if ((touched || validateOnMount) && submitError) {
+    return { validated: "error" };
+  }
+
+  if ((touched || validateOnMount) && warning) {
+    return { validated: "warning" };
+  }
```
lucasgarfield, line 16: I can't find `meta.warning` in the react-final-form docs, just double checking that it really exists.

https://final-form.org/docs/react-final-form/types/FieldRenderProps

jkozol: https://data-driven-forms.org/schema/validator-warnings
https://www.patternfly.org/v4/components/text-input#textinput

So the warning in meta should mimic the warning option in Patternfly's form validation

jkozol:
> Maybe not relevant for the first version, but might make sense to look at how these things are done in edge fleet managment in crc if we can steal anything from there.

@teg Unfortunately they don't allow customizations and only build `RHEL for Edge Commit (.tar)` and `RHEL for Edge Installer (.iso)` image types.

I do want to steal their review page though :) I think it is very clean.

components/Wizard/formComponents/FormGroupCustom.js
jkozol: Currently this isn't used by us but will probably be useful in the future. I.e. if a user sets a strange value that won't necessarily fail but might. Image size of 10 TB or something...

lucasgarfield:
Everything looks good to me, I think we can merge!

lucasgarfield (approved):

PR #1470, opened 2022-08-03
Fix blueprint tests

jkozol:


test/verify/check-blueprint
```diff
@@ -149,8 +141,8 @@ class TestBlueprint(composerlib.ComposerCase):
         b.wait_visible("#cmpsr-modal-export")
         with b.wait_timeout(300):
             b.wait_in_text("#textInput2-modal-markup", "openssh-server")
```
lucasgarfield, line 143: I'm getting a timeout failure here.

components/Toolbar/BlueprintsToolbar.js
```diff
@@ -1,79 +1,77 @@
...
+          <SearchInput
+            aria-label="blueprints search input"
+            onChange={onInputChange}
+            value={inputValue}
+            onClear={() => onInputChange("")}
+          />
+        </ToolbarItem>
+        <ToolbarItem variant="icon-button-group">
+          {(props.sortValue === "DESC" && (
+            <Button variant="plain" aria-label="sort asc" onClick={() => props.sortSetValue("ASC")}>
```
lucasgarfield, line 35: Would it be better for screen readers if the aria label was `sort ascending`?

components/Toolbar/BlueprintsToolbar.js
```diff
@@ -1,79 +1,77 @@
...
+              <SortAlphaDownIcon />
+            </Button>
+          )) ||
+            (props.sortValue === "ASC" && (
+              <Button variant="plain" aria-label="sort desc" onClick={() => props.sortSetValue("DESC")}>
```
lucasgarfield, line 40: Same as above, `sort descending` might be better for screen readers...

components/Toolbar/BlueprintsToolbar.js
jkozol: Good point!

components/Toolbar/BlueprintsToolbar.js
jkozol: Good point!

PR #1473, opened 2022-08-04
Test: Remove check-dependence tests

lucasgarfield:
check-dependence performs tests on the Edit Packages page, which has
been removed.


lucasgarfield:
The first test, `testComponent`, checks for the presence of the dependencies on the Edit Blu

The second test, `testDetails`, adds a package and removes it. This workflow will be tested in the to-be-written `check-wizard` test.

jkozol (approved):

PR #1474, opened 2022-08-04
Toolbar: fix blueprint search

jkozol:
The blueprint list can be sorted and filtered by name. Also, the Manage
sources modal is no longer hidden by a dropdown kebab.


jkozol:
Screenshot from 2022-08-04 16-56-33
Screenshot from 2022-08-04 16-56-17

PR #1475, opened 2022-08-04
Test: Fix check-service testing

lucasgarfield:
All tests now pass.

### DIFF
--- a/test/verify/check-service
+++ b/test/verify/check-service
@@ -3,9 +3,6 @@
 # checkpoint:
 # 1. osbuild-composer.socket can be started and enabled
 # 2. no blueprint by default and can add and delete blueprint
-import string
-import random
-
 import composerlib
 import testlib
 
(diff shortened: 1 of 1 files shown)


test/verify/check-service
```diff
@@ -39,46 +36,21 @@ class TestService(composerlib.ComposerCase):
...
         # blueprint description
         b.set_input_text("#textInput2-modal-markup", bp_desc)
         # create
         b.click("#create-blueprint-modal-create-button")
-        b.wait_not_present("div[role=dialog] #cmpsr-modal-crt-blueprint")
+        b.wait_not_present("#cmpsr-modal-crt-blueprint")
 
-        # wait for all available compontents visible
+        # wait for back to blueprint link and click
         with b.wait_timeout(300):
```
jkozol, line 52: Can you remove this timeout. The standard timeout should be good enough for blueprint creation so if it is taking 5 minutes we should fail.

jkozol (changes requested):

test/verify/check-service
lucasgarfield: Done!

PR #1479, opened 2022-08-08
Wizard tests

jkozol:
This is rebased on top of #1478


jkozol:
Created an issue for the popovers. See #1480

jkozol:
The rhel-9-1 test is failing which is tracked in #1486 but this doesn't need to block these tests

PR #1481, opened 2022-08-09
test: fix testNonAdminUser test

jkozol:


jkozol:
testNonAdminUser passes so I would consider the failed tests passes since they will be fixed in the create-wizard test PR #1479

test/verify/check-service
```diff
@@ -112,12 +112,12 @@ class TestService(composerlib.ComposerCase):
         b.wait_visible("button:contains('Start')")
 
         # disable this test due to https://github.com/osbuild/cockpit-composer/issues/941
```
lucasgarfield, line 114: I realize that the issue is still open but I imagine this comment could be really confusing if left in place in a few years when we've all forgotten about the unification...

jkozol: nice catch.

lucasgarfield (approved):
LGTM!

PR #1489, opened 2022-08-10
Bugfix/wizard packages step

lucasgarfield:


components/Wizard/formComponents/Packages.js
```diff
@@ -59,7 +59,23 @@ const Packages = ({ defaultArch, ...props }) => {
       setPackagesChosenLoading(false);
     };
 
-    fetchPackagesChosen();
+    const fetchPackagesChosenFromFormState = async (packages) => {
+      if (packages.length !== 0) {
```
jkozol, line 63: This can be removed. See other comment.

components/Wizard/formComponents/Packages.js
```diff
@@ -59,7 +59,23 @@ const Packages = ({ defaultArch, ...props }) => {
...
+        const packageInfo = await composer.getComponentInfo(packages);
+        const selectedPackages = packageInfo.map((pkg) => ({ name: pkg.name, summary: pkg.summary }));
+        setPackagesChosenSorted(selectedPackages);
+        setPackagesChosenLoading(false);
+      }
+    };
+
+    const packages = getState()?.values["selected-packages"];
+
+    if (packages) {
```
jkozol, line 74: You can check `packages?.length` here. It'll check that length isn't 0 and if packages is undefined it wont error.

jkozol (approved):
I'm approving my suggestion is such a small nitpick. Everything works as expected.

components/Wizard/formComponents/Packages.js
lucasgarfield: Good point! Done!

components/Wizard/formComponents/Packages.js
lucasgarfield: In this case, we are really checking whether it is undefined or not. If it is undefined, it is the first time the packages have been requested so they should be loaded from the blueprint. If it exists - regardless of the length - then the packages have already been fetched from the blueprint, so we should be using the form state. 

Would it be better if it was more explicit, i.e. `if (packages === undefined)`?

PR #1490, opened 2022-08-10
Wizard: Fix number of dependencies shown on Review step

lucasgarfield:
The number of dependencies is determined by depsolving the blueprint.
The array of dependencies does not include the blueprint packages, so
the number of packages should not be subtracted from the number of
dependencies.


lucasgarfield:
One thing to think through - what should the behavior be if one of the packages in the blueprint's [[Packages]] is a dependency of one of the other packages in the blueprint's [[Packages]] list?

jkozol (approved):

Issue #1471, opened 2022-08-04
Unification: Wizard review step should have an update blueprint button


lucasgarfield:
At the moment, the only way to update a blueprint is to click the 'Create image' button on the Wizard's review step. It would be nice to be able to update the blueprint without building an image. The review step could have two buttons, 'Create image' and 'Update blueprint', both with tooltips to explain the differences in behavior.

It would be straightforward to do this with the Patternfly Wizard as it accepts a `footer` prop where custom buttons can be defined but the Data Driven Forms version of the PF Wizard does not seem to accept the `footer` prop. I created a discussion on the DDF project's discussion page to ask for clarification about whether it is possible: https://github.com/data-driven-forms/react-forms/discussions/1297#discussion-4276151

PR #1491, opened 2022-08-11
Wizard: Add update blueprint button

lucasgarfield:
This commit adds a button on the review step that allows the user to
update a blueprint without building an image.


lucasgarfield:
See #1471 for additional info about this PR.

PR #1499, opened 2022-08-15
Wizard save blueprint

jkozol:


lucasgarfield:
I still don't have my head around all of the Redux code in the codebase yet, but I've done a surface level review/sanity check as well as manual testing and I think this looks great!

PR #1502, opened 2022-08-16
Add customizations to the Wizard

jkozol:
Migrate hostname and installation device to the Wizard. Also cleanup the Wizard steps a little.

This is rebased ontop of #1499


lucasgarfield:
Little bug on the `image output` step, it looks like the component doesn't get the state of the type from the store so if you go to the next step and then click back to the `image output` step the select will incorrectly show "select output type":
image output bug

components/Wizard/steps/customizations.js
```diff
@@ -0,0 +1,20 @@
...
+  name: "customizations",
+  nextStep: "packages",
+  fields: [
+    {
+      component: "text-field-custom",
+      name: "customizations-hostname",
+      className: "pf-u-w-75",
+      type: "text",
+      label: "Hostname",
+      helperText: "If no hostname is provided, the hostname will be determined by the OS",
```
lucasgarfield, line 17: Super nitpick but this sentence needs a period at the end.

pages/blueprint/index.js
```diff
@@ -393,27 +393,11 @@ class BlueprintPage extends React.Component {
             <div className="tab-container row">
               <div className="form-customizations col-sm-12 pf-u-ml-md">
                 <div className="form-horizontal">
-                  <div
-                    id="input-hostname"
-                    className={`form-group ${editHostnameInvalid ? "has-error" : ""}`}
-                    data-form="hostname"
-                    onSubmit={() => this.handleEditHostname("commit")}
-                  >
+                  <div className="form-group" id="input-hostname">
```
lucasgarfield, line 376: Why is this `id="input-hostname"` and not simply `id="hostname"`?

lucasgarfield (approved):
Looks great!

components/Wizard/steps/customizations.js
jkozol: fixed :)

pages/blueprint/index.js
jkozol: this id is just a leftover from when these divs where text inputs.

lucasgarfield (approved):

PR #1511, opened 2022-08-17
wizard: mark strings for translation

chloenayon:


components/Wizard/CreateImageWizard.js
```diff
@@ -50,6 +51,16 @@ const CreateImageWizard = (props) => {
     setHasSaved(true);
   };
 
+  const intl = useIntl();
```
jkozol, line 54: this should be placed inside the functional component so it rerenders.

components/Wizard/CreateImageWizard.js
```diff
@@ -50,6 +51,15 @@ const CreateImageWizard = (props) => {
...
+
+  const messages = defineMessages({
+    createImage: {
+      id: "wizard.create.image",
+      defaultMessage: "Create image",
+    },
+  });
+
```
jkozol, line 62: We can define messages outside of the functional component. This reduces clutter as well as can prevent messages from rerendering especially in memoized components.

components/Wizard/CreateImageWizard.js
```diff
@@ -181,9 +191,9 @@ const CreateImageWizard = (props) => {
         variant="secondary"
         onClick={handleOpen}
         isDisabled={!props.blueprint?.name || !props.imageTypes?.length}
-        aria-label="Create image"
+        aria-label={intl.formatMessage(messages.createImage)}
       >
-        Create image
+        <FormattedMessage id="wizard.createImage.button" defaultMessage="Create image" />
```
jkozol, line 196: You can use the same id here as in `messages.createImage`

components/Wizard/formComponents/FormGroupCustom.js
```diff
@@ -45,7 +46,9 @@ const FormGroupCustom = ({
   >
     {description && (
       <TextContent>
-        <Text component="small">{description}</Text>
+        <Text component="small">
+          <FormattedMessage id="formgroup.text" defaultMessage={description} />
+        </Text>
```
jkozol, line 51: Variables shouldn't be marked for translation since they can't be translated. Only the original strings should be marked.

components/Wizard/formComponents/ImageOutputSelect.js
```diff
@@ -19,6 +20,15 @@ const ImageOutputSelect = ({ label, isRequired, ...props }) => {
...
 
+  const intl = useIntl();
+
+  const messages = defineMessages({
+    outputType: {
+      id: "imageoutput.select.type",
+      defaultMessage: "Select output type",
+    },
+  });
+
```
jkozol, line 31: Same as other `messages` comment

package.json
```diff
@@ -21,7 +21,7 @@
     "prop-types": "15.8.1",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-intl": "5.25.1",
+    "react-intl": "^5.25.1",
```
jkozol, line 24: No need to add the carot operator here.

jkozol (commented):
Thanks for the contribution. I have included this commit as part of #1516

PR #1515, opened 2022-08-18
Wizard: add user customization fields

jkozol:
Users can be created in the Wizard now. This has been made a substep of
Customizations and the existing customizations step is now the Details
step which is also a substep.

The UI is rough and a MVP but it works.


components/Wizard/CreateImageWizard.js
```diff
@@ -151,6 +152,18 @@ const CreateImageWizard = (props) => {
...
       formState["customizations-hostname"] = blueprint.customizations.hostname;
       formState["customizations-install-device"] = blueprint.customizations.installation_device;
+      formState["customizations-users"] = [];
+      if (blueprint.customizations.user?.length) {
+        blueprint.customizations.user.forEach((user) => {
+          const formUser = {
+            name: user.name,
+            password: user.password,
+            "is-admin": user.groups?.includes("wheel"),
+            "ssh-key": user.key,
```
lucasgarfield, line 162: Not really a problem but for my own knowledge... I suppose `"is-admin"` and `"ssh-key"` are strings because otherwise you can't have a `-` in a property name, but why not just use camelCase? I think with the kebab-case we won't be able to access these properties using dot notation, right?

lucasgarfield:
There seems to be a bug related to the steps for Fedora IoT Commit images:
1515 bug

components/Wizard/CreateImageWizard.js
lucasgarfield: Done to be consistent with Data Driven Forms.

lucasgarfield (dismissed):
Looks good to me, will be a good starting point for future refinements.

PR #1517, opened 2022-08-18
wizard: translate image output step

jkozol:
The image output step is now a function which returns a json schema. This functional component can use the useIntl() hook to connect the step to the translations provider.


components/Wizard/CreateImageWizard.js
```diff
@@ -238,7 +238,7 @@ const CreateImageWizard = (props) => {
                   submit: intl.formatMessage(messages.createImage),
                 },
                 fields: [
-                  imageOutput,
+                  imageOutput(),
```
lucasgarfield, line 241: Incredible.

lucasgarfield (approved):
LGTM

PR #1530, opened 2022-08-19
wizard: fix the supported customizations for ostree image types

jkozol:
The edge-raw-image supports the Parent not the URL setting.

The edge-installer only supports User customizations.


lucasgarfield (approved):
LGTM

PR #1540, opened 2022-08-23
core: fix blueprintsGetAll for when there are no blueprints

jkozol:
The blueprints/info route should not be called with an empty list of blueprint names. This will return an error which will cause us to load an error state instead of the No blueprints state.


lucasgarfield (approved):
LGTM

PR #1555, opened 2022-09-22
Lint simplification

jkozol:


jkozol:
@lucasgarfield If you could just check that this builds for you.

lucasgarfield:
> @lucasgarfield If you could just check that this builds for you.

`make` is successful my machine!

jkozol:
@lucasgarfield seems to be a subdependency thing. I plan on using a package-lock and `npm ci` workflow like we do in image-builder-frontend once the webpack and dependencies are cleaned up more.

PR #1600, opened 2022-11-03
.github: update workflows

jkozol:
Test workflows are now labelled with "check-" so as to differentiate them from general workflows. Also, add an automerge feature for dependabot since if PRs pass testing the dependencies are fine.


croissanne (approved):
lgtm!

PR #1744, opened 2023-01-24
Add support for gcp uploads

jkozol:


src/components/Wizard/CreateImageWizard.js
```diff
@@ -142,6 +144,17 @@ const CreateImageWizard = (props) => {
...
           },
         };
+      } else if (formValues?.image?.type === "gce") {
+        uploadSettings = {
+          image_name: formValues.image.upload.image_name,
+          provider: "gcp",
+          settings: {
+            region: formValues.image.upload.settings.region,
+            bucket: formValues.image.upload.settings.bucket,
+            object: formValues.image.upload.settings.object,
```
thozza, line 154: I should update the osbuild guides, but the `object` is no longer required to be provided. It can be, but IMO it does not make much sense. The reason is that the object name is used only temporarily for the image when uploaded to GCP Storage for importing and the object is afterwards deleted. Composer will generate a unique object name if it was not provided.

The relevant change is https://github.com/osbuild/osbuild-composer/pull/3023

src/forms/steps/gcp.js
```diff
@@ -0,0 +1,174 @@
...
+  regionPopoverAria: {
+    defaultMessage: "Region help",
+  },
+  objectPopoverBody: {
+    defaultMessage:
+      "The object is the name of an intermediate storage object. It must not exist before the upload, and it will be deleted when the upload process is done. If the object name does not end with .tar.gz, the extension is automatically added to the object name.",
+  },
+  objectPopoverAria: {
+    defaultMessage: "Object help",
+  },
```
thozza, line 35: Probably best to delete this based on my other comment.

src/constants.js
```diff
@@ -21,6 +21,7 @@ export const ImageTypeLabels = {
   "edge-simplified-installer": "RHEL for Edge Simplified Installer (.iso)",
   vhd: "Microsoft Azure (.vhd)",
   vmdk: "VMWare VSphere (.vmdk)",
+  gce: "Google Cloud Platform (.gce)",
```
thozza, line 24: The image actually has `.tar.gz` extension.

thozza (commented):
Thanks a lot for implementing the support for GCE images!

Looks good from my PoV to the extent to which I can read JavaScript 😅 

I think that the `object` does not need to be provided by the user. My bad, I should have updated the guides... Alternatively it can be made optional, instead of being completely removed...

One additional comment WRT the credentials file, which may be good to discuss with UX. My hope was that the user would just provide credentials as downloaded from GCP in the JSON format and cockpit-composer would encode the content of the file as Base64 before calling Weldr API. I think that it would simplify the user experience by not requiring the user to do additional step (encoding the file in base64). Now it looks as if the user has to first download JSON file with credentials from GCP, encode the file content in base64 and save it as a file and lastly upload that encoded file to cockpit-composer.

src/constants.js
jkozol: Fixed

Screenshot from 2023-01-30 13-22-52

src/components/Wizard/CreateImageWizard.js
jkozol: Ok. Ya I removed this.

jkozol:
@thozza Cockpit composer now expects the plain json file and will encode it before passing to osbuild-composer.

lucasgarfield (approved):

PR #1750, opened 2023-01-30
steps: Add validation to the username input

### LINKED ISSUE #1706: Validate the username input

regexowl:
Fixes #1706.

This adds validation to the username input. The pattern is consistent with `man` page of `useradd`. The username can begin with a lower case letter or an underscore, and can only contain lower case letters, digits, underscores, or dashes.

### DIFF
--- a/src/forms/steps/users.js
+++ b/src/forms/steps/users.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { defineMessages, FormattedMessage } from "react-intl";
+import { validatorTypes } from "@data-driven-forms/react-form-renderer";
 
 const messages = defineMessages({
   customizationsStepTitle: {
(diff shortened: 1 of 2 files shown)


src/forms/steps/users.js
```diff
@@ -44,6 +45,15 @@ const users = (intl) => {
               {
                 type: "required",
               },
+              {
+                type: validatorTypes.PATTERN,
+                pattern: "^[a-z_][a-z0-9_-]*$",
+                message:
```
jkozol, line 51: Good but needs to be marked for translation.

jkozol (changes requested):

src/forms/steps/users.js
regexowl: Marked. I've used `intl.formatMessage` because an attempt to use `<FormattedMessage>` lead to this DDF error https://github.com/data-driven-forms/react-forms/issues/1352

jkozol (approved):

PR #1751, opened 2023-01-30
BlueprintWizard: Fix blueprint name validation

### LINKED ISSUE #1698: Disallow creating blueprint names with spaces

lucasgarfield:
Fixes  #1698.

Adds validation for blueprint name on details step. Blueprint name must not have spaces and there must not be an existing blueprint with the same name.


src/forms/validators/blueprintNameValidator.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable react/display-name */
+import React from "react";
+import { FormattedMessage } from "react-intl";
+import { selectAllBlueprintNames } from "../../slices/blueprintsSlice";
+import store from "../../store";
+
+const blueprintNameValidator = () => (value) => {
+  const state = store.getState();
+  const blueprintNames = selectAllBlueprintNames(state);
```
jkozol, line 9: Can you move this into the `BlueprintWizard` component and pass it into the validator. Currently, every key press will call the store state.

Also, you can use the `useSelector` hook instead of importing the store. 

```
const blueprintNames = useSelector((state) => selectAllBlueprintNames(state));
```

jkozol (changes requested):
See above :)

src/forms/validators/blueprintNameValidator.js
lucasgarfield: Good idea, didn't realize you could pass data into DDF validators!

jkozol: Ya I'm not sure if you pass it directly to the validator or to the form state which you access from the validator. Docs aren't the clearest so you might need to play around :D

src/components/Wizard/BlueprintWizard.js
```diff
@@ -61,6 +68,10 @@ const BlueprintWizard = (props) => {
     useSelector((state) => selectAllImageTypes(state));
   const imageTypes = getImageTypes();
 
+  const getBlueprintNames = () =>
+    useSelector((state) => selectAllBlueprintNames(state));
+  const blueprintNames = getBlueprintNames();
```
jkozol, line 73: A shorter way of doing this is just 
`const blueprintNames = useSelector((state) => selectAllBlueprintNames(state));`

jkozol: The image types call above is excessive and I should fix it.

jkozol (approved):

PR #1759, opened 2023-01-31
Table: Set the default sort of images to "Created"

### LINKED ISSUE #1696: Set the default sort to "Created" on images list

regexowl:
Fixes #1696.

This sets the default sort of the images table to sorting by `job_created` column in descending order.


src/components/Table/ImageTable.js
```diff
@@ -171,8 +171,8 @@ const ImageTable = (props) => {
   const intl = useIntl();
   // start sorting by created date
   const [activeSortIndex, setActiveSortIndex] = useState(null);
```
jkozol, line 173: You should also set the activeSortIndex to something related to the "Created" column. I think you could combinw both `fieldId` and `culumnIndex` in the sort handlers.

jkozol (changes requested):
The sort is correct but it isn't set as active in the headers.

src/components/Table/ImageTable.js
regexowl: I've updated the `activeSortIndex`

regexowl:
I've set the sort to active in the headers. I was trying to look into changing the sort arrow so it indicates the direction as well, but so far had no luck.

src/components/Table/ImageTable.js
```diff
@@ -170,9 +170,10 @@ ImageRow.propTypes = {
 const ImageTable = (props) => {
   const intl = useIntl();
   // start sorting by created date
-  const [activeSortIndex, setActiveSortIndex] = useState(null);
-  const [isSortAscending, setIsSortAscending] = useState(true);
-  const [sortBy, setSortBy] = useState("id");
+  const [activeSortIndex, setActiveSortIndex] = useState("job_created");
```
jkozol, line 173: So if you set this to `useState(3)` it will set the sort index as expected. The defaultSort code isn't necessary imo.

jkozol (changes requested):

src/components/Table/ImageTable.js
regexowl: Without the check of the `defaultSort` the Created column remains as selected even after selecting another sort index
Screenshot from 2023-02-01 14-18-08

Renamed to `initialSort` as that name might be a better fit.

src/components/Table/ImageTable.js
```diff
@@ -170,9 +170,9 @@ ImageRow.propTypes = {
 const ImageTable = (props) => {
   const intl = useIntl();
   // start sorting by created date
-  const [activeSortIndex, setActiveSortIndex] = useState(null);
-  const [isSortAscending, setIsSortAscending] = useState(true);
-  const [sortBy, setSortBy] = useState("id");
+  const [activeSortIndex, setActiveSortIndex] = useState(3);
+  const [isSortAscending, setIsSortAscending] = useState(false);
+  const [sortBy, setSortBy] = useState(3);
```
jkozol, line 175: This should still be "job_created". The setup is a little weird here but basically... the `activeSortIndex` is used by the table headers for UI purposes, `isSorbtAscending` is used in both the headers and the sorting function, `sortBy` is used only in the sorting function.

regexowl: Oh ok, it's a bit confusing, but I think I understand it now. Let's give the review one more try

jkozol (approved):

PR #1765, opened 2023-02-01
Makefile: allow testing with the Python bridge

jelly:
The Python bridge is the new cockpit-bridge rewritten in Python. Before Cockpit switches it's package over we want to be able to test if all projects are supported.


martinpitt:
"SystemError: Unsupported browser: pybridge" -- @jelly  you know the drill :grin:

martinpitt (approved):
Dankjewel!

PR #1767, opened 2023-02-02
components: Ensure deleting a blueprint also deletes any associated images

### LINKED ISSUE #1694: Ensure that deleting a blueprint also deletes any associated images

regexowl:
Fixes #1694.

This adds a deletion of any associated images when a blueprint is deleted.

Copy for the modal was also updated to inform the user that the images will be deleted together with the blueprint.

### DIFF
--- a/src/components/Modal/DeleteBlueprint.js
+++ b/src/components/Modal/DeleteBlueprint.js
@@ -1,14 +1,16 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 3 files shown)


jkozol (changes requested):
This works as expected :)

But can you create a filtered thunk for deletion in the images. That way we can dispatch an action to delete filtered images more generally. This also would mean the `deleteDlueprint` thunk will only have to dispatch a single action instead of loop dispatching actions.

regexowl:
@jkozol Updated. Please let me know if it's similar to what you had in mind.

regexowl:
@jkozol This PR might also be a bit stale. Should I close it?

regexowl:
Closing this PR as stale

PR #1769, opened 2023-02-02
SourceCard: allow entering RHSM bool

### LINKED ISSUE #1388: Allow specifying requirement to use RHSM for new source
### Description of problem
When adding new sources (repository) using the cockpit-composer, there is no way to specify that the added source requires RHSM subscription to be accessed. This mean that e.g. for RHEL, the customer can not add any additional source (e.g. for a layered product), which would be then usable for building images. Such sources can be added only using the CLI or they must be modified using CLI in case they were added using the cockpit-composer UI.

### Component (web, API, etc)
web

latest RHEL-8.6

### How often reproducible
always

### Steps to Reproduce

1. Add new source with URL from RH CDN e.g. "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/openstack/16.2/os"
2. Try build an image
3.

### Actual results
Image build fails, because the newly added source can not be accessed without RHSM credentials

### Expected results
The cockpit Image Builder UI should allow checking that the added source needs RHSM credentials to be accessed. This should be a simple true/false configuration.

### Additional info
One has to do the following to use the source:
```shell
# added "rhel_openstack" repo using Web UI
# print source info on the CLI
[root@ci-vm-10-0-138-15 ~]# composer-cli sources info rhel_openstack
check_gpg = false
check_ssl = true
id = "rhel_openstack"
name = "rhel_openstack"
rhsm = false
system = false
type = "yum-baseurl"
url = "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/openstack/16.2/os"

# save the sources config into a file
root@ci-vm-10-0-138-15 ~]# composer-cli sources info rhel_openstack > rhel_openstack.toml

# Edit the "rhsm" option to "true"

# Update the sources using CLI
[root@ci-vm-10-0-138-15 ~]# composer-cli sources change rhel_openstack.toml

# verify the source configuration
[root@ci-vm-10-0-138-15 ~]# composer-cli sources info rhel_openstack
check_gpg = false
check_ssl = true
id = "rhel_openstack"
name = "rhel_openstack"
rhsm = true
system = false
type = "yum-baseurl"
url = "https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/openstack/16.2/os"
```

jkozol:
This seems to be broken on composer. I create a source with an object of:
```
{
rhsm: true
...sourceParams
}
```
No error is reported on creation but on fetch from `projects/source/info/` `rhsm` is set to false.

Can you confirm @bcl @thozza :)

Fixes #1388


bcl:
Looks like a bug of some sort, it looks like it should be working, store has it copied over, so the state should persist.

regexowl (approved):
Looks good

PR #1771, opened 2023-02-02
Table: fix download button for logs

### LINKED ISSUE #1683: Fix file download issue that only works on Chrome, not Firefox

jkozol:
The Download logs button now works as the images one does. It uses cockpit to transport the logs from weldr api to the users browser.

Fixes #1683


regexowl (approved):
LGTM :+1:

PR #1772, opened 2023-02-02
Wizard: enable clicking on sidebar

### LINKED ISSUE #1687: Implement a way for the sidebar to skip for ddf
For blueprint creation, only the blueprint name is required. The ddf blueprint wizard should not restrict clicking on steps in the sidebar since no fields are required beyond the name.

jkozol:
The sidebar is now clickable with valid fields. This requires having a maxStepIndex and a list of all step names. This is because of ddf.

Also, the form values may not be included in the onSubmit when clicking through the wizard so the formApi may need to be used to get the form state.

Fixes #1687


regexowl (approved):
This is amazing!

PR #1884, opened 2023-06-06
Move to a webpack module, update cockpit test API to 290

jelly:

PR #1885, opened 2023-06-08
Update shared cockpit files for Chromium 113 fix

jelly:
The most recent Chromium 113 does not work any more with sizzle, tests were hanging as soon as they switched to a frame. The latest cockpit testlib applied a workaround [1]. As the latest Cockpit library also moved to PatternFly 5, but that would be prohibitively expensive for cockpit-composer's main branch, switch to the (temporary) "274+chromium-sizzle" branch, which has that workaround backported.

Alternative approach of https://github.com/osbuild/cockpit-composer/pull/1884 so PR's can be landed at least.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ machine: bots
 
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 test/common:
-	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 278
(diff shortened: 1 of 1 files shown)


jelly:
@jkozol please take a look

martinpitt (approved):
Thanks @jelly ! Let's get this in to unblock composer and image refreshes, and I'll take a look at #1884 with less pressure.

PR #1921, opened 2023-08-28
ImportBlueprint: scope file preview min-height to just this component

ondrejbudai:
823c4f13a8fa9e6f808d82cf6e60e67d5edd55f1 introduced min-height to all file upload previews. However, we don't want to have min-height for all of them, and especially not for those that have hideDefaultPreview="true". For these, having the "global" min-height causes 30rem of empty vertical space to appear, which breaks certain stuff. An example is the UploadOCIFile component used in Oracle Cloud uploads.

Fix this by scoping the CSS rule to just the ImportBlueprint component.


lavocatt (approved):
Lgtm

PR #1965, opened 2024-02-16
App.js: add polling for blueprints and images (COMPOSER-2153)

jkozol:
Fetch blueprints and images every 30 seconds in case of changes via the cli


ochosi (approved):
Tried this out locally - love the simplicity of the fix.

PR #1967, opened 2024-02-21
test: update rhel repo overrides

jkozol:
The centos stream repo declarations still work as expected but the rhel repo overrides need to be declared and updated.


martinpitt (approved):
Thanks! Landing in lockstep with https://github.com/cockpit-project/bots/pull/5945

PR #1969, opened 2024-02-22
src: add other fields card modal (RHEL-4591)

jkozol:
When a user enters an installation_device or hostname they are able to view and edit the fields from the blueprint customizations tab.


lucasgarfield:
Running this locally, on my machine the new card appears in kind of a weird spot (seems strange that the first column has two cards and the third column has four cards) and it is missing the divider:
image

PR #1991, opened 2024-03-06
test: bump to Fedora 40

supakeen:
A Fedora 37 machine was used for testing, Fedora 37 has been EOL for a while and is no longer available in cockpit-machines. Let's bump it to Fedora 40 which is currently branched but should go beta soon.


jkozol (approved):
Looks good. Hopefully the tests go smooth :)

PR #1997, opened 2024-04-10
test - do not merge

lucasgarfield:
According to discussion on Cockpit issue, this test should pass if `<tr>` is used instead of `<li>`. That's very peculiar because I can see in my browser it is a `<li>`.

https://github.com/cockpit-project/cockpit/issues/20138#issuecomment-1982817505


lucasgarfield:
@jkozol what's up with `Pending — Manual testing required`?

PR #2010, opened 2024-05-15
helpers: remove empty customization fields [COMPOSER-2204]

jkozol:
Customizations that contain empty arrays or objects instead of being undefined cause failed image builds even with a valid blueprint.

The clean function recursively removes this empty fields from the blueprint object and fixes the edge image build failures.


mvo5 (commented):
I have no context to the project so take what I say with a huge grain of salt - I wonder if this should have a test?

PR #2026, opened 2024-07-30
blueprintwizard: remove empty customizations (RHEL-28534)

kingsleyzissou:
This commit sanitizes the form data and makes sure that the values of
the form fields are truly empty. For example, the filesystem customization
option was being initialized to an empty array. This was causing `edge-commit`
and `edge-commit` image types to break since they are not allowed
filesystem customizations.

Additionally, this commit ensures that all the fields are undefined if they have
not been set. This ensures that the blueprint is saved correctly and the below
no longer happens:

```
[customizations]
[customizations.kernel]
append = ""
[customizations.timezone]
[customizations.locale]
[customizations.firewall]
[customizations.firewall.services]
[customizations.services]
[customizations.fdo]
[customizations.ignition]
[customizations.ignition.firstboot]
```

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -49,19 +49,61 @@ export const blueprintToFormState = (blueprint) => {
 
 const formStateToCustomizations = (customizations) => {
   // the form state of these matches the api state
-  const {
-    hostname,
-    kernel,
-    sshkey,
-    group,
-    timezone,
-    locale,
-    services,
-    installation_device,
-    fdo,
-    firewall,
-    ignition,
-  } = customizations;
+  const { hostname, sshkey, group, installation_device } = customizations;
+
+  let fdo;
+  if (
+    customizations?.fdo?.diun_pub_key_hash ||
+    customizations?.fdo?.diun_pub_key_insecure ||
+    customizations?.fdo?.diun_pub_key_root_certs ||
+    customizations?.fdo?.manufacturing_server_url
+  ) {
+    fdo = customizations.fdo;
+  }
+
+  let firewall;
+  if (
+    customizations?.firewall?.ports?.length > 0 ||
+    customizations?.firewall?.zones ||
+    customizations?.firewall?.services?.enabled?.length > 0 ||
+    customizations?.firewall?.services?.disabled?.length > 0
+  ) {
+    firewall = customizations.firewall;
+  }
+
+  let ignition;
+  if (customizations?.ignition?.firstboot?.url) {
+    ignition = customizations.ignition;
+  }
+
+  let kernel;
+  if (customizations?.kernel?.name || customizations?.kernel?.append) {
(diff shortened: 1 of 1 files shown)


src/components/Wizard/BlueprintWizard.js
```diff
@@ -87,6 +87,20 @@ const BlueprintWizard = (props) => {
         ? formValues
         : formApi.getState().values;
     const blueprintData = formStateToBlueprint(formState);
+
+    // when customizations are provided, the users and filesystem customizations
+    // are initialized to an empty array. This creates issues for image types that
+    // aren't allowed these customizations
+    if (blueprintData.customizations) {
```
kingsleyzissou, line 94: Ideally we would make sure that this isn't set in the first place. But the data driven forms state is a bit awkward

src/components/Wizard/BlueprintWizard.js
```diff
@@ -87,6 +87,20 @@ const BlueprintWizard = (props) => {
...
+      if (blueprintData.customizations.filesystem?.length === 0) {
+        delete blueprintData.customizations.filesystem;
+      }
+
+      if (blueprintData.customizations.user?.length === 0) {
```
kingsleyzissou, line 99: We may actually need to check some other fields such as services, locale etc if the user has stepped through those form steps, some of those values might be empty too.

kingsleyzissou: Hmm I've checked the other steps and we get this when I click through each step but don't edit anything:

```
[customizations]
[customizations.kernel]
append = ""
[customizations.timezone]
[customizations.locale]
[customizations.firewall]
[customizations.firewall.services]
[customizations.services]
[customizations.fdo]
[customizations.ignition]
[customizations.ignition.firstboot]
```

However, it seems like the only one that was causing issues was the filesystem customizations.

kingsleyzissou:
This PR fixes both:
- RHEL-4668
- RHEL-28534

But RHEL-28534 is longer standing and was initially a blocker to the RHEL-9.4 release

src/helpers.js
```diff
@@ -111,7 +155,15 @@ const formStateToCustomizations = (customizations) => {
     services,
   };
 
-  return customizationsParsed;
+  // if a user steps through each component in the form without filling any
+  // of the steps out, there is a possibility that each form field is empty.
+  // if this is the case, we want to return an empty (or undefined) `customizationsParsed`
+  // object
+  const everythingUndefined = Object.keys(customizationsParsed).every(
```
kingsleyzissou, line 162: I could simplify this by only initialising the `customizationsParsed` if a field is not empty... but this was a quick win

regexowl (approved):
Haven't tested this locally, but the changes make sense and the tests are passing :sweat_smile:

PR #2027, opened 2024-07-31
createimagewizard: new edge types (RHEL-4689)

kingsleyzissou:
Add the `edge-ami` and `edge-vsphere` image types. These image types require both their respective upload settings as well as the `ostree-settings`.

### DIFF
--- a/src/components/Wizard/CreateImageWizard.js
+++ b/src/components/Wizard/CreateImageWizard.js
@@ -93,7 +93,10 @@ const CreateImageWizard = (props) => {
 
     let uploadSettings;
     if (formValues?.image?.isUpload) {
-      if (formValues?.image?.type === "ami") {
+      if (
+        formValues?.image?.type === "ami" ||
+        formValues?.image.type === "edge-ami"
+      ) {
         uploadSettings = {
           image_name: formValues.image.upload.image_name,
           provider: "aws",
@@ -116,7 +119,8 @@ const CreateImageWizard = (props) => {
         };
       } else if (
         formValues?.image?.type === "vmdk" ||
-        formValues?.image?.type === "ova"
+        formValues?.image?.type === "ova" ||
+        formValues?.image?.type === "edge-vsphere"
       ) {
         uploadSettings = {
           image_name: formValues.image.upload.image_name,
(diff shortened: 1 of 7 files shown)


src/components/Wizard/CreateImageWizard.js
```diff
@@ -93,7 +93,10 @@ const CreateImageWizard = (props) => {
 
     let uploadSettings;
     if (formValues?.image?.isUpload) {
-      if (formValues?.image?.type === "ami") {
+      if (
+        formValues?.image?.type === "ami" ||
+        formValues?.image.type === "edge-ami"
```
kingsleyzissou, line 98: Looking at our integration tests, it doesn't look like we upload directly to aws for this image type.
https://github.com/osbuild/osbuild-composer/blob/dc389eaa71936790349577b81281fae7ff4c5d9e/test/cases/ostree-ami-image.sh#L637

src/components/Wizard/CreateImageWizard.js
```diff
@@ -116,7 +119,8 @@ const CreateImageWizard = (props) => {
         };
       } else if (
         formValues?.image?.type === "vmdk" ||
-        formValues?.image?.type === "ova"
+        formValues?.image?.type === "ova" ||
+        formValues?.image?.type === "edge-vsphere"
```
kingsleyzissou, line 123: Same as the above here:
https://github.com/osbuild/osbuild-composer/blob/dc389eaa71936790349577b81281fae7ff4c5d9e/test/cases/ostree-vsphere.sh#L491

src/constants.js
```diff
@@ -19,6 +19,8 @@ export const ImageTypeLabels = {
   "edge-raw-image": "RHEL for Edge Raw Image (.raw.xz)",
   "image-installer": "RHEL Installer (.iso)",
   "edge-simplified-installer": "RHEL for Edge Simplified Installer (.iso)",
+  "edge-ami": "RHEL for Edge AMI (.raw)",
+  "edge-vsphere": "RHEL for Edge VMware VSphere (.vmdk)",
```
regexowl, line 23: Looking at https://www.vmware.com/products/cloud-infrastructure/vsphere the capitalisation should probably look like this.

```suggestion
  "edge-vsphere": "RHEL for Edge VMware vSphere (.vmdk)",
```

regexowl (commented):
One small nitpick, otherwise looks good!

src/constants.js
kingsleyzissou: Thanks!! Good catch

kingsleyzissou: It's capitalised incorrectly in other places, so I've fixed those too.

regexowl: Excellent, thanks!

regexowl (approved):
LGTM!

PR #2030, opened 2024-08-08
packit: build RPMs in COPR for commits to `main`

kingsleyzissou:
Make COPR rpm builds for any commits that land in main. This will allow us to test the project more easily.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -15,8 +15,9 @@ srpm_build_deps:
 
 jobs:
   - job: copr_build
-    trigger: release
-    metadata:
+    trigger: commit
+    metadata: &build_metadata
+      branch: main
       owner: "@osbuild"
       project: "cockpit-composer"
       preserve_project: True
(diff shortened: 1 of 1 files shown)


croissanne (approved):

packit.yaml
```diff
@@ -26,14 +27,19 @@ jobs:
...
       # HACK: tarball for releases (copr_build, koji, etc.), copying spec's Source0; this
       # really should be the default, see https://github.com/packit/packit-service/issues/1505
       create-archive:
         - sh -exc "curl -L -O https://github.com/osbuild/cockpit-composer/releases/download/${PACKIT_PROJECT_VERSION}/${PACKIT_PROJECT_NAME_VERSION}.tar.gz"
         - sh -exc "ls ${PACKIT_PROJECT_NAME_VERSION}.tar.gz"
 
+  - job: copr_build
+    trigger: release
+    metadata: *build_metadata
+    actions: *build_actions
```
thozza, line 41: Just note that this may be unnecessary. The commit that corresponds to the release would be anyway already built in the repo from `main`, because releases are based on tags (so no new commit). In addition, the release build would be superseded by the next commit on `main` and its build.

So not a problem, it just IMHO adds very little value.

kingsleyzissou: Okay thanks for that input, I wasn't a 100% sure so I did it in case.

PR #2038, opened 2024-08-14
test: Move from CDP to BiDi

croissanne:
See cockpit-project/cockpit#20832.

Explicitly add a "glob" devDependency to follow suit with cockpit-project/cockpit@680decc155a. It was previously used implicitly through a transient dependency of something else, but our esbuild po-plugin uses it explicitly.


croissanne:
Drafting this, centos-10 is really flaky on this.

kingsleyzissou (approved):

PR #2044, opened 2024-08-20
Align testing with cockpit-project

croissanne:
Cockpit defines their gating tests upstream, in `test/browser`, let's do the same.


kingsleyzissou (approved):
Looks good to me, from what I can tell. Do we want to try ask for one of the cockpit team to give a review too?

test/browser/browser.sh
```diff
@@ -0,0 +1,46 @@
+set -eux
+
+cd "${0%/*}/../.."
+
+# HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2033020
+dnf update -y pam || true
```
martinpitt, line 6: That was fixed over a year ago, you probably don't need this.

martinpitt (changes requested):
Thanks @croissanne this makes sense! However, to really validate this I recommend to actually enable at least one OS in your [packit config](https://github.com/osbuild/cockpit-composer/blob/main/packit.yaml), so that they run in upstream PRs. After all, you want to shift left the gating tests, no? Otherwise this will just bitrot.

martinpitt:
Note that COPR has some serious problems today, builds take ages.. so the TF results will also take a while, but they do come in eventually.

croissanne:
Is it just
```
  - job: tests
    identifier: self
    trigger: pull_request
    targets:
      -...

```?

martinpitt:
@croissanne I saw some "is it just.." question fly by, apparently you removed the comment. But the extra commit looks good in general. You mostly need to keep the `targets:` list up to date and to what you support in main. So let's wait for the results! :coffee:

martinpitt (dismissed):
Looks good in general, bots have the last word.

martinpitt:
@croissanne FTR, your [downstream dist-git](https://src.fedoraproject.org/rpms/cockpit-composer/tree/rawhide) still needs two things: The [.fmf/version stamp](https://src.fedoraproject.org/rpms/cockpit-podman/blob/rawhide/f/.fmf/version) and a [plan](https://src.fedoraproject.org/rpms/cockpit-podman/blob/rawhide/f/plans/upstream.fmf) that runs the tests from the dist-git's source tarball.

croissanne:
> @croissanne FTR, your [downstream dist-git](https://src.fedoraproject.org/rpms/cockpit-composer/tree/rawhide) still needs two things: The [.fmf/version stamp](https://src.fedoraproject.org/rpms/cockpit-podman/blob/rawhide/f/.fmf/version) and a [plan](https://src.fedoraproject.org/rpms/cockpit-podman/blob/rawhide/f/plans/upstream.fmf) that runs the tests from the dist-git's source tarball.

will add those as soon as the tests succeed here

kingsleyzissou:
Weird, looks like some of those rpm builds completed successfully. But the jobs are still pending

croissanne:
> Weird, looks like some of those rpm builds completed successfully. But the jobs are still pending

naw it's the srpm jobs that are completed. Copr is rlly struggling still 
afbeelding

https://copr.fedorainfracloud.org/status/stats/

kingsleyzissou (approved):
🎉 🎉

